PR #14, opened 2020-10-11
Improve documentation for `BLS12` fields

ValarDragon:


Pratyush:
Thanks!

ValarDragon:
Maybe we should generate these field files entirely from sage scripts?

PR #20, opened 2020-10-21
Add Polynomial interface and multivariate polynomials

ryanleh:


poly/Cargo.toml
```diff
@@ -17,6 +17,7 @@ ark-ff = { path = "../ff", default-features = false }
 ark-std = { git = "https://github.com/arkworks-rs/utils", default-features = false }
 rand = { version = "0.7", default-features = false }
 rayon = { version = "1", optional = true }
+derivative = { version = "2", default-features = false }
```
Pratyush, line 20:
```suggestion
derivative = { version = "2", default-features = false, features = [ "use_core" ] }
```

poly/src/polynomial/mod.rs
```diff
@@ -1,164 +1,93 @@
...
+    + PartialEq
+    + Eq
+    + Send
+    + Sync
+    + for<'a> AddAssign<&'a Self>
+    + for<'a> AddAssign<(F, &'a Self)>
+    + for<'a> SubAssign<&'a Self>
+{
+    /// The domain of the polynomial.
+    type Domain: Sized + Clone + Ord + Debug + Sync;
```
Pratyush, line 31: Let's come up with a better name for the associated type; this might cause confusion with the `EvaluationDomain` related-things.

poly/src/polynomial/mod.rs
```diff
@@ -1,164 +1,93 @@
...
 
-impl<'a, F: Field> Into<DensePolynomial<F>> for DenseOrSparsePolynomial<'a, F> {
-    fn into(self) -> DensePolynomial<F> {
-        match self {
-            DPolynomial(p) => p.into_owned(),
-            SPolynomial(p) => p.into_owned().into(),
-        }
-    }
+    /// Sample a random point from `Self::Domain`.  
+    fn rand_domain_point<R: Rng>(domain_size: Option<usize>, rng: &mut R) -> Self::Domain;
```
Pratyush, line 54: What is the use case for this? What is `domain_size` doing here? Can we just require `Domain: UniformRand`?

poly/src/polynomial/mod.rs
```diff
@@ -1,164 +1,93 @@
...
-        SPolynomial(Cow::Borrowed(other))
-    }
-}
+    /// If `num_vars` is `None`, outputs a polynomial a univariate polynomial
+    /// of degree `d` where each coefficient is sampled uniformly at random.
+    ///
+    /// If `num_vars` is `Some(l)`, outputs an `l`-variate polynomial which
+    /// is the sum of `l` `d`-degree univariate polynomials where each coefficient
+    /// is sampled uniformly at random.
+    fn rand<R: Rng>(d: usize, num_vars: Option<usize>, rng: &mut R) -> Self;
```
Pratyush, line 51: I presume that `num_vars` is required for hiding in MarlinPST13? Can we move it to a helper method outside the trait?

poly/src/polynomial/mod.rs
```diff
@@ -1,164 +1,93 @@
...
 
-    #[inline]
-    fn leading_coefficient(&self) -> Option<&F> {
-        match self {
-            SPolynomial(p) => p.last().map(|(_, c)| c),
-            DPolynomial(p) => p.last(),
-        }
-    }
+    /// Constructs a new polynomial from a list of tuples of the form `(Self::Term, coeff)`
+    fn from_coefficients_vec(num_vars: usize, terms: Vec<(Self::Term, F)>) -> Self;
```
Pratyush, line 80: Does it make sense to move these methods into `Polynomial`? Then `UVPolynomial` will set `Term = F` or something along those lines, and `from_coefficients_vec` will just invoke that of `Polynomial`?

poly/src/polynomial/mod.rs
```diff
@@ -1,164 +1,93 @@
...
-                Evaluations::from_vec_and_domain(d.coeffs, domain)
-            }
-        }
-    }
+    /// Given some point `z`, compute the quotients `w_i(X)` s.t
+    ///
+    /// `p(X) - p(z) = (X_1-z_1)*w_1(X) + (X_2-z_2)*w_2(X) + ... + (X_l-z_l)*w_l(X)`
+    ///
+    /// These quotients can always be found with no remainder.
+    fn divide_at_point(&self, point: &Self::Domain) -> Vec<Self>
```
Pratyush, line 90: Can this be made into helper methods in `poly-commit`, instead of being a required method on the trait?

poly/src/polynomial/mod.rs
ryanleh: Hmm so currently `UVPolynomial` doesn't have the `Term` associated type since I don't think it would ever be something besides `F`? But also the multivariate version of `from_coefficients_vec` additionally requires `num_vars` to be passed so I think it makes sense to have them separate considering that's what we did for `rand` as well.

poly/src/polynomial/mod.rs
```diff
@@ -1,164 +1,81 @@
...
+    + Hash
+    + PartialEq
+    + Eq
+    + Send
+    + Sync
+    + for<'a> AddAssign<&'a Self>
+    + for<'a> AddAssign<(F, &'a Self)>
+    + for<'a> SubAssign<&'a Self>
+{
+    /// The domain of the polynomial.
```
Pratyush, line 30:
```suggestion
    /// The type of evaluation points for this polynomial
```

poly/src/polynomial/multivariate/mod.rs
```diff
@@ -0,0 +1,153 @@
...
+use rayon::prelude::*;
+
+mod sparse;
+pub use sparse::SparsePolynomial;
+
+/// Describes the interface for a term of a multivariate polynomial.
+pub trait Term:
+    Clone + PartialOrd + Ord + PartialEq + Eq + Hash + Default + Debug + Deref + Send + Sync
+{
+    /// Create a new Term from a list of tuples of the form `(variable, power)`
```
Pratyush, line 21:
```suggestion
    /// Create a new `Term` from a list of tuples of the form `(variable, power)`
```

poly/src/polynomial/multivariate/mod.rs
```diff
@@ -0,0 +1,153 @@
...
+    fn new(term: Vec<(usize, usize)>) -> Self;
+
+    /// Returns the sum of all variable powers in `self`
```
Pratyush, line 24:
```suggestion
    /// Returns the total degree of `self`. This is the sum of all variable powers in `self`
```

poly/src/polynomial/multivariate/mod.rs
```diff
@@ -0,0 +1,153 @@
...
+    fn degree(&self) -> usize;
+
+    /// Returns a list of variables in `self`
+    fn vars(&self) -> Vec<usize>;
+
+    /// Returns a list of variable powers in `self`
```
Pratyush, line 30:
```suggestion
    /// Returns a list of the powers of each variable in `self`
```

poly/src/polynomial/multivariate/mod.rs
```diff
@@ -0,0 +1,153 @@
...
+    fn powers(&self) -> Vec<usize>;
+
+    /// Returns whether `self` is a constant
+    fn is_constant(&self) -> bool;
+
+    /// Evaluates `self` at the given `point` in the field.
+    fn evaluate<F: Field>(&self, point: &[F]) -> F;
```
Pratyush, line 37:
```suggestion
    /// Evaluates `self` at the point `p` in the field.
    fn evaluate<F: Field>(&self, p: &[F]) -> F;
```

poly/src/polynomial/multivariate/mod.rs
```diff
@@ -0,0 +1,153 @@
...
+}
+
+/// Stores a single term in a multivariate polynomial. Each element
+/// is of the form `(variable, power)`.
```
Pratyush, line 41: What do you mean by "single term"?

poly/src/polynomial/multivariate/sparse.rs
```diff
@@ -0,0 +1,403 @@
...
+
+/// Stores a sparse multivariate polynomial in coefficient form.
+#[derive(Derivative)]
+#[derivative(Clone, PartialEq, Eq, Hash, Default)]
+pub struct SparsePolynomial<F: Field, T: Term> {
+    /// The number of variables the polynomial supports
+    #[derivative(PartialEq = "ignore")]
+    pub num_vars: usize,
+    /// List of every term along with its coefficient
+    pub terms: Vec<(T, F)>,
```
Pratyush, line 26: I think storing in `(coeff, term)` form would be clearer as it maps to how you write the polynomial, eg: p(x, y) = 2 * (x^2 * y^2)`.

```suggestion
    pub terms: Vec<(F, T)>,
```

poly/src/polynomial/multivariate/mod.rs
ryanleh: Single term = monomial. Would you prefer to use `Monomial` instead of `Term`? Or just 'single term' -> 'term'?

Pratyush: Ahh I was just confused by seeing single term and then seeing a vec inside. I understand now. I'm fine with either term or monomial; which one do you think would be clearer?

ryanleh: Gotcha. To me `Term` is clearer and cleaner in the code but I don't think it's a huge difference either way. Perhaps we can just make sure it's clear from the documentation that term=monomial?

Pratyush: Yeah let's go with that. `Term` is also shorter

poly/src/polynomial/mod.rs
```diff
@@ -1,164 +1,84 @@
...
-            }
-            DPolynomial(Cow::Owned(mut d)) => {
-                domain.fft_in_place(&mut d.coeffs);
-                Evaluations::from_vec_and_domain(d.coeffs, domain)
-            }
-        }
-    }
+    /// Outputs an `l`-variate polynomial which is the sum of `l` `d`-degree univariate
+    /// polynomials where each coefficient is sampled uniformly at random.
+    fn rand<R: Rng>(d: usize, num_vars: usize, rng: &mut R) -> Self;
```
Pratyush, line 83: Maybe we should generalize this to take in the monomial structure of the polynomial? I.e. a list of terms?

poly/src/polynomial/multivariate/mod.rs
```diff
@@ -0,0 +1,163 @@
...
+        self.len() == 0
+    }
+
+    /// Evaluates `self` at the given `point` in the field.
+    fn evaluate<F: Field>(&self, point: &[F]) -> F {
+        cfg_into_iter!(self)
+            .map(|var| {
+                let term = point.get(var.0).unwrap();
+                term.pow(&[var.1 as u64])
+            })
```
Pratyush, line 115:
```suggestion
            .map(|(var, power)| point[var].pow(&[power as u64]))
```

poly/src/polynomial/multivariate/sparse.rs
```diff
@@ -0,0 +1,408 @@
...
+    /// Checks if the given polynomial is zero.
+    fn is_zero(&self) -> bool {
+        self.terms.is_empty() || self.terms.iter().all(|(c, _)| c.is_zero())
+    }
+
+    /// Returns the total degree of the polynomial
+    fn degree(&self) -> usize {
+        self.terms
+            .iter()
+            .map(|(_, term)| (*term).degree())
```
Pratyush, line 55: I think this should work?
```suggestion
            .map(|(_, term)| term.degree())
```

poly/src/polynomial/multivariate/sparse.rs
```diff
@@ -0,0 +1,408 @@
...
+        for term in terms {
+            match terms_dedup.last_mut() {
+                Some(prev) => {
+                    if prev.1 == term.1 {
+                        *prev = (prev.0 + term.0, prev.1.clone());
+                        continue;
+                    }
+                }
+                _ => {}
+            };
```
Pratyush, line 90:
```suggestion
            if let Some(prev) = terms_dedup.last_mut() {
                if prev.1 == term.1 {
                    *prev = (prev.0 + term.0, prev.1.clone());
                    continue;
                }
            }
```

poly/src/polynomial/multivariate/sparse.rs
```diff
@@ -0,0 +1,408 @@
...
+            result.push(match which {
+                Some(Ordering::Less) => cur_iter.next().unwrap().clone(),
+                Some(Ordering::Equal) => {
+                    let other = other_iter.next().unwrap();
+                    let cur = cur_iter.next().unwrap();
+                    (cur.0 + other.0, cur.1.clone())
+                }
+                Some(Ordering::Greater) => other_iter.next().unwrap().clone(),
+                None => break,
+            });
```
Pratyush, line 158: I'd extract the match out to a separate intermediate variable

poly/src/polynomial/multivariate/sparse.rs
```diff
@@ -0,0 +1,408 @@
...
+            num_vars: other.num_vars,
+            terms: other
+                .terms
+                .iter()
+                .map(|(coeff, term)| (*coeff * &f, term.clone()))
+                .collect(),
+        };
+        *self = &*self + &other;
+        self.remove_zeros()
+    }
```
Pratyush, line 189: Doesn't this leave duplicates? (Might not be a problem, I just noticed that you do deduplication elsewhere.)

poly/src/polynomial/mod.rs
ryanleh: Hm I'm not quite sure I understand. What would be a use case where that might be useful?

poly/src/polynomial/multivariate/sparse.rs
ryanleh: It shouldn't because the implementation of `Add` combines any duplicates

poly/src/polynomial/mod.rs
ryanleh: Added an issue for future resolution

Pratyush:
@weikengchen @npwardberkeley  friendly heads up, since you looked at this code recently too, you might want to take a look

There are some downstream impacts, but mostly on imports

Pratyush (approved):

PR #38, opened 2020-11-13
Add Add/Neg trait bounds to Polynomial

ryanleh:


Pratyush:
You probably also want at least `Add<&Self>`, otherwise you’ll have to clone all over the place

PR #50, opened 2020-11-19
Add serialization trait bounds to `Polynomial`

ryanleh:


Pratyush:
Hmm for the time being I would implement it just for `UnivariatePolynomial`; I don't think it makes sense to implement for multivariate polynomials just yet. I would also not include it as trait bound here; instead let's just add the bounds at the use sites in the delegation project.

ryanleh:
Ok, I just derived it for `DensePolynomial`

ryanleh:
Or wait, I think it will need to be a bound on `UVPolynomial` for it to work with the structs in `poly-commit`

poly/src/polynomial/univariate/sparse.rs
```diff
@@ -2,10 +2,11 @@
 use crate::univariate::{DenseOrSparsePolynomial, DensePolynomial};
 use crate::{EvaluationDomain, Evaluations, UVPolynomial};
 use ark_ff::{FftField, Field};
+use ark_serialize::*;
 use ark_std::{collections::BTreeMap, fmt, vec::Vec};
 
 /// Stores a sparse polynomial in coefficient form.
-#[derive(Clone, PartialEq, Eq, Hash, Default)]
+#[derive(Clone, PartialEq, Eq, Hash, Default, CanonicalSerialize, CanonicalDeserialize)]
```
Pratyush, line 9: Does this need to be serializable?

ryanleh: In order for the `marlin_pc::Randomness` struct to be serializable, you need `UVPolynomial: CanonicalSerialize + CanonicalDeserialize` so ya.

Pratyush: but `SparsePolynomial` doesn't implement `UVPolynomial`, no? Or am I not finding the right impl?

(I don't see `DenseOrSparsePolynomial` in the diff, so I assume you're not deriving it for that)

ryanleh: Oh good point haha I forgot about that - will adjust.

Issue #34, opened 2020-11-12
Make CI test against `curves` repo


Pratyush:
Add a new job in `CI` that does the following steps:

- Clone `curves` repo.
- Add a [patch](https://doc.rust-lang.org/edition-guide/rust-2018/cargo-and-crates-io/replacing-dependencies-with-patch.html) in `curves/Cargo.toml` to force the dependency tree to use the local version of the `algebra` crates.
- Run all tests in `curves`.

This can prevent changes from being merged in here that actually break downstream

PR #51, opened 2020-11-19
Remove unitary inverse

### LINKED ISSUE #49: Remove unitary_inverse from quadratic extension
On a quadratic extension field, `x.unitary_inverse() = x.clone().conjugate()`. The method is barely used, its only used in cyclotomic exponentiation. 

imo, it feels like we shouldn't be using the name `unitary_inverse` and conjugate. Having both methods is just API bloat / confusion. I'd suggest deleting it, or alternatively renaming them to `conjugate()` and `conjugate_in_place()`.

ValarDragon:
Closes #49

### DIFF
--- a/ff/src/fields/mod.rs
+++ b/ff/src/fields/mod.rs
@@ -108,7 +108,7 @@ pub trait Field:
     type BasePrimeField: PrimeField;
 
     /// Returns the characteristic of the field,
-    /// in big-endian representation.
+    /// in little-endian representation.
     fn characteristic<'a>() -> &'a [u64] {
         Self::BasePrimeField::characteristic()
     }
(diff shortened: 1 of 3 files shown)


ff/src/fields/mod.rs
```diff
@@ -107,7 +107,8 @@ pub trait Field:
 {
     type BasePrimeField: PrimeField;
 
-    /// Returns the characteristic of the field.
+    /// Returns the characteristic of the field,
+    /// in big-endian representation.
```
Pratyush, line 111:
```suggestion
    /// in little-endian representation.
```

Pratyush:
btw can you confirm that `curves` compiles with this change? You can follow the steps here: https://github.com/arkworks-rs/algebra/issues/34

ff/src/fields/mod.rs
ValarDragon: ah you're right, got them mixed up 😅

ValarDragon:
It still compiles (assuming I did the instructions in #34 correctly)

Pratyush (approved):

PR #53, opened 2020-11-19
Add `Zero` trait bound to `Polynomial`

ryanleh:
These were already required methods for `Polynomial`, but moving it to the appropriate trait allows us to generically handle polynomials as additive shares (which require an additive identity).


Pratyush (approved):

Pratyush:
Ahh I forgot to ask you to update the changelog; do you mind submitting a new PR for that?

PR #96, opened 2020-11-27
Upgrade `field_new` macro to take in signed integers 

Pratyush:
So far, the `field_new!` macro is quite unintuitive: it requires users to write prime field elements in Montgomery form already, which usually requires a detour through a sage script. This PR adds a feature for converting to Montgomery form *at compile time*, which makes it possible to specify constants in normal form, and have these be automatically converted. 

---

write a little note why.

- [x] Wrote unit tests


ValarDragon:
Awesome change! 

We should also add a short description somewhere for finding where the actual arithmetic of a field is implemented. I'm worried that this adds another layer of separation between specifying field parameters and being able to find / understand the field arithmetic's implementation

Perhaps this is what should go in an `ark-ff` readme? This doesn't need to block this pr though

Pratyush:
Yeah, there's a couple of levels of indirection that need to be traced through to find the actual impl. We should indeed document the code to explain the structure.

(I don't think this adds any more levels compared to the previous impl though)

jon-chuang:
Is it possible to compile other parameters, like `MODULUS_MINUS_ONE_DIV_TWO` at compile time too?

jon-chuang:
Also, wouldn't you say that utilising the macros for adc/sbb/mac_with_carry etc. is a step backwards in terms of relying on macros instead of functions?

Is there any real need for it?

Pratyush:
> Is there any real need for it? My guess is you want to deduplicate code across the `const fn` and non const `fn` uses?

Yup!
> In this case, I think one should convert all the original functions to `const fn`. As a `const fn` can ingest a non-constant result and output a non-const result. So it can play the dual purpose.

I did do this initially, but the problem right now is that `const fn`s don't support mutable parameters, which means that I had to write the code as follows:
```rust
let (new_a, new_carry) = adc(a, b, carry);
a = new_a;
carry = new_carry;
```
This is much less readable than the original =/.

> One can also get rid of the duplication of code that is impl for `FpParameters`.
> 

Yes, I believe so, at least partially. And I think we can do it in a way that's more auditable than the way it's done in `ff_derive`.


> I'm not sure if `const fn` can be used to define `const` values generically in traits. (Unfortunately, I think not). 

It is possible.



> I think most of the `const fn` functionality should reside in biginteger, we should add a div with remainder as well. This can possibly be used to compute R at compile-time too.

That would be lovely =).

> 
> Generators are by default only 2 for prime fields right, with no variation? In which case the only params left to impl concretely are modulus and two_adicity.

two-adicity can also be found in const fn, just keep dividing by two until there are no more factors left.


> Theoretically, one could also find square roots of unity and thus find the two-adic root of unity

I think this one is a bit more tricky. `ff_derive` does it by exponentiating by `FpParameters::T`, which might exhaust [`const_eval_limit`](https://doc.rust-lang.org/beta/unstable-book/language-features/const-eval-limit.html). (I ran into this limit in a prior version of this PR).

jon-chuang:
> I did do this initially, but the problem right now is that const fns don't support mutable parameters...
I see, I think there is some work to allow mutable parameters.
However, the situation is not as bad as you describe. One can simply write `let (mut a, mut carry) = adc(a, b, carry);`

jon-chuang:
Unfortunately, `impl const ..` is still an incomplete feature (https://github.com/rust-lang/rfcs/pull/2632, https://github.com/rust-lang/rust/issues/71971), so there is no straightforward way to make `neg()` a `const fn`. Very sad.

jon-chuang:
In the opposite direction then, I propose to rewrite the traits in terms of the `const fn` definitions. It's less than ideal but at least cleaner than having multiple definitions.

Pratyush:
> However, the situation is not as bad as you describe. One can simply write `let (mut a, mut carry) = adc(a, b, carry);`

This doesn't work when `a` is actually `array[$i]`.

> In the opposite direction then, I propose to rewrite the traits in terms of the const fn definitions. It's less than ideal but at least cleaner than having multiple definitions.

I would rather just wait until `impl const Trait...` is stable; it should hopefully be happening within the year.

jon-chuang:
Fair enough, let's consolidate into an issue

jon-chuang:
@Pratyush how do you feel about using proc macros to perform all constant evaluation instead? I feel much error-prone copy pasta can be avoided as desired with little downside.

Pratyush:
This is the approach `ff_derive` takes, but it only does it for `PrimeField`s. We could certainly adapt that technique to expand it to other domains, but doing so is tricky because macros can't access values, which means they can't access things like the value of the modulus.

jon-chuang:
Hmm, can't one do as per your proc macro in this PR and parse a &[u64] etc? How is this not accessing the value?

Pratyush:
What I meant is that it's not possible to access the value when you're given just the identifier. For example, in `cool_macro!(P::MODULUS)`, the only thing `cool_macro` receives are the tokens `P::MODULUS`; not the actual value of `P::MODULUS`. So `cool_macro` can only perform syntactic transformation on `P::MODULUS`.

jon-chuang:
I don't think you're right about this, this is certainly true of ordinary macros, but in your `to_sign_and_limbs` macro, particularly in invoking `BigInteger::from_str`, one does get access to the values which can be used for computation.

jon-chuang:
In conclusion, I believe utilising `const fn` at this stage of the rustc stabilisation to be completely unecessary, and potentially introducing undesirable bloat.

Pratyush:
> I don't think you're right about this, this is certainly true of ordinary macros, but in your `to_sign_and_limbs` macro, particularly in invoking `BigInteger::from_str`, one does get access to the values which can be used for computation.

Right, in this case I explicitly provide the token "1234555555" (for example) to the macro, so the macro can parse it and can manipulate a string "123455555".

> Since recreating this functionality is complicated, it seems more straightforward to create functionality to translate between num_bigint BigInts and our BigIntegers, perhaps strictly for the proc macro crate internal use.
> 
> I think that as long as we can tweak num_bigint to build with BigDigit = u64 all the time, this would be perfectly fine to build on any target (otherwise this translation would be hairy for 32-bit targets.)
> 
> If not, one can simply check if BigDigit is u32 or u64 and conditionally make the conversion. (one can also just use `to_u32_digits` to get a vec).
> 
> Or, better still, one could work directly with the BigInt algebra to derive the params.

I'm not sure I understand what you mean here; for use inside the proc macro implementation, we don't care about the internal representation of `num_bigint::BigInt`, just that it can output its internal representation into `u64` limbs, which it can.

jon-chuang:
Well sure, but what I'm saying is that one can access the u64 values and generate other parameters from them entirely within the proc macros, without relying on const fn. However, one does have to redefine some functionality, such as Fp mul_assign etc. Given that one has to do so for either path, now I'm unsure which approach is better.

Pratyush:
To clarify, I agree that this approach works when generating parameters for prime fields. It *does not* work when generating parameters for other things (elliptic curves, extension fields, etc) unless you explicitly pass the value of the underlying prime modulus to the macro.

daira:
It would certainly have been easier to add the Pasta curves (#21) with this improvement. In particular it wasn't entirely clear which constants needed to be in Montgomery form, since there's no naming convention that distinguishes Montgomery vs plain representations.

Pratyush:
Yeah sorry about that 😅

(Those frustrations were precisely what motivated this PR)

ff-macros/src/lib.rs
```diff
@@ -0,0 +1,88 @@
+#![deny(
+    warnings,
+    unused,
+    future_incompatible,
+    nonstandard_style,
+    rust_2018_idioms
+)]
+#![forbid(unsafe_code)]
```
ValarDragon, line 8: I'm confused as to why we want forbid here vs deny

ff-macros/src/lib.rs
```diff
@@ -0,0 +1,88 @@
...
+#![recursion_limit = "128"]
```
ValarDragon, line 9: Could a comment be added for why we set a recursion limit here?
IIRC, you had mentioned that its something to do with constraints on compile time macros at the moment

ff/src/fields/arithmetic.rs
```diff
@@ -155,19 +151,22 @@ macro_rules! impl_field_square_in_place {
             r[1] <<= 1;
 
             for i in 0..$limbs {
-                r[2 * i] = fa::mac_with_carry(r[2 * i], (self.0).0[i], (self.0).0[i], &mut carry);
-                r[2 * i + 1] = fa::adc(r[2 * i + 1], 0, &mut carry);
+                r[2 * i] = mac_with_carry!(r[2 * i], (self.0).0[i], (self.0).0[i], &mut carry);
+                #[allow(unused_assignments)]
```
ValarDragon, line 155: Where is the unused assignment coming from?

ff/src/fields/macros.rs
```diff
@@ -115,6 +116,117 @@ macro_rules! impl_Fp {
             pub const fn new(element: $BigIntegerType) -> Self {
                 Self(element, PhantomData)
             }
+
+            #[ark_ff_asm::unroll_for_loops]
+            const fn const_is_zero(&self) -> bool {
```
ValarDragon, line 121: Should all changes to field arithmetic now be duplicated in these macros?

ff-macros/src/lib.rs
Pratyush: forbid cannot be bypassed with an `allow`

ff/src/fields/arithmetic.rs
Pratyush: The last value assigned to carry (inside the macro) is unused

ff/src/fields/macros.rs
Pratyush: Hmm I don't anticipate these methods will change much, but if you're asking whether we should have a comment on the structs indicating that both copies should be updated if either one is, then yes. Longer term, the goal would be to deduplicate the code, and just use one impl in both const and non-const contexts.

ff/src/fields/arithmetic.rs
ValarDragon: That makes sense, and we just assume that the compiler will remove that in an optimization pass. Could a comment be added to that affect?

ff-macros/src/lib.rs
ValarDragon: Ohh, that makes sense. Why don't we default to having things be forbidden, unless we want a local allow then?

Pratyush: I think in most crates we have a forbid?

ff/src/fields/macros.rs
ValarDragon: Cool! Can you add a comment to that affect for each one

Pratyush: Will do

(maybe I'll even do a partial deduplication now)

ff-macros/src/lib.rs
Pratyush: I'll have to check, but I think it's not an issue in ff-macros

Pratyush: Addressed

ff/src/fields/macros.rs
Pratyush: I did a partial deduplication, so that common logic is stored in a separate method that both the `const` and non-`const` versions invoke.

ValarDragon (approved):
This LGTM.

Issue #83, opened 2020-01-14
Make FFTs nlog(d)


ValarDragon:
FFTs in the library are all nlog(n). However they could be nlog(d), which is an efficiency improvement when doing polynomial multiplication. 

This can be taken advantage of in two possible ways:
* Split up the FFT into n/d independent FFTs operating over the same polynomial, but evaluating on distinct cosets. Both of these sub-FFTs are dlog(d). At the end concatenate these two vectors.
* Skip the first few rounds of butterflying / arithmetic in the current FFT algorithm, and just copy the polynomial coefficients into the correct positions. This is implemented in libiop.

This probably doesn't matter for the current pairing based SNARKs as they are dominated by MSM time.

Pratyush:
This would help any time we do a polynomial multiplication, since the domain is necessarily equal to the sum of the degrees of the involved polynomial, and is hence usually at least double their degree.

Pratyush:
@andrewmilson this change becomes possible with your work in #487; if you would like to take a stab at implementing the first approach @ValarDragon mentioned, that would be fantastic!

andrewmilson:
Thanks @Pratyush. Yeah keen to take a stab!

Pratyush:
Awesome! Let me know if you have any questions about the approach; you can ask here or on Discord.

andrewmilson:
@Pratyush unless I'm missing something I think it might be easier going straight to the second approach @ValarDragon mentioned. 

With the first approach It's easy to obtain the values for the smaller FFTs evaluated on the coset domains but transposing the values so they are in order is not so easy. I haven't spent too much time looking into it but I don't think it's trivial to do without allocating extra memory.

Pratyush:
You're right that the shuffles to put things in place are somewhat non-trivial, and might cause bad non-locality. It would be good to check what the situation would be like for the other approach.

Re: extra memory, we will need to allocate extra memory either way, since the output (evaluations) is larger than the input (polynomial)

andrewmilson:
With the first approach I mean additional memory on top of resizing the coefficients to the size of the domain.

The second approach has better locality and no need to allocate more memory than necessary. I put together a quick draft PR #493 just for ideas. I know you mentioned above you are "intending to refactor the FFT infrastructure". I'd be keen to hear what your thoughts are here. I'm more than happy to take the implementation further, just let me know.

PR #103, opened 2020-11-28
Add documentation for parallel FFT algorithm

ValarDragon:
## Description

Renames variables and documents the parallel FFT. I was going to start work on #83 , but then realized I didn't actually understand our paralell FFT algorithm, so didn't know how to best adapt it to this setting. The implemented algorithm seems like it was a direct translation of the C++ parallel FFT algorithm in libfqfft, which is also completely undocumented.

I'm still not sure what the underlying algorithm is. (Its not quite Cooley Tukey, since both the FFT's being done are of the same size) Its definitely slower than it ought to be at the moment.

I left comments in the code pointing out how to reduce its arithmetic complexity at least from `2N + |coset size| FFT` to `2 * |coset size| FFT` per thread. However that is still higher than Cooley Tukey (but perhaps Cooley Tukey is less clear to parallelize or will have less cache locality).

---

write a little note why.

- [x] Linked to Github issue with discussion and accepted design OR have an explanation in the PR that describes this work. - N/A, functionality unchanged at the moment
- [ ] Wrote unit tests - N/A
- [x] Added a relevant changelog entry to the `Pending` section in `CHANGELOG.md` - N/A

### DIFF
--- a/poly/src/domain/utils.rs
+++ b/poly/src/domain/utils.rs
@@ -58,40 +58,64 @@ pub(crate) fn parallel_fft<T: DomainCoeff<F>, F: FftField>(
     serial_fft: fn(&mut [T], F, u32),
 ) {
     assert!(log_n >= log_cpus);
+    // For documentation purposes, comments explain things
+    // as though `a` is a polynomial that we are trying to evaluate.
 
+    // Partition `a` equally into the number of threads.
+    // each partition is then of size m_div_num_threads
     let m = a.len();
-    let num_chunks = 1 << (log_cpus as usize);
-    assert_eq!(m % num_chunks, 0);
-    let m_div_num_chunks = m / num_chunks;
+    let num_threads = 1 << (log_cpus as usize);
+    let num_cosets = num_threads;
+    assert_eq!(m % num_threads, 0);
+    let coset_size = m / num_threads;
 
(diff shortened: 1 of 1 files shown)


Pratyush:
Amazing, thanks for this! this might explain the slowdown here: https://github.com/arkworks-rs/groth16/issues/11

ValarDragon:
Oh cool! Glad theres something to validate that this is bad (I was a bit worried that noone had flagged this, since it seemed like a parallelism blocker). I think this does explain that slowdown. I can work on trying to go from `2N -> 2 coset size FFT` work per thread after this.

At some later time, we probably need to have a broader consideration of the parallel FFT algorithm. Its possible to parallelize evenly with no arithmetic overhead, but this is bad for data locality and one would expect would slow things down. But theres a ton of different trade-offs to go for in the "increase arithmetic complexity, improve data locality / parallel speed" that we need to be explicit about. (I'm pretty confident that the current method is what we will end with!)

Pratyush:
So far this LGTM! I can't say that I fully understand the details of the algorithm; I mostly ported it over from the original Bellman implementation 😅

Pratyush (approved):

weikengchen:
(I can't wait to update my benchmark results)

PR #106, opened 2020-12-03
Add Zeroize to fields and elliptic curves

ValarDragon:
## Description

This PR implements Zeroize for the Field, AffineGroup, and ProjectiveGroup traits.  

Why Zeroize is helpful:

If you operate in the threat model of an attacker getting user space memory leakage (E.g. Heartbleed, cold boot attacks, DMA attacks), then a method to hopefully lower the probability of a succesful attack is to reduce the amount of time secrets are in memory. 

This is pretty hard to do in the ideal sense, as you'd want all associated temporaries to be wiped as soon as possible (including register spills, the thread being switched etc.).

However if you have all temporaries be created in stack, then you can get some level of memory wiping by Zero'ing out your secret itself and everything on heap when you are done with them, and then hoping that everything that was created on stack gets written over relatively soon (or manually doing that with unsafe / some other method)

Adding Zeroize gives end application developers more of an ability to do the above, since they can now zeroize secrets, and use the `secret` trait, to allocate secrets on heap and have them be zero'd out when they're done. 

cref: https://github.com/arkworks-rs/snark/issues/111

---

write a little note why.

- [ ] Wrote unit tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 - #51 (ark-ff) Removed `unitary_inverse` from `QuadExtField`. Make this change by
     replacing `x.unitary_inverse()` with `let mut tmp = x.clone(); tmp.conjugate()`
 - #53 (ark-poly) Add `Zero` trait bound to `Polynomial`.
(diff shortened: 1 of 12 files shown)


Pratyush:
Amazing, thanks!

Pratyush (approved):

Issue #61, opened 2020-11-03
Change all scalar mul to accept slices instead of bigint


jon-chuang:
By utilising bigint instead of field, we are already losing some invariant on the size of the arg. 

Further, it's clear this has little bearing on code correctness. Hence, I suggest that either scalar_mul be parameterizable by a bigint param, or we simply accept slices of &[u64] to allow for a more flexible interface.

PR #110, opened 2020-12-04
Change EC scalar mul from taking a bigint to taking AsRef<[u64]>

ValarDragon:
## Description

Changes the elliptic curve scalar mul function to now take in `AsRef<[u64]>`. This change has been tested to not require any update to `curves` or `r1cs-std`, per the steps in #34. This is making me suspect that the change is not breaking, but that would be surprising. 

I don't see why `Into<Bigint>` would also be `AsRef<[u64]>`, so I'm listing this as a breaking change for now. Do you know @Pratyush?

closes: #61 
---

write a little note why.

- [x] Linked to Github issue with discussion and accepted design OR have an explanation in the PR that describes this work. - Not sure if it was accepted per se, but its more consistent, and lets one multiply by a small constant more easily
- [x] Wrote unit tests - correctness is caught by existing tests
- [x] Updated relevant documentation in the code - don't see any docs to update

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
     replacing `x.unitary_inverse()` with `let mut tmp = x.clone(); tmp.conjugate()`
 - #53 (ark-poly) Add `Zero` trait bound to `Polynomial`.
 - #106 (ark-ff, ark-ec) Add `Zeroize` trait bound to `Field, ProjectiveGroup, AffineGroup` traits.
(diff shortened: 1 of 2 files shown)


Pratyush:
`BigInteger`s are required to be `AsRef<[u64]>`, so if you're passing in those then nothing breaks. I think generally most uses of scalar multiplication are either via the `Mul` impl wrt the `ScalarField`, or via the `mul` method on `AffineCurve` (as that uses mixed addition)

ValarDragon:
Oh that makes sense, I forgot that all types T automatically implement Into<T>.

I don't think any update to the changelog is needed then (its already in the breaking changes section)

Pratyush (approved):

Issue #95, opened 2020-11-27
Too liberal bool deserialize?


jon-chuang:
This is the current deserialize for bool:
`Ok(u8::deserialize(reader)? == 1)`

However, this is not necessarily the behaviour we want. In fact, I think we should error if the result is not in `{0, 1}`, which I think might have been the old behaviour.

ValarDragon:
I'm in favor of ensuring that its in `{0,1}` (and in general making encodings unique)

PR #112, opened 2020-12-06
Make `bool`'s deserialization non-malleable

ValarDragon:
## Description

Makes the deserialization of bool non-malleable. Now the only strings that can be deserialized as a boolean are the encodings of `0u8, 1u8`.

closes: #95 
---

write a little note why.

- [x] Wrote unit tests


Pratyush:
Small comment: instead of arbitrarily mutating the data, why not try mutating it to every possible byte value? That’ll remove the dependence on ark-ff

ValarDragon:
That works for testing bool, but wouldn't generalize if we wanted to test this for any data type with a larger serialization.

We have no other data type implemented in serialize (AFAIK) which has a non-1-to-1 encoding to test for atm, so maybe that point is moot. (Also for the current API to generalize well, we'd probably need to change `valid_mutation` to return type `T` instead of a `valid/invalid` boolean)

Maybe it'd make sense to refactor this method out of tests, and do the above change, so other types (e.g. Fields) can check this?

serialize/Cargo.toml
```diff
@@ -16,6 +16,10 @@ edition = "2018"
 ark-serialize-derive = { path = "../serialize-derive", optional = true }
 ark-std = { git = "https://github.com/arkworks-rs/utils", default-features = false }
 
+[dev-dependencies]
+ark-ff = { path = "../ff" }
```
Pratyush, line 20: I think you can just use `thread_rng()` from `rand`; no need to import `ark-ff` for this.

ValarDragon: thread_rng requires std https://docs.rs/crate/rand/0.7.0/source/src/lib.rs

Pratyush: How about `StdRng`? Given that we want to clean up this infrastructure in the future, I don't think we should add this dependency here (even if it results in some duplication rn)

ValarDragon: Done!

Pratyush (approved):

PR #115, opened 2020-12-07
Add parallel implementation of operations on Evaluations

ValarDragon:
## Description

Add parallel implementation of operations on the Evaluations struct. 

---

write a little note why.

- [ ] Wrote unit tests - should be covered by existing tests? Not sure if there are any though. Batch inversion change is covered by prior tests.
- [x] Updated relevant documentation in the code - n/a

### DIFF
--- a/poly/src/evaluations.rs
+++ b/poly/src/evaluations.rs
@@ -68,8 +68,7 @@ impl<'a, F: FftField, D: EvaluationDomain<F>> MulAssign<&'a Evaluations<F, D>>
     #[inline]
     fn mul_assign(&mut self, other: &'a Evaluations<F, D>) {
         assert_eq!(self.domain, other.domain, "domains are unequal");
-        self.evals
-            .iter_mut()
+        ark_std::cfg_iter_mut!(self.evals)
             .zip(&other.evals)
             .for_each(|(a, b)| *a *= b);
     }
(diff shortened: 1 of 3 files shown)


poly/src/evaluations.rs
```diff
@@ -146,8 +146,7 @@ impl<'a, F: FftField, D: EvaluationDomain<F>> DivAssign<&'a Evaluations<F, D>>
     #[inline]
     fn div_assign(&mut self, other: &'a Evaluations<F, D>) {
         assert_eq!(self.domain, other.domain, "domains are unequal");
-        self.evals
-            .iter_mut()
+        ark_std::cfg_iter_mut!(self.evals)
```
Pratyush, line 149: I guess we should use batch inversion here? (I know the old code didn’t)

ValarDragon: Sure, I'll add a parallel batch inversion method.

Pratyush:
hmm the new batch inversion does more inversions, no? maybe you can improve the cost further by doing a `serial_batch_inversion` across all the per-batch inversions?

ValarDragon:
The new batch inversion only does one inversion per thread, which is all done in parallel

Pratyush (approved):

Issue #116, opened 2020-12-07
Make SparsePolynomial implement Polynomial


ValarDragon:
SparsePolynomial ought to implement the Polynomial trait. 

It doesn't make sense for SparsePolynomial to implement the current UVPolynomial Trait, as it handles coefficients differently. Perhaps the UVPolynomial trait ought to be changed?

(Its coefficient type is `(usize, F)` as opposed to `F`)

PR #117, opened 2020-12-07
Sparse poly implement poly

ValarDragon:
## Description

This PR makes SparsePolynomial implement Polynomial, and adds several more tests to SparsePolynomial.

It also adds a paralell implementation of `SparsePolynomial.evaluate()`.

closes: #116 

---

write a little note why.

- [x] Wrote unit tests


Pratyush:
Hmm is there a place where the parallel evaluation of sparse poly would help? The use case I had in mind was polynomials similar to the vanishing polynomial for multiplicative subgroups, where parallel evaluation would definitely hurt

ValarDragon:
Ah thats fair. I was thinking that we should probably special case regardless the case where all indices are powers of two, to not repeat squaring terms. (This captures vanishing polynomials of the union of cosets, and the vp for binary fields)

I was at the time of writing thinking that this would be more important here at lower numbers of terms than it would for the case of dense polynomial evaluation, since each term here recomputes the entire power. 

Though I think that really that tells us that we should just compute all the squares first, and then just have additions remaining for each term. (And then as shown in #119, not really a need to parallelize the additions)

I'll remove parallelization from evaluate for now, in favor of eventually doing the above types of methods.

poly/src/domain/radix2.rs
```diff
@@ -225,11 +228,13 @@ pub(crate) fn serial_radix2_fft<T: DomainCoeff<F>, F: FftField>(a: &mut [T], ome
     }
 
     let mut m = 1;
-    for _ in 0..log_n {
+    for _i in 1..=log_n {
```
Pratyush, line 231: Hmm so the old code was subtly wrong?

Pratyush: Or no, because we didn't use the index anywhere.

ValarDragon: It wasn't wrong, but writing the invariant to understand the code was awkward with the prior bounds

Pratyush (approved):

PR #124, opened 2020-12-09
Use `Shl` and `Shr` traits for `BigInteger` s

Pratyush:
## Description

Uses `Shl(Assign)` and `Shr(Assign)` traits for big-integer shifting operations

While this is based on #96, I could rebase it on master. (I just forgot to switch branches before making changes lol)

---

write a little note why.

- [x] Wrote unit tests (existing tests should catch breakage)

### DIFF
--- a/ec/src/msm/variable_base.rs
+++ b/ec/src/msm/variable_base.rs
@@ -47,11 +47,9 @@ impl VariableBaseMSM {
                                 res.add_assign_mixed(base);
                             }
                         } else {
-                            let mut scalar = scalar;
-
                             // We right-shift by w_start, thus getting rid of the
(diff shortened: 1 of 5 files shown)


Pratyush:
Hmm after doing some investigation, I've found that standard unsigned integers actually panic in debug mode if there's an overflow when left-shifting, or underflow when right-shifting. Our BigIntegers, on the other hand, do not, explicitly return zero in these cases.  The Rust standard library provides the `core::num::Wrapping` type which is supposed to provide this behaviour, but it's not clear to me whether this is necessary for us. Any thoughts? @kobigurk this relates to the findings in the audit commissioned by Celo.

PR #126, opened 2020-12-11
Use `ark_ff::batch_inversion` for point normalization

Pratyush:
## Description

Reuses existing batch inversion code for batch normalization. Enables better parallelization, at the expense of some more allocation.

---

write a little note why.

- [x] Wrote unit tests

### DIFF
--- a/ec/src/models/twisted_edwards_extended.rs
+++ b/ec/src/models/twisted_edwards_extended.rs
@@ -444,38 +444,12 @@ impl<P: Parameters> ProjectiveCurve for GroupProjective<P> {
     }
 
     fn batch_normalization(v: &mut [Self]) {
-        // Montgomery’s Trick and Fast Implementation of Masked AES
-        // Genelle, Prouff and Quisquater
-        // Section 3.2
-
-        // First pass: compute [a, ab, abc, ...]
-        let mut prod = Vec::with_capacity(v.len());
-        let mut tmp = P::BaseField::one();
-        for g in v.iter_mut()
-            // Ignore normalized elements
+        let mut z_s = v
+            .iter()
             .filter(|g| !g.is_normalized())
-        {
-            tmp *= &g.z;
-            prod.push(tmp);
-        }
-
-        // Invert `tmp`.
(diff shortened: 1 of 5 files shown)


ec/src/models/twisted_edwards_extended.rs
```diff
@@ -444,51 +444,19 @@ impl<P: Parameters> ProjectiveCurve for GroupProjective<P> {
...
         // Perform affine transformations
-        v_iter.filter(|g| !g.is_normalized()).for_each(|g| {
-            g.x *= &g.z; // x/z
-            g.y *= &g.z;
-            g.t *= &g.z;
-            g.z = P::BaseField::one(); // z = 1
-        });
+        ark_std::cfg_iter_mut!(v)
+            .zip(z_s)
+            .filter(|(g, _)| !g.is_normalized())
```
ValarDragon, line 453: Should the filter and zip be swapped?

ValarDragon: Oh this code is different than the SW case, where this one doesn't filter what goes into the batch inversion.

ValarDragon: Can we make the two cases the same? (Either filter out the already normalized elements in both, or apply the batch inversion to everything?)

Pratyush: It wasn't compiling for some reason in that order

ValarDragon: Oh no, this is just something that changed after the commit, ok this makes sense.

ValarDragon (approved):
LGTM, glad we'll get parallel batch inversion speedups here!

I wonder if we should eventually restructure this to use split the vector into chunks, and operate on each chunk in parallel. (Using a serial batch inversion on that chunk)
That would keep the data being mutated in the final loop thread-local

Pratyush:
I think that's possible if we don't unify the logic with `batch_inversion` (essentially if we duplicate the logic in `batch_inversion`)

ValarDragon:
Yeah I think we would have to duplicate that. We probably should move that standard chunk setup code to an ark-std macro

PR #127, opened 2020-12-11
Merge celo/zexe w/ arkworks

jon-chuang:
Features:
- Batch Affine ops
- Batch and ordinary w-NAF
- GLV framework. GLV precomputation script.
- GLV impl for both batch affine and projective mul
- Batch bucket addition tree
- Batched MSM (1.4-1.7x bump)
- Batch subgroup verification based on addition to random buckets
- 12-limb .s asm impl for montgomery mul and constant-time add/sub
~~2D test matrix feature-gating with axes (curve, test type)~~
~~Refactor tests to use common template~~
~~Code timing instrumentation with fine grained control of functions and callers~~

Combined speedup (with w-NAF) for:

- Batched mul (with GLV): 2.3-2.7x.
- Non-batch projective mul (with GLV): 2.1x.
- Without GLV: 1.65-1.8x and 1.4x resp.

Speedup for subgroup verification:
- Over old impl 25-80x.
- Over new batch mul: 10-30x.

Spec sheet (also included in code comments): https://hackmd.io/@celo/S10UeWUBD

I am not willing to break this PR into smaller PRs as that was already done in Zexe, and now I have had to repeat work to merge again to arkworks, so it's too much work, no thank you.

TODOs:
- Add glv generation script.
- Sister PR for curves which implement glv, cuda, test feature gating, tests for new features etc.
- Feature gate for BigInt768 asm (.s rather than ff-asm).

---

Before we can merge this PR, please make sure that all the following items have been
checked off. If any of the checklist items are not applicable, please leave them but
write a little note why.

- [ ] Targeted PR against correct branch (main)
- [X] Linked to Github issue with discussion and accepted design OR have an explanation in the PR that describes this work.
- [X] Wrote unit tests
- [X] Updated relevant documentation in the code
- [ ] Added a relevant changelog entry to the `Pending` section in `CHANGELOG.md`
- [ ] Re-reviewed `Files changed` in the Github PR explorer

PR #130, opened 2020-12-13
Celo: Serialization changes

ValarDragon:
## Description

Copies celo's serialization changes from #127

The key change is it adds `serialize_uncompressed_unchecked`.

Currently the implication of `serialize_unchecked` is to serialize in the uncompressed form, so this PR makes sense in the context of needing to serialize unchecked, compressed data. 

The usecase for that is not clear to me, as the check needed for deserializing an affine curve point on a prime order curve is effectively the same as getting the affine representation of the point. (We currently don't implement arithmetic in montgomery form AFAIK)

Is the usecase for skipping subgroup checks for compressed curve points?

@jon-chuang, @Pratyush, is the above correct for where serializing unchecked compressed comes in handy?

I'll handle fixing the merge conflicts, and copying over the changes to make elliptic curves implement this if we agree on the need for the change.

We also have the question of whether the default should be for `serialize_unchecked` to be compressed or uncompressed. The current form of this PR is a breaking change that will fail silently in many places, by changing the implementation of `serialize_unchecked`.

Not sure what the best API to be using here right now is.

---

write a little note why.

- [ ] Wrote unit tests

### DIFF
--- a/serialize/src/lib.rs
+++ b/serialize/src/lib.rs
@@ -16,6 +16,7 @@ use ark_std::{
     borrow::{Cow, ToOwned},
     collections::{BTreeMap, BTreeSet},
     convert::TryFrom,
+    vec,
     vec::Vec,
 };
 pub use error::*;
@@ -66,6 +67,7 @@ pub trait ConstantSerializedSize: CanonicalSerialize {
 ///
 /// If your code depends on `algebra` instead, the example works analogously
 /// when importing `algebra::serialize::*`.
+
 pub trait CanonicalSerialize {
     /// Serializes `self` into `writer`.
     /// It is left up to a particular type for how it strikes the
@@ -82,17 +84,26 @@ pub trait CanonicalSerialize {
 
     fn serialized_size(&self) -> usize;
 
+    /// Serializes `self` into `writer` with compression, and without
+    /// performing validity checks. Should be used *only* when there is no
+    /// danger of adversarial manipulation of the output.
+    #[inline]
+    fn serialize_unchecked<W: Write>(&self, writer: W) -> Result<(), SerializationError> {
+        self.serialize(writer)
+    }
+
     /// Serializes `self` into `writer` without compression.
     #[inline]
     fn serialize_uncompressed<W: Write>(&self, writer: W) -> Result<(), SerializationError> {
         self.serialize(writer)
     }
 
-    /// Serializes `self` into `writer` without compression, and without
-    /// performing validity checks. Should be used *only* when there is no
-    /// danger of adversarial manipulation of the output.
+    /// Serializes `self` into `writer` without compression.
     #[inline]
-    fn serialize_unchecked<W: Write>(&self, writer: W) -> Result<(), SerializationError> {
+    fn serialize_uncompressed_unchecked<W: Write>(
+        &self,
+        writer: W,
+    ) -> Result<(), SerializationError> {
         self.serialize_uncompressed(writer)
     }
 
@@ -134,6 +145,13 @@ pub trait CanonicalDeserialize: Sized {
     /// Reads `Self` from `reader`.
     fn deserialize<R: Read>(reader: R) -> Result<Self, SerializationError>;
 
+    /// Reads `self` from `reader` with compression, and without performing
+    /// validity checks. Should be used *only* when the input is trusted.
+    #[inline]
+    fn deserialize_unchecked<R: Read>(reader: R) -> Result<Self, SerializationError> {
+        Self::deserialize(reader)
+    }
+
     /// Reads `Self` from `reader` without compression.
     #[inline]
     fn deserialize_uncompressed<R: Read>(reader: R) -> Result<Self, SerializationError> {
@@ -143,7 +161,7 @@ pub trait CanonicalDeserialize: Sized {
     /// Reads `self` from `reader` without compression, and without performing
     /// validity checks. Should be used *only* when the input is trusted.
     #[inline]
-    fn deserialize_unchecked<R: Read>(reader: R) -> Result<Self, SerializationError> {
+    fn deserialize_uncompressed_unchecked<R: Read>(reader: R) -> Result<Self, SerializationError> {
         Self::deserialize_uncompressed(reader)
     }
 }
@@ -211,16 +229,29 @@ impl CanonicalDeserialize for usize {
     }
 }
 
-impl<T: CanonicalSerialize> CanonicalSerialize for [T] {
-    #[inline]
-    fn serialize<W: Write>(&self, mut writer: W) -> Result<(), SerializationError> {
-        let len = self.len() as u64;
-        len.serialize(&mut writer)?;
-        for item in self.iter() {
-            item.serialize(&mut writer)?;
-        }
-        Ok(())
+macro_rules! impl_serialize_for_slice {
+    ($($name:ident),*) => {
+        $(
+            #[inline]
+            fn $name<W: Write>(&self, mut writer: W) -> Result<(), SerializationError> {
+                let len = self.len() as u64;
+                len.serialize(&mut writer)?;
+                for item in self.iter() {
+                    item.$name(&mut writer)?;
+                }
+                Ok(())
+            }
+        )*
     }
+}
+
+impl<T: CanonicalSerialize> CanonicalSerialize for [T] {
+    impl_serialize_for_slice!(
+        serialize,
+        serialize_unchecked,
+        serialize_uncompressed,
+        serialize_uncompressed_unchecked
+    );
 
     #[inline]
     fn serialized_size(&self) -> usize {
@@ -228,26 +259,6 @@ impl<T: CanonicalSerialize> CanonicalSerialize for [T] {
             .iter()
             .map(|item| item.serialized_size())
             .sum::<usize>()
-    }
-
-    #[inline]
-    fn serialize_uncompressed<W: Write>(&self, mut writer: W) -> Result<(), SerializationError> {
-        let len = self.len() as u64;
-        len.serialize(&mut writer)?;
-        for item in self.iter() {
-            item.serialize_uncompressed(&mut writer)?;
-        }
-        Ok(())
-    }
-
-    #[inline]
-    fn serialize_unchecked<W: Write>(&self, mut writer: W) -> Result<(), SerializationError> {
-        let len = self.len() as u64;
-        len.serialize(&mut writer)?;
-        for item in self.iter() {
-            item.serialize_unchecked(&mut writer)?;
-        }
-        Ok(())
     }
 
     #[inline]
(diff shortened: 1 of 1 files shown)


Pratyush:
> Is the usecase for skipping subgroup checks for compressed curve points?

Yes, this is the intended usecase, I believe. I’m not sure what the overhead of decompression vs subgroup checks is for common curves

Pratyush:
@jon-chuang Could you elaborate further on the applications/potential use cases for this? Isn't taking the square root pretty expensive?

jon-chuang:
Hi, yes, the use case is indeed to skip the subgroup check. We can use a batched subgroup check instead which is about 30-75x faster, iirc.

jon-chuang:
It would be nice to specialise the impl for vec<T> or [T] where T: AffineCurve etc. to perform the batch subgroup check. Perhaps a change for the future.

Pratyush:
Hmm since it's useful, I propose renaming `(de)serialize_unchecked` to `(de)serialize_compressed_unchecked`. I am not undecided on whether we should also rename `serialize` to `serialize_compressed` in the process. WDYT @ValarDragon @jon-chuang @kobigurk?

jon-chuang:
Upon further thought, I personally dont like the idea of having checked and unchecked in the API, and am more in favour of one of the two solutions:
- Remove such checks entirely, then the responsibility/manner of checking falls on the programmer
- create a trait for each type representing the batch version of checking functions, with a default impl doing nothing, and impl for curves doing a batch check

Pratyush:
> Remove such checks entirely, then the responsibility/manner of checking falls on the programmer

I feel like this is probably going to lead to protocol problems elsewhere, and furthermore breaks the contracts of the curve types (which assume that the input is in the prime order subgroup)

Pratyush:
We could try playing around with adding a default parameter to the `serialize` and `serialize_uncompressed` methods, so that by default it performs the checks, but if you change the default type parameter to a different one (which avoids the checks), then it does not perform on-curve checks.

jon-chuang:
Hmm yeah, I guess the solution proposed is meant to allow for minimal breakage. 

I guess compression and data integrity checks do form a reasonable quadrant on the type of (de)serialisation one would like generically speaking.

Would you mind giving an example of the type parameter method?

ValarDragon:
we could separate the unchecked serialization methods into its own trait, that is auto-implemented if canonical serialize is implemented

ValarDragon:
If something is unchecked, the API should definitely make it clear imo

ValarDragon:
Im in favor of the approach of having 4 methods
```
* serialize_compressed
* serialize_uncompressed
* serialize_compressed_unchecked
* serialize_uncompressed_unchecked
```

jon-chuang:
> Hmm since it's useful, I propose renaming `(de)serialize_unchecked` to `(de)serialize_compressed_unchecked`. I am undecided on whether we should also rename `serialize` to `serialize_compressed` in the process. WDYT @ValarDragon @jon-chuang @kobigurk?

Wrt to this question specifically, are you proposing to have e.g. `deserialize_compressed_..` alongside `deserialize_uncompressed..`? Shouldn't one be a default, so e.g. we retain just `deserialize`?

jon-chuang:
> Im in favor of the approach of having 4 methods
> 
> ```
> * serialize_compressed
> * serialize_uncompressed
> * serialize_compressed_unchecked
> * serialize_uncompressed_unchecked
> ```
> 
> I think its fine to expose the four options here, and let downstream applications decide which method they want, and which methods they want to expose.

Hmm, well barring curves, most types do not have an option to compress, so shouldn't we have a default that leads to minimal breakage?

Pratyush:
Here's some example code, that allows incorporating the batch checks as well:
```rust
trait CanonicalDeserialize {
	Specifies how the check should be performed
	type Validator: Checker<Self>;
	
	fn deserialize<R: Read, C: ShouldCheck = AlwaysCheck>(reader: R) -> Self {
		// get item from R
		if C::should_check() {
			assert!(Self::Validator::check(&[item]));
		}
		item
	}

	
	fn deserialize_uncompressed<R: Read, C: ShouldCheck = AlwaysCheck>(reader: R) -> Self {
		// get uncompressed item from R
		if C::should_check() {
			assert!(Self::Validator::check(&[item]));
		}
		item
	}
}

/// Actually performs the check
trait Checker<T> {
	fn check(&[T]) -> bool;
}

trait ShouldCheck {
	fn should_check() -> bool;
}

struct AlwaysCheck;

impl ShouldCheck for AlwaysCheck {
	fn should_check() -> bool {
		true
	}
}

struct NoCheck;

impl ShouldCheck for NoCheck {
	fn never_check() -> bool {
		false
	}
}
```

Here's how you would use it to do both checked and unchecked serialization:

```rust
let item = T::deserialize(&mut reader)?;
let item = T::deserialize::<_, NoCheck>(&mut reader)?;
```

and here's how you would use the `Checker` API to implement batch checking without specialization


```rust

impl<T: CanonicalDeserialize> CanonicalDeserialize for Vec<T> {
	type Validator = VecChecker<T>;
	
	fn deserialize<R: Read, C: ShouldCheck = AlwaysCheck>(reader: R) -> Self {
		// get item from R
		if C::should_check() {
			assert!(T::Validator::check(&[item]));
		}
		item
	}
}

struct VecChecker<T>(PhantomData<T>);
impl<T: CanonicalDeserialize> Checker<Vec<T>> for VecChecker<T> {
	fn check(items:&[Vec<T>]) -> bool {
		for item in items {
			T::check(item.as_slice())
		}
	}
}
```
The idea would be to have different implementations of checkers for different types.

Pratyush:
Note that the changes relating to the `Validator` associated type are completely orthogonal to the changes related to the `ShouldCheck` trait; we could adopt either one.

jon-chuang:
> Here's some example code, that allows incorporating the batch checks as well:
> 
> ```rust
> trait CanonicalDeserialize {
> 	Specifies how the check should be performed
> 	type Validator: Checker<Self>;
> 	
> 	fn deserialize<R: Read, C: ShouldCheck = AlwaysCheck>(reader: R) -> Self {
> 		// get item from R
> 		if C::should_check() {
> 			assert!(Self::Validator::check(&[item]));
> 		}
> 		item
> 	}
> 
> 	
> 	fn deserialize_uncompressed<R: Read, C: ShouldCheck = AlwaysCheck>(reader: R) -> Self {
> 		// get uncompressed item from R
> 		if C::should_check() {
> 			assert!(Self::Validator::check(&[item]));
> 		}
> 		item
> 	}
> }
> 
> /// Actually performs the check
> trait Checker<T> {
> 	fn check(&[T]) -> bool;
> }
> 
> trait ShouldCheck {
> 	fn should_check() -> bool;
> }
> 
> struct AlwaysCheck;
> 
> impl ShouldCheck for AlwaysCheck {
> 	fn should_check() -> bool {
> 		true
> 	}
> }
> 
> struct NoCheck;
> 
> impl ShouldCheck for NoCheck {
> 	fn never_check() -> bool {
> 		false
> 	}
> }
> ```
> 
> Here's how you would use it to do both checked and unchecked serialization:
> 
> ```rust
> let item = T::deserialize(&mut reader)?;
> let item = T::deserialize::<_, NoCheck>(&mut reader)?;
> ```
> 
> and here's how you would use the `Checker` API to implement batch checking without specialization
> 
> ```rust
> impl<T: CanonicalDeserialize> CanonicalDeserialize for Vec<T> {
> 	type Validator = VecChecker<T>;
> 	
> 	fn deserialize<R: Read, C: ShouldCheck = AlwaysCheck>(reader: R) -> Self {
> 		// get item from R
> 		if C::should_check() {
> 			assert!(T::Validator::check(&[item]));
> 		}
> 		item
> 	}
> }
> 
> struct VecChecker<T>(PhantomData<T>);
> impl<T: CanonicalDeserialize> Checker<Vec<T>> for VecChecker<T> {
> 	fn check(items:&[Vec<T>]) -> bool {
> 		for item in items {
> 			T::check(item.as_slice())
> 		}
> 	}
> }
> ```
> 
> The idea would be to have different implementations of checkers for different types.

I like this strategy as it exposes optional fine grained control with fewer breaking changes and more separation of duties.

jon-chuang:
Instead of the above approach to `Vec`, I propose instead to have 

```
trait Checker<T> {
    fn check(value: T) -> Result;
    fn check_batch(values: &[T]) -> Result {
        for value in values {
            Self::check(value)?;
        }
        Ok(())
    }
}
```

So we have a default batch impl, which can be overriden for types.

jon-chuang:
Hi, @Pratyush @ValarDragon, could I get permission to push to non-protected branches?

jon-chuang:
Btw, where is `serialize_derive` actually used? I can't seem to find it used anywhere. For instance, `derive_canonical_deserialize` isn't.

weikengchen:
The `serialize-derive` repo is often used in a project by adding a dependency on `serialize` with the feature `derive`.
https://github.com/arkworks-rs/algebra/blob/master/serialize/src/lib.rs#L19

This allows the Rust to derive the canonical serialization and deserialization.
e.g., https://github.com/arkworks-rs/poly-commit/blob/master/src/data_structures.rs#L107

weikengchen:
By the way, thanks a lot for the serialization system by Celo. It has helped one of my projects significantly!

jon-chuang:
Hi @weikengchen thanks, I guess it's not used in algebra itself, with arrays and simple structs. I see that it's necessary for cases where there are nested structs. 

That's great, but according to the above, though, the API seems like it's about to change again.

jon-chuang:
@Pratyush unfortunately it seems that your strategy doesn't work, becuase:
`defaults for type parameters are only allowed in `struct`, `enum`, `type`, or `trait` definitions.`
So I'm still looking for a nice workaround to specify default behaviour without introducing breaking changes and/or clutter.

jon-chuang:
@weikengchen actually with regards to your example https://github.com/arkworks-rs/poly-commit/blob/master/src/data_structures.rs#L107, it does seem one can get rid of the ugly code bloat by implementing canonical (de)serialize for the `String` type.

weikengchen:
Yes. I will take a look!

jon-chuang:
Things I am currently unsure about: how to impl derive to incorporate `(de)serialize_uncompressed_unchecked`.

weikengchen:
Is the problem due to whether it should default to `(de)serialize_uncompressed` or `(de)serialize_unchecked`?

Pratyush:
btw Jon, an alternative to prevent downstream users from having to implement all methods manually would be to have a "Mode", as outlined in https://github.com/arkworks-rs/algebra/issues/145

The idea is to have a mode enum:

```rust
enum Compress {
	Yes,
	No,
}
enum Validate {
	Yes,
	No
}
```
And this enum is passed into a single method

```
trait Serialize {
	fn serialize_with_mode<R: Read>(self, reader: R, compression: Compression, validation: Validation) -> ioResult;
	
	fn serialize_compressed(...) {
		self.serialize_with_mode(reader, Compress::Yes, Validate::Yes)
	}
	
	fn serialize_compressed_unchecked(...) {
		self.serialize_with_mode(reader, Compress::Yes, Validate::No)
	}
	
	fn serialize_uncompressed(...) {
		self.serialize_with_mode(reader, Compress::No, Validate::Yes)
	}
	
	fn serialize_uncompressed_unchecked(...) {
		self.serialize_with_mode(reader, Compress::No, Validate::No)
	}
}
```

So users only have to implement the single method, and everything else is implemented by default?

This should also make writing the `Derive` macro implementation easier.

Pratyush:
Incorporating the ideas above, let me propose the following design:
```rust
pub enum Compress {
	Yes,
	No,
}
pub enum Validate {
	Yes,
	No,
}

pub trait Valid {
	fn check(&self) -> Result<(), SerializationError>;
	fn batch_check(batch: &[self]) -> Result<(), SerializationError> {
		for item in batch {
			item.check()?;
		}
		Ok(())
	}
}

pub trait CanonicalSerialize {
	fn deserialize_with_mode<W: Write>(writer: W, compress: Compress) -> Result<(), SerializationError> {
		// write item to W, handling compression as per `compress`
	}
	
	fn serialize_compressed(...) {
		self.serialize_with_mode(reader, Compress::Yes)
	}
	
	fn serialize_uncompressed(...) {
		self.serialize_with_mode(reader, Compress::No)
	}
}

pub trait CanonicalDeserialize: Valid {
	fn deserialize_with_mode<R: Read>(reader: R, compress: Compress, validate: Validate) -> Self {
		// get item from R, handling compression as per `compress`
		if let Validate::Yes = validate {
			Self::check(&item)?
		}
		item
	}
	
	fn serialize_compressed(...) {
		self.serialize_with_mode(reader, Compress::Yes, Validate::Yes)
	}
	
	fn serialize_compressed_unchecked(...) {
		self.serialize_with_mode(reader, Compress::Yes, Validate::No)
	}
	
	fn serialize_uncompressed(...) {
		self.serialize_with_mode(reader, Compress::No, Validate::Yes)
	}
	
	fn serialize_uncompressed_unchecked(...) {
		self.serialize_with_mode(reader, Compress::No, Validate::No)
	}
}

impl<T: Valid> CanonicalDeserialize for Vec<T> {
	fn deserialize_with_mode<R: Read>(reader: R, compress: Compress, validate: Validate) -> Self {
		// get items: Vec<T> from R, handling compression as per `compress`
		if let Validate::Yes = validate {
			T::batch_check(&items)?
		}
		items
	}
}
```

Issue #133, opened 2020-12-14
`ark-ff` fails to compile in release mode


ryanleh:
Currently, `ark-ff` compiles in dev mode but fails in release mode. The breakage appears to be caused by https://github.com/arkworks-rs/algebra/commit/4cbe4625a35623b39bb39eeb2e28d966c87d5d53: rolling back to https://github.com/arkworks-rs/algebra/commit/61bde284e4ed1e61cb0e86df0d5ed73b8e6dfd45 resolves the issue.

`cargo build --release`
```
error: unused import: `crate::fields::LegendreSymbol::QuadraticNonResidue`                                                                                                                     
   --> ff/src/fields/arithmetic.rs:227:13                                                                                       
    |                                                                                                                                                                                                                                                                                                                                                                                          227 |         use crate::fields::LegendreSymbol::QuadraticNonResidue;                                                                                                                          
    |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^                                                                                                                                                                                                                                                                                                                               |                                                                                         
   ::: ff/src/fields/models/mod.rs:20:1                                                                                         
    |                                                                                                                                                                                                                                                                                                                                                                                          20  | impl_Fp!(Fp256, Fp256Parameters, BigInteger256, BigInteger256, 4);                                                                                                                                                                                                                                                                                                                       
    | ------------------------------------------------------------------ in this macro invocation                                       
    |                                                                                                                                                                                                                                                                                                                                                                                          note: the lint level is defined here                                                          
   --> ff/src/lib.rs:4:5                                                                                                        
    |                                                                                                                                   
4   |     unused,                                                                                                                                                                                                                                                                                                                                                                                  |     ^^^^^^                                                                                                                                                                                                                                                                                                                                                                               
    = note: `#[deny(unused_imports)]` implied by `#[deny(unused)]`                                                                                                                                                                                                                                                                                                                                 = note: this error originates in a macro (in Nightly builds, run with -Z macro-backtrace for more info)                                                                                    
                                                                                                                                        
error: unused import: `crate::fields::LegendreSymbol::QuadraticNonResidue`                                                      
   --> ff/src/fields/arithmetic.rs:227:13                                                                                       
    |                                                                                                                                                                                                                                                                                                                                                                                          227 |         use crate::fields::LegendreSymbol::QuadraticNonResidue;                                                                                                                          
[...]

PR #138, opened 2020-12-15
Convert warnings to be warnings

ValarDragon:
## Description

Convert what we previously denied to just be warnings

cref: https://github.com/arkworks-rs/algebra/issues/133#issuecomment-744632853

---

write a little note why.

- [x] Wrote unit tests
- [ ] Added a relevant changelog entry to the `Pending` section in `CHANGELOG.md` - I don't think this needs one?

### DIFF
--- a/ec/src/lib.rs
+++ b/ec/src/lib.rs
@@ -1,9 +1,5 @@
 #![cfg_attr(not(feature = "std"), no_std)]
-#![warn(
-    future_incompatible,
-    nonstandard_style,
-    rust_2018_idioms
-)]
+#![warn(future_incompatible, nonstandard_style, rust_2018_idioms)]
 #![forbid(unsafe_code)]
 #![allow(
     clippy::op_ref,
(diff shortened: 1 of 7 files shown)


ec/src/lib.rs
```diff
@@ -1,11 +1,5 @@
 #![cfg_attr(not(feature = "std"), no_std)]
-#![deny(
-    warnings,
-    unused,
-    future_incompatible,
-    nonstandard_style,
-    rust_2018_idioms
-)]
+#![warn(future_incompatible, nonstandard_style, rust_2018_idioms)]
```
Pratyush, line 2: Is the `unused` warn by default? Otherwise, we should keep it here.

ValarDragon: A couple of the unused warns are allowed by default (namely 
```
unused_crate_dependencies
unused_extern_crates
unused_import_braces
unused_lifetimes
unused_qualifications
unused_results
```
)
I'll `unused` back to warn

PR #141, opened 2020-12-16
Add `Fp64`

Pratyush:
## Description

Adds `Fp64`.

---

write a little note why.

- [ ] Wrote unit tests

### DIFF
--- a/ff/src/fields/arithmetic.rs
+++ b/ff/src/fields/arithmetic.rs
@@ -240,17 +240,6 @@ macro_rules! sqrt_impl {
         let mut b = x * &w;
 
         let mut v = $P::TWO_ADICITY as usize;
-        // t = self^t
-        #[cfg(debug_assertions)]
-        {
-            let mut check = b;
-            for _ in 0..(v - 1) {
-                check.square_in_place();
-            }
-            if !check.is_one() {
-                panic!("Input is not a square root, but it passed the QR test")
-            }
-        }
 
         while !b.is_one() {
             let mut k = 0usize;
--- a/ff/src/fields/arithmetic.rs
+++ b/ff/src/fields/arithmetic.rs
@@ -236,10 +236,14 @@ macro_rules! impl_prime_field_from_int {
             }
         }
     };
-    ($field: ident, $int: ident, $params: ident) => {
+    ($field: ident, $int: ident, $params: ident, $limbs:expr) => {
         impl<P: $params> From<$int> for $field<P> {
             fn from(other: $int) -> Self {
-                Self::from_repr(P::BigInt::from(u64::from(other))).unwrap()
+                if $limbs == 1 {
+                    Self::from_repr(P::BigInt::from(u64::from(other) % P::MODULUS.0[0])).unwrap()
+                } else {
+                    Self::from_repr(P::BigInt::from(u64::from(other))).unwrap()
+                }
             }
         }
     };
--- a/ff/src/fields/arithmetic.rs
+++ b/ff/src/fields/arithmetic.rs
@@ -221,19 +221,29 @@ macro_rules! impl_prime_field_standard_sample {
 }
 
 macro_rules! impl_prime_field_from_int {
-    ($field: ident, u128, $params: ident) => {
+    ($field: ident, u128, $params: ident, 1) => {
         impl<P: $params> From<u128> for $field<P> {
             fn from(other: u128) -> Self {
+                let mut default_int = P::BigInt::default();
+                default_int.0[0] = (other % u128::from(P::MODULUS.0[0])) as u64;
+                Self::from_repr(default_int).unwrap()
+            }
+        }
+    };
+    ($field: ident, u128, $params: ident, $limbs:expr) => {
+        impl<P: $params> From<u128> for $field<P> {
+            fn from(other: u128) -> Self {
+                let mut default_int = P::BigInt::default();
                 let upper = (other >> 64) as u64;
                 let lower = ((other << 64) >> 64) as u64;
-                let mut default_int = P::BigInt::default();
+
                 default_int.0[0] = lower;
                 default_int.0[1] = upper;
                 Self::from_repr(default_int).unwrap()
             }
         }
     };
-    ($field: ident, $int: ident, $params: ident) => {
+    ($field: ident, $int: ident, $params: ident, $limbs:expr) => {
         impl<P: $params> From<$int> for $field<P> {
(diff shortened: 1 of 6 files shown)


ff/src/fields/arithmetic.rs
```diff
@@ -80,22 +94,28 @@ macro_rules! impl_field_into_repr {
...
-            let mut all_bits_set = P::MODULUS.0[$limbs - 1] == !0 - (1 << 63);
-            for i in 1..$limbs {
-                all_bits_set &= P::MODULUS.0[$limbs - i - 1] == core::u64::MAX;
-            }
-            let _no_carry: bool = !(first_bit_set || all_bits_set);
-
             #[cfg(use_asm)]
             #[allow(unsafe_code, unused_mut)]
             {
+                // Checking the modulus at compile time
```
Pratyush, line 111: Drive-by refactor that moves code used only when `use_asm` is true inside the conditionally-compiled portion.

ff/src/fields/arithmetic.rs
```diff
@@ -5,12 +5,26 @@
 /// `P::MODULUS` has (a) a non-zero MSB, and (b) at least one
 /// zero bit in the rest of the modulus.
 macro_rules! impl_field_mul_assign {
+    (1) => {
+        fn mul_assign(&mut self, other: &Self) {
+            #[inline(always)]
+            fn mont_reduce(&mut self, mul_result: u128) {
```
Pratyush, line 11: This is just a specialization of the standard algorithm to the case where there is just one limb.

ff/src/fields/arithmetic.rs
```diff
@@ -120,8 +121,11 @@ macro_rules! impl_field_square_in_place {
             }
             r[$limbs * 2 - 1] = r[$limbs * 2 - 2] >> 63;
             for i in 0..$limbs {
-                r[$limbs * 2 - 2 - i] =
-                    (r[$limbs * 2 - 2 - i] << 1) | (r[$limbs * 2 - 3 - i] >> 63);
+                let subtractor = (2 * ($limbs - 1usize))
+                    .checked_sub(i + 1)
+                    .map(|index| r[index])
+                    .unwrap_or(0);
```
Pratyush, line 127: Required to avoid underflowing. I have confirmed locally that it doesn't affect performance.

ff/src/fields/arithmetic.rs
```diff
@@ -201,14 +205,20 @@ macro_rules! impl_prime_field_standard_sample {
...
-                default_int.0[1] = upper;
+                if $limbs == 1 {
+                    default_int.0[0] = (other % u128::from(P::MODULUS.0[0])) as u64;
+                } else {
+                    let upper = (other >> 64) as u64;
+                    let lower = ((other << 64) >> 64) as u64;
+                    let limbs = [lower, upper];
+                    for (cur, other) in default_int.0.iter_mut().zip(&limbs) {
+                        *cur = *other;
+                    }
```
Pratyush, line 220: Again, required to satisfy the compiler.

ff/src/fields/arithmetic.rs
```diff
@@ -11,6 +11,7 @@ macro_rules! impl_field_mul_assign {
         fn mul_assign(&mut self, other: &Self) {
             // Checking the modulus at compile time
             let first_bit_set = P::MODULUS.0[$limbs - 1] >> 63 != 0;
+            #[allow(unused_mut)]
```
ValarDragon, line 14:
```suggestion
            // $limbs can be 1, hence we can run into a case with an unused mut.
            #[allow(unused_mut)]
```

ff/src/fields/arithmetic.rs
```diff
@@ -22,7 +23,7 @@ macro_rules! impl_field_mul_assign {
                 #[cfg(use_asm)]
                 #[allow(unsafe_code, unused_mut)]
                 {
-                    if $limbs <= 6 {
+                    if $limbs <= 6 && $limbs > 1 {
```
ValarDragon, line 26: Not related to this PR, but could there be a comment for why theres a different method for limbs > 6? Is it just that the larger limb size case isn't implemented? (Doesn't need to block this PR)

ff/src/fields/arithmetic.rs
ValarDragon: Can a comment be added for the arithmetic operations these lines are doing / intending to do? (Looks like propagating a carry bit?)

I agree this matches what was previously implemented.

ff/src/fields/arithmetic.rs
```diff
@@ -201,14 +205,20 @@ macro_rules! impl_prime_field_standard_sample {
...
-                let lower = ((other << 64) >> 64) as u64;
                 let mut default_int = P::BigInt::default();
-                default_int.0[0] = lower;
-                default_int.0[1] = upper;
+                if $limbs == 1 {
+                    default_int.0[0] = (other % u128::from(P::MODULUS.0[0])) as u64;
+                } else {
+                    let upper = (other >> 64) as u64;
+                    let lower = ((other << 64) >> 64) as u64;
+                    let limbs = [lower, upper];
```
ValarDragon, line 217:
```suggestion
                    // This is equivalent to the following, but satisfying the compiler:
                    // default_int.0[0] = lower;
                    // default_int.0[1] = upper;
                    let limbs = [lower, upper];
```

ValarDragon (commented):
LGTM, left a couple of requests for additional comments, then I think its good to merge

ff/src/fields/arithmetic.rs
Pratyush: I'm not too sure myself, but I added a comment explaining what is computed.

ff/src/fields/arithmetic.rs
Pratyush: Yes, larger limb size is not implemented

ff/src/fields/arithmetic.rs
ValarDragon: Ok cool!

Kinda worrisome if neither of us understand the code, we should file an issue to decipher & document whats going on these functions 😅

ValarDragon (approved):
LGTM

ff/src/fields/arithmetic.rs
Pratyush: Agreed, @jon-chuang should have the best state, I believe?

Pratyush:
This passes tests locally for the prime 257.

PR #147, opened 2020-12-19
WIP: Add Hash to curve trait, and shallue_van_de_woestijne hashing

ValarDragon:
## Description

This PR adds a hash to curve trait (inspired by the work in https://github.com/celo-org/celo-bls-snark-rs, and discussions with @kobigurk on relevant issues in that repository). 

Furthermore, all of the hashes in https://tools.ietf.org/html/draft-irtf-cfrg-hash-to-curve-09 are constructed from two components, a hash to the field, and a map to the curve. This PR implements traits for each of those components, and obtaining a hash to curve out of them. 

This PR will (eventually) also implement `shallue_van_de_woestijne` hashing, by implementing it as a map to curve. (This is a relatively slow, constant time hash to curve, that works for all short weierstrass curves)

Please leave comments on thoughts regarding these as traits!

Remaining TODO's:
- [ ] Complete implementing the `DefaultFieldHasher`
- [ ] Implement shallue_van_de_woestijne map to curve
- [ ] Implement curve parameter generation for shallue_van_de_woestijne
- [ ] Add a README to the hash to curve folder
- [ ] write/find test vectors for HashToField's correctness

---

write a little note why.

- [ ] Wrote unit tests


ec/src/hashing/shallue_van_de_woestijne/mod.rs
```diff
@@ -0,0 +1,16 @@
...
+    // a non-zero element of F meeting the below criteria.
+    // Let g(x) be the value of y^2, e.g. g(x) = x^3 + Ax + B
+    // 1.  g(Z) != 0 in F.
+    // 2.  -(3 * Z^2 + 4 * A) / (4 * g(Z)) != 0 in F.
+    // 3.  -(3 * Z^2 + 4 * A) / (4 * g(Z)) is square in F.
+    // 4.  At least one of g(Z) and g(-Z / 2) is square in F.
+    const Z : Self::BaseField;
+}
+
+pub struct shallue_van_de_woestijne_hasher {
```
Pratyush, line 14: We should probably write these in Rust CamelCase?

ValarDragon:
Update on my current thought on interfaces.

Theres the following HashToCurve trait:

```
/// Trait for hashing arbitrary data to a group element on an elliptic curve
pub trait HashToCurve<T: AffineCurve>: Sized {
    /// Create a new hash to curve instance, with a given domain.
    fn new(domain: &[u8]) -> Result<Self, HashToCurveError>;

    /// Produce a hash of the message, which also depends on the domain.
    /// The output of the hash is a curve point in the prime order subgroup
    /// of the given elliptic curve.
    fn hash(&self, message: &[u8]) -> Result<T, HashToCurveError>;
}
```

Then there are separate traits for hashes constructed from HashToField and a MapToCurve. This encompasses all of the hash types in the IETF draft. ( Celo's library has traits / infrastructure for hashing to a field constructed from try and increment)

There is a struct which has a HashToField and MapToCurve templated, and implements a HashToCurve. 

Implementations of MapToCurve will each be in there own crate. Should we do the same thing for `HashToField`? I imagine its the sort of thing where there will be a bunch of slightly different implementations, based on different performance / security requirements, and different methods for handling domains.

daira:
What's the motivation for focussing on SW rather than simplified SWU?

Simplified SWU applies to any curve with AB ≠ 0 *or* that has an efficient isogeny from a curve with AB ≠ 0. That applies to almost all of the curves that anyone cares about in practice, with the possible exception of BN-254. The optimizations of [Wahby–Boneh 2019](https://eprint.iacr.org/2019/403) can be applied to any such curve.

For example, a [hash to Pallas](https://github.com/zcash/pasta/blob/master/hashtocurve.sage) can be done in 578S + 141M with no inversions, if output is in Jacobian coordinates. More generally, an inversion-free, constant-time, random oracle hash-to-curve using simplified SWU (compatible with [the Internet Draft](https://www.ietf.org/archive/id/draft-irtf-cfrg-hash-to-curve-10.html)) only requires a little more than two divsqrt operations (for input N/D, find either sqrt(N/D) or sqrt(g * N/D) where g is any fixed nonsquare), and a divsqrt takes a little more than a full-width exponentiation. (There is no restriction on p for efficient divsqrt; it can be implemented generically in a constant number of operations for any 2-adicity.) SW, on the other hand, starts at 6 square root operations for a constant-time implementation of the algorithm in the Internet Draft, and I don't know how to optimize it further. (You can sacrifice constant-timeness, but then the *best* case is around the same as simplified SWU's constant-time case.)

Simplified SWU is clearly the option that the zk community should standardize on.

ValarDragon:
Definitely agreed on standardizing around simplified SWU for applicable curves!

My reasoning for focusing on SW first was that it works for all curves, seems useful to have in the library, and seems much easier to implement than simplified SWU. My primary goal for the PR was to setup the hash to curve infrastructure, not picking the map to curve to standardize on. I'd like to eventually implement / integrate implementations of simplified SWU, elligator 2 and try and increment. (With a README in the hashing folder indicating suggestions for when to use each algorithm)

weikengchen:
Just add a note. As Wahby-Boneh's paper shows, if one does not want "constant-time" but just "data-independent time", then the original SW map is not too slow than their more complicated SWU map.

Since there are non-constant-time algorithms (implemented in GMP) that compute the Legendre symbols efficiency, and the timing patterns could be removed by blinding the value with a random quadratic residue before computing its Legendre symbols efficiency.

Also, the Wahby-Boneh map is a little bit complicated to implement...

@daira Although the Wahby-Boneh map applies to any curve, finding the isogeny for each pairing-friendly curve is not a small work since we don't have one-shot scripts for doing that at this moment, and we did not have implementations for isogeny maps in the arkworks right now. 

...therefore, the general-purpose SW map might suffice...

weikengchen:
@daira Added on the Wahby-Boneh map.

There seems to be a number of curves where an efficient isogeny cannot be found, such as another BN curve that is about 446 if I remember correctly. So, the map proposed in the paper would not generalize to all the curves.

daira:
> finding the isogeny for each pairing-friendly curve is not a small work since we don't have one-shot scripts for doing that at this moment,

It's just a few lines of sage, though? https://github.com/zcash/pasta/blob/f0f7068552a3565786cb338448cb58bc36a8314a/amicable.sage#L217-L227

> and we did not have implementations for isogeny maps in the arkworks right now.

This is a somewhat more significant obstacle. I think it would be worth it though. (The maps of a given degree have the same equations modulo constants, so you could just implement the maps of prime degree up to 13, say, and that would catch most of the curves with maps you would want to use.)

weikengchen:
In another project of mine, I have initiated some attempts on the Wahby-Boneh map...

maybe I will continue from there and move the code here...

https://github.com/oblivious-file-sharing/netherite-algebra/tree/main/src/wahby_boneh_map/sage_scripts

Pratyush:
Is there anything in this PR that's not subsumed by #343 ? If not, we can close this PR.

Pratyush:
This is superseded by #343

PR #154, opened 2020-12-22
Made more consistent use of cfg_iter

rozbb:
## Description

MSM uses `cfg_iter` in some place but not in others. This makes it a bit more consistent. Unclear what the performance benefits are.

---

write a little note why.

Haven't done any of the below. This is a nit change.

- [ ] Wrote unit tests

### DIFF
--- a/ec/src/msm/variable_base.rs
+++ b/ec/src/msm/variable_base.rs
@@ -1,5 +1,5 @@
 use ark_ff::prelude::*;
-use ark_std::vec::Vec;
+use ark_std::{cfg_into_iter, cfg_iter, vec::Vec};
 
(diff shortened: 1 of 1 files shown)


rozbb:
Okay, this is actually worse than expected. The `cfg_iter` abstraction leaks, and it's necessary to explicitly feature gate this part of the function. The result is kinda gross. It's also hard to tell if this is still correct, since there are no unit tests

Pratyush:
Thanks for this. I benchmarked this, and it seems that there's no perf improvement (likely because the number of windows is usually small (< 50))

rozbb:
Gotcha, then this complication definitely isn’t worth it. Feel free to close

PR #157, opened 2020-12-29
Simplify MSM implementation, and small speedup

Pratyush:
## Description

As per the title, this PR simplifies the internal implementation of variable-base MSMs, and also provides a small speedup (by doing less work).

---

write a little note why.

- [x] Wrote unit tests

### DIFF
--- a/ec/src/msm/variable_base.rs
+++ b/ec/src/msm/variable_base.rs
@@ -13,10 +13,15 @@ impl VariableBaseMSM {
         bases: &[G],
         scalars: &[<G::ScalarField as PrimeField>::BigInt],
     ) -> G::Projective {
-        let c = if scalars.len() < 32 {
+        let size = ark_std::cmp::min(bases.len(), scalars.len());
+        let scalars = &scalars[..size];
+        let bases = &bases[..size];
+        let scalars_and_bases_iter = scalars.iter().zip(bases).filter(|(s, _)| !s.is_zero());
+
+        let c = if size < 32 {
             3
         } else {
-            super::ln_without_floats(scalars.len()) + 2
+            super::ln_without_floats(size) + 2
         };
 
         let num_bits = <G::ScalarField as PrimeField>::Params::MODULUS_BITS as usize;
@@ -24,11 +29,6 @@ impl VariableBaseMSM {
 
         let zero = G::Projective::zero();
         let window_starts: Vec<_> = (0..num_bits).step_by(c).collect();
-
-        let size = ark_std::cmp::min(bases.len(), scalars.len());
-        let scalars = &scalars[..size];
-        let bases = &bases[..size];
-        let scalars_and_bases = scalars.iter().zip(bases).filter(|(s, _)| !s.is_zero());
 
         // Each window is of size `c`.
         // We divide up the bits 0..num_bits into windows of size `c`, and
@@ -38,7 +38,9 @@ impl VariableBaseMSM {
                 let mut res = zero;
                 // We don't need the "zero" bucket, so we only have 2^c - 1 buckets
                 let mut buckets = vec![zero; (1 << c) - 1];
-                scalars_and_bases.clone().for_each(|(&scalar, base)| {
+                // This clone is cheap, because the iterator contains just a
+                // pointer and an index into the original vectors
+                scalars_and_bases_iter.clone().for_each(|(&scalar, base)| {
                     if scalar == fr_one {
                         // We only process unit scalars once in the first window.
                         if w_start == 0 {
@@ -63,6 +65,7 @@ impl VariableBaseMSM {
                     }
                 });
 
+                // Compute sum_{i in 0..num_buckets} (sum_{j in i..num_buckets} bucket[j])
                 // We could first normalize `buckets` and then use mixed-addition
                 // here, but that's slower for the kinds of groups we care about
                 // (Short Weierstrass curves and Twisted Edwards curves).
--- a/ec/src/msm/variable_base.rs
+++ b/ec/src/msm/variable_base.rs
@@ -63,6 +63,9 @@ impl VariableBaseMSM {
                     }
                 });
 
+                // We could first normalize `buckets` and then use mixed-addition
+                // here, but that's slower for the kinds of groups we care about
+                // (Short Weierstrass curves and Twisted Edwards curves).
                 let mut running_sum = G::Projective::zero();
                 buckets.into_iter().rev().for_each(|b| {
                     running_sum += &b;
--- a/ec/src/msm/variable_base.rs
+++ b/ec/src/msm/variable_base.rs
@@ -9,13 +9,10 @@ use rayon::prelude::*;
 pub struct VariableBaseMSM;
 
 impl VariableBaseMSM {
-    fn msm_inner<G: AffineCurve>(
+    pub fn multi_scalar_mul<G: AffineCurve>(
         bases: &[G],
         scalars: &[<G::ScalarField as PrimeField>::BigInt],
-    ) -> G::Projective
-    where
-        G::Projective: ProjectiveCurve<Affine = G>,
-    {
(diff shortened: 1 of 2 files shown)


ec/src/msm/variable_base.rs
```diff
@@ -36,42 +38,39 @@ impl VariableBaseMSM {
...
-                                buckets[(scalar - 1) as usize].add_assign_mixed(base);
-                            }
+                        // If the scalar is non-zero, we update the corresponding
+                        // bucket.
+                        // (Recall that `buckets` doesn't have a zero bucket.)
+                        if scalar != 0 {
+                            buckets[(scalar - 1) as usize].add_assign_mixed(base);
                         }
-                    });
-                let buckets = G::Projective::batch_normalization_into_affine(&buckets);
```
Pratyush, line 67: This is the only functional change happening here. Instead of using mixed addition, we avoid normalizing and instead just use normal addition. In the end, the overhead of normalizing + allocating is higher than the savings from mixed addition.

The new approach is also what happens in the Bellman implementation.

weikengchen:
The PR looks good. Do we need to update the CHANGELOG.md for this? It belongs to an improvement.

Pratyush:
I'll update the CHANGELOG, I guess it is a slight performance improvement.

ec/src/msm/variable_base.rs
```diff
@@ -36,42 +38,36 @@ impl VariableBaseMSM {
...
-                    .filter(|(s, _)| !s.is_zero())
-                    .for_each(|(&scalar, base)| {
-                        if scalar == fr_one {
-                            // We only process unit scalars once in the first window.
-                            if w_start == 0 {
-                                res.add_assign_mixed(base);
-                            }
-                        } else {
-                            let mut scalar = scalar;
+                scalars_and_bases.clone().for_each(|(&scalar, base)| {
```
ValarDragon, line 41: Why is this cloning the vector? Seems like this vector isn't mutated

Pratyush: It's not a vector, it's an iterator that contains only a pointer and an index, so it's cheap to clone.

ec/src/msm/variable_base.rs
```diff
@@ -28,6 +25,11 @@ impl VariableBaseMSM {
         let zero = G::Projective::zero();
         let window_starts: Vec<_> = (0..num_bits).step_by(c).collect();
 
+        let size = ark_std::cmp::min(bases.len(), scalars.len());
+        let scalars = &scalars[..size];
+        let bases = &bases[..size];
+        let scalars_and_bases = scalars.iter().zip(bases).filter(|(s, _)| !s.is_zero());
```
Pratyush, line 31: This change is because in some downstream projects, (poly-commit) algorithms that use MSM are slower if the vectors are mismatched. The `zip` should fix this, but in benchmarks it doesn't.

(For that matter, neither does this new code, so I can remove it if that's better)

ec/src/msm/variable_base.rs
ValarDragon: I see. Perhaps this should be renamed to `scalars_and_bases_iter`?

ec/src/msm/variable_base.rs
ValarDragon: This makes sense. Should a comment be added here to explain why its going with this approach?

ec/src/msm/variable_base.rs
```diff
@@ -36,42 +38,36 @@ impl VariableBaseMSM {
...
+                    }
+                });
 
                 let mut running_sum = G::Projective::zero();
-                for b in buckets.into_iter().rev() {
-                    running_sum.add_assign_mixed(&b);
-                    res += running_sum;
-                }
-
+                buckets.into_iter().rev().for_each(|b| {
```
ValarDragon, line 67: As an aside, we should probably parallelize this loop. (Will have some complexity added, to handle the running sum's updates for subsequent chunks)

Pratyush: Is that worthwhile? We're already parallelizing the outer loop (over the windows)

ValarDragon: Hrmm, probably should benchmark this. I thought its a linear number of buckets in the degree

ec/src/msm/variable_base.rs
ValarDragon: Looks to me like this is due to line 19, where `c` is set just on the scalars length? That should probably be moved down, and calculated based on size.

ec/src/msm/variable_base.rs
```diff
@@ -36,42 +38,39 @@ impl VariableBaseMSM {
...
+                            buckets[(scalar - 1) as usize].add_assign_mixed(base);
                         }
-                    });
-                let buckets = G::Projective::batch_normalization_into_affine(&buckets);
+                    }
+                });
 
+                // We could first normalize `buckets` and then use mixed-addition
+                // here, but that's slower for the kinds of groups we care about
+                // (Short Weierstrass curves and Twisted Edwards curves).
```
ValarDragon, line 68: Thanks for adding the comment! Could it be amended to also list out the operation its calculating? I believe it is 
`sum_{i in 0..num_buckets} sum_{j in i..num_buckets} bucket[j]`

ValarDragon: (Also fine if we punt documenting the logic, just figured its helpful to add in now since we're touching this code right now)

ec/src/msm/variable_base.rs
Pratyush: >  I thought its a linear number of buckets in the degree

It is, but the idea right now is that each thread gets its own window to operate over. Usually the number of windows is more than the number of threads, so the existing allocation is already fine for those cases. However if the number of threads is higher, then we could try leverage the extra threads for better parallelism; but we have to do so in a way that doesn't harm the normal mode of operation.

One idea would be to conditionally parallelize only if we have spare threads; this can be done via [rayon's `ThreadPool`](https://docs.rs/rayon/1.5.0/rayon/struct.ThreadPool.html). The idea would be allocate some number of threads to the pool (say 2), and then execute operations inside the pool. The threadpool approach would ensure that all Rayon operations inside the pool use at most two threads

ec/src/msm/variable_base.rs
Pratyush: To clarify, this was happening when `bases.len() > scalars.len()`, so the `c` calculation wasn't affected.

(But you're right, we should move the `c` calculation down anyway)

ec/src/msm/variable_base.rs
ValarDragon: Ah I see. Didn't realize its already parallelized, then I agree

ec/src/msm/variable_base.rs
ValarDragon: Also scalars should never be bigger then the number of bases, right?

Pratyush: Hm not sure if that's necessary; you could imagine a case where zero-padding makes `scalars` larger, but because the extras are zero, it doesn't affect the result

(This might be happening in `ark-poly-commit` already)

ValarDragon: Hrmm, the function doesn't really make sense in the case where its non-zero. (And in the case where there are more 0's than bases, its being inefficient in allocating them)

Regardless of if this function accepts extra zeroes, imo we should (eventually) change poly-commit to not allocate extra zeros

ValarDragon (approved):
LGTM now! Out of curiousity, any idea how much of a speedup this was?

Pratyush:
It was pretty small, less than 5% at best

(it did get slightly more pronounced with more threads, but that might have been because normalization is multi-threaded by default, so there was some cross-thread interference)

ValarDragon:
I see, we should probably expose a single core batch inversion then

PR #160, opened 2020-12-30
Fix #47

### LINKED ISSUE #47: Correctly handle serialization of curves where `BaseField::MODULUS_SIZE = m * 64  - 1`
Handle this as follows:

1) Change the [`ConstantSerializedSize` trait](https://github.com/arkworks-rs/algebra/blob/3e5802db76d851640439c113d723305b32d18b3a/serialize/src/lib.rs#L39) to use methods instead of constants:

```rust
pub trait ConstantSerializedSize {
	fn serialized_size<F: Flags>() -> usize;
	fn uncompressed_size<F: Flags>() -> usize;
}
```

2) Change the implementation of `ConstantSerializedSize` for all fields to additionally use `F::len()` when calculating the number of required bytes, and similarly for elliptic curves

cc @therealyingtong; I believe this is the reason why your tests are failing in https://github.com/arkworks-rs/curves/pull/8

also cc @paberr @kobigurk since y'all were involved in the original serialization code; does this strategy seems reasonable?

Pratyush:
## Description

closes: #47

---

write a little note why.

- [x] Wrote unit tests


ec/src/models/twisted_edwards_extended.rs
```diff
@@ -142,17 +142,15 @@ impl<P: Parameters> AffineCurve for GroupAffine<P> {
     }
 
     fn from_random_bytes(bytes: &[u8]) -> Option<Self> {
-        let x = P::BaseField::from_random_bytes_with_flags(bytes);
-        if let Some((x, flags)) = x {
-            let parsed_flags = EdwardsFlags::from_u8(flags);
+        P::BaseField::from_random_bytes_with_flags::<EdwardsFlags>(bytes).and_then(|(x, flags)| {
+            // if x is valid and is zero and only the infinity flag is set, then parse this
+            // point as infinity. For all other choices, get the original point.
```
Pratyush, line 147:
```suggestion
            // if x is valid and is zero, then parse this
            // point as infinity.
```

ff/src/fields/macros.rs
```diff
@@ -303,28 +299,49 @@ macro_rules! impl_Fp {
...
+                if F::BIT_SIZE > 8 {
+                    return None
+                } else {
+                    let mut result_bytes = [0u8; $limbs * 8 + 1];
+                    // Copy the input into a temporary buffer.
+                    result_bytes.iter_mut().zip(bytes).for_each(|(result, input)| {
+                        *result = *input;
+                    });
+                    // This mask retains everything that is below P::MODULUS_BITS
+                    let field_element_mask = (u64::MAX >> P::REPR_SHAVE_BITS).to_le_bytes();
```
Pratyush, line 312:
```suggestion
                    // This mask retains everything in the last limb
                    // that is below `P::MODULUS_BITS`.
                    let last_limb_mask = (u64::MAX >> P::REPR_SHAVE_BITS).to_le_bytes();
```

ff/src/fields/macros.rs
```diff
@@ -303,28 +299,49 @@ macro_rules! impl_Fp {
...
+                    let mut last_bytes_mask = [0u8; 9];
+                    last_bytes_mask[..8].copy_from_slice(&field_element_mask);
```
Pratyush, line 314:
```suggestion
                    last_bytes_mask[..8].copy_from_slice(&last_limb_mask);
```

serialize/src/flags.rs
```diff
@@ -1,39 +1,59 @@
 pub trait Flags: Default + Clone + Copy + Sized {
+    /// The number of bits required to encode `Self`.
+    /// This should be at most 8.
+    const BIT_SIZE: usize;
+
+    // Returns a bit mask corresponding to `self`.
+    // For example, if `Self` contains two variants, there are just two possible
+    // bit masks: `0` and `1 << 8`.
```
Pratyush, line 8:
```suggestion
    // bit masks: `0` and `1 << 7`.
```

Pratyush:
Upstream tests pass for `ark-pallas` and `ark-vesta`!

CHANGELOG.md
```diff
@@ -67,6 +76,7 @@ The main features of this release are:
 - #112 (ark-serialize) Make `bool`s checked serialization methods non-malleable.
 - #119 (ark-poly) Fix bugs in degree calculation if adding/subtracting same degree polynomials
      whose leading coefficients cancel.
+- #160 (ark-serialize, ark-ff, ark-ec) Support serializing when `MODULUS_BITS + FLAG_BITS` is greater than a multiple of 8 (fixes #47).
```
ValarDragon, line 79: errm, aren't all numbers greater than a multiple of 8 =p

Pratyush: oh lol fair

```suggestion
- #160 (ark-serialize, ark-ff, ark-ec) Support serializing when `MODULUS_BITS + FLAG_BITS` is greater than the multiple of 8 just greater than `MODULUS_BITS` (fixes #47).
```

serialize/src/flags.rs
```diff
@@ -1,39 +1,59 @@
...
+            None
+        }
     }
 }
 
 /// Flags to be encoded into the serialization.
 /// The default flags (empty) should not change the binary representation.
-#[derive(Clone, Copy)]
+#[derive(Clone, Copy, PartialEq, Eq)]
 pub enum SWFlags {
```
ValarDragon, line 57: btw, why are these flags here instead of the SWCurveModel file?

serialize/src/flags.rs
```diff
@@ -91,33 +113,21 @@ impl Flags for SWFlags {
...
     #[inline]
-    fn from_u8(value: u8) -> Self {
+    fn from_u8(value: u8) -> Option<Self> {
         let x_sign = (value >> 7) & 1 == 1;
         let is_infinity = (value >> 6) & 1 == 1;
         match (x_sign, is_infinity) {
-            (_, true) => SWFlags::Infinity,
-            (true, false) => SWFlags::PositiveY,
-            (false, false) => SWFlags::NegativeY,
+            (true, true) => None,
```
ValarDragon, line 120: Why is `true, true` invalid? Is this just to avoid there being two ways to serialize the point at infinity? If so, can there be a comment here to make it explicit that that is why?

serialize/src/flags.rs
```diff
@@ -1,39 +1,59 @@
 pub trait Flags: Default + Clone + Copy + Sized {
+    /// The number of bits required to encode `Self`.
+    /// This should be at most 8.
+    const BIT_SIZE: usize;
+
+    // Returns a bit mask corresponding to `self`.
+    // For example, if `Self` contains two variants, there are just two possible
+    // bit masks: `0` and `1 << 7`.
     fn u8_bitmask(&self) -> u8;
```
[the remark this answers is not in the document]
Pratyush: Because `Self::BIT_SIZE` is at most 8.

CHANGELOG.md
ValarDragon: that LGTM. Perhaps it should point out that this is the case for pallas/vesta curves?

ValarDragon:
I'm generally confused as to why `ConstantSerializedSize` was removed, that seems like a really useful trait to me.

For this new class of curves, presumably the serialization with flags is still constant sized, its just one byte longer than the serialization without flags?

Pratyush:
It was removed because the size depends dynamically on the flag used (and we can't do generics in constants yet). For example, if I'm serializing with `EdwardsFlags`, a 255-bit field will still fit into 32 bytes. However, if I'm serializing with `SWFlags`, a 255-bit field will fit only into 33 bytes.

serialize/src/flags.rs
Pratyush:
```suggestion
			// This is invalid because we only want *one* way to serialize
			// the point at infinity.
            (true, true) => None,
```

serialize/src/flags.rs
```diff
@@ -1,39 +1,59 @@
 pub trait Flags: Default + Clone + Copy + Sized {
```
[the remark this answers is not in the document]
Pratyush: Yes to the latter

serialize/src/lib.rs
```diff
@@ -154,19 +151,14 @@ macro_rules! impl_uint {
...
-        impl ConstantSerializedSize for $ty {
-            const SERIALIZED_SIZE: usize = core::mem::size_of::<$ty>();
-            const UNCOMPRESSED_SIZE: usize = Self::SERIALIZED_SIZE;
-        }
-
         impl CanonicalDeserialize for $ty {
             #[inline]
             fn deserialize<R: Read>(mut reader: R) -> Result<Self, SerializationError> {
-                let mut bytes = [0u8; Self::SERIALIZED_SIZE];
+                let mut bytes = [0u8; core::mem::size_of::<$ty>()];
```
ValarDragon, line 161: Is using `core::mem` safe? I thought this is the size of the in-memory representation, which could differ between architectures?

serialize/src/flags.rs
```diff
@@ -91,33 +113,21 @@ impl Flags for SWFlags {
...
-    fn len() -> usize {
-        2
-    }
 }
 
 /// Flags to be encoded into the serialization.
 /// The default flags (empty) should not change the binary representation.
-#[derive(Clone, Copy)]
+#[derive(Clone, Copy, PartialEq, Eq)]
 pub enum EdwardsFlags {
```
ValarDragon, line 131: Can there be comments for how the flags are added to the encoding? (Appears to be using the last bit of the serialization)

serialize/src/flags.rs
ValarDragon: Also can there be a comment for how flags are added to the encoding?

serialize/src/flags.rs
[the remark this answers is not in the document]
Pratyush:
```suggestion
/// Represents bits that are appended to the end of
/// a serialization of a field element. These are useful
/// for point compression.
pub trait Flags: Default + Clone + Copy + Sized {
```

serialize/src/flags.rs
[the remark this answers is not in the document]
ValarDragon: oh I see. So serialize with flags takes the length of the serialized data in bits, and either changes the last `BIT_SIZE` bits, or if there aren't `BIT_SIZE` free bits left, then it adds a new byte?

serialize/src/lib.rs
Pratyush: Only `usize` differs between architectures, but for `usize` we explicitly serialize it as `u64`.

ec/src/models/twisted_edwards_extended.rs
```diff
@@ -742,15 +740,11 @@ impl<P: Parameters> CanonicalSerialize for GroupAffine<P> {
 
     #[inline]
     fn uncompressed_size(&self) -> usize {
-        Self::UNCOMPRESSED_SIZE
+        // x  + y
+        P::BaseField::zero().serialized_size() + P::BaseField::zero().serialized_size()
```
ValarDragon, line 744: Seems like fields need a trait guarantee that they have constant serialization size, in order for this to be safe?

Pratyush: Not sure I understand what you mean?

serialize/src/lib.rs
ValarDragon: That makes sense

serialize/src/flags.rs
ValarDragon: LGTM

ValarDragon:
> It was removed because the size depends dynamically on the flag used (and we can't do generics in constants yet). For example, if I'm serializing with EdwardsFlags, a 255-bit field will still fit into 32 bytes. However, if I'm serializing with SWFlags, a 255-bit field will fit only into 33 bytes.

I'm a bit confused about the structure of this. Currently it sounds like the field has a method `serialize_with_flags`, but why is that done versus the following:

1) Have a method/trait to serialize a field element, and state the exact number of bits used in the serialization. (Basically add `serialized_size_bits` to CanonicalSerialize)
2) Have a method that generically takes a Flags object, and a `b` bit serialized blob, and append the flag bits to it

I'm not seeing why we would want to specialize the appending of flags to fields. Is there an efficiency gain in the current form?

serialize/src/flags.rs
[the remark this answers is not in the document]
ValarDragon: Perhaps
```suggestion
/// Represents metadata to be appended to an objects serialization. For
/// example, when serializing elliptic curve points, one can
/// use a `Flag` to represent whether the serialization is the point
/// at infinity, or whether the `y` coordinate is positive or not.
/// These bits will be appended to the end of the points serialization,
/// or included in a new byte, depending on space available.
/// This struct is meant to be provided to `CanonicalSerializeWithFlags` and
/// `CanonicalDeserializeWithFlags`
pub trait Flags: Default + Clone + Copy + Sized {
```

Pratyush:
Not all objects care about `Flags`; in fact, only `Field`s care about them. So it doesn't make sense to add these methods to the `CanonicalSerialize` trait, which is why we have the `CanonicalSerializeWithFlags` trait

ec/src/models/twisted_edwards_extended.rs
Pratyush: Do you mean that we don't do varint-style encoding?

ValarDragon: Suppose a field specialized how they serialize `0`. e.g. they serialized `0` with 1 byte `0u8`, and every other element is serialized as `1 + 2 * data_value_in_bits`.

(Or more generally, if a field serialized two elements `a` and `b` in ways such that `serialized_size(a) != serialized_size(b)`, which can arise with varint encodings for instance)

serialize/src/flags.rs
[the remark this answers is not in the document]
Pratyush: Yes!

ec/src/models/twisted_edwards_extended.rs
Pratyush: Yeah, we need this guarantee (and it's satisfied by all existing field implementations)

Pratyush: But I can change it to this to make it more accurate:
```suggestion
        self.x.serialized_size() + self.y.serialized_size()
```

ec/src/models/short_weierstrass_jacobian.rs
```diff
@@ -746,15 +745,13 @@ impl<P: Parameters> CanonicalSerialize for GroupAffine<P> {
 
     #[inline]
     fn uncompressed_size(&self) -> usize {
-        Self::UNCOMPRESSED_SIZE
+        // x
+        P::BaseField::zero().serialized_size()
+            // y
+            + P::BaseField::zero().serialized_size_with_flags::<SWFlags>()
```
Pratyush, line 751:
```suggestion
        self.x.serialized_size() + self.y.serialized_size_with_flags::<SWFlags>()
```

ValarDragon:
The only method that would be added to `CanonicalSerialize` would be `serialized_size_bits` (which could even have an auto-impl to just be `8 * serialized_size`). 

The concept of flags is essentially taking two serializable objects, and concatenating their serializations at the bit-level instead of the byte level. I don't currently understand why we'd prefer the flag trait versus making a serialization method / deserialization method that concatenates object serializations at the bit-boundary, instead of the byte boundary. Maybe we should put this into an issue, and revisit later? A more general method would also enable saving space in serializing many 257 bit field elements

ec/src/models/short_weierstrass_jacobian.rs
ValarDragon: lgtm

Pratyush:
Hmm that is interesting, but out of scope for this PR =P

I've filed an issue about it, so that we can revisit it in the future (it might be that the space saving would be blown out by the implementation complexity)

Pratyush:
@ValarDragon is there anything else you'd like to see changed here?

ValarDragon (approved):
LGTM! Thanks for adding more documentation around all of these changes as well

Issue #156, opened 2020-12-28
Make Field::from_bytes method that reduces the input


ValarDragon:
Currently fields have a `from_random_bytes` method, which will parse a field element from an input that is the correct number of words long.

However for hashing to a field element, you need the field element to be uniformly sampled from the field. This requires either rejection sampling, or to sample a much larger number, and perform modular reduction. The approach taken in #147 relies on performing modular reduction of a much larger number, into the field.

I think we should rename the method `F::from_random_bytes` to `F::from_bytes` and make it allow for much larger inputs, and do modular reduction on it accordingly.

PR #164, opened 2020-12-31
Add PrimeField::from_bytes_mod_order

ValarDragon:
## Description

Adds a method `from_bytes_mod_order`, needed for hashing to the curve (#147). 

This also includes a python file I used for creating test vectors for the method. I don't know how to write useful tests for this function for a generic field (since we'd need to do arithmetic on integers of a length longer than the field). Hence this function is only tested on one field, but includes a script to add test vectors for any field.

ref: https://github.com/zkcrypto/ff/issues/45
closes: #156 
---

write a little note why.

- [x] Wrote unit tests


weikengchen (approved):
The PR looks good. Especially the test is comprehensive.

ff/src/fields/macros.rs
```diff
@@ -443,6 +443,31 @@ macro_rules! impl_Fp {
                 }
             }
```
Pratyush, line 445: Can this just be a default method on the PrimeField trait? It seems to use only generic methods, and doesn’t seem to require any specific information about the PrimeField impl

Pratyush:
> I don't know how to write useful tests for this function for a generic field

We have access to the Size of the modulus generically, so we can use that to generate enough bytes, right? And for the independent source of truth we can use something like num-bigint as a dev-dependency

ValarDragon:
Didn't realize there were pre-existing variable sized big-int libraries. Do you think we should go with a general approach for testing before merging this, or have a todo for using num-bigint as a dev-dependency?

weikengchen:
I think it would be fine to use `num-bigint`. Indeed, nonnative already uses it (https://github.com/arkworks-rs/nonnative/blob/master/Cargo.toml)

ValarDragon:
Switched to using num-bigint for checking this. Is it fine to keep it with the test values being generated via python?

CHANGELOG.md
```diff
@@ -36,6 +36,7 @@ The main features of this release are:
     if the top bits of the `u8` value do *not* correspond to one of the possible outputs of `Flags::u8_bitmask`, then these methods output `None`, whereas before they output
     a default value.
   Downstream users other than `ark-curves` should not see breakage unless they rely on these methods/traits explicitly.
+- #164 (ark-ff) Add a method `from_bytes_mod_order` to the `PrimeField` trait. Only users who implement prime fields without using the macros in ark-ff will see breakage.
```
Pratyush, line 39:
```suggestion
- #164 (ark-ff) Add a method `from_bytes_mod_order` to the `PrimeField` trait. Only users who implement prime fields without using the macros in `ark-ff` will see breakage.
```

ff/src/fields/mod.rs
```diff
@@ -344,6 +345,32 @@ pub trait PrimeField:
     /// Returns the underlying representation of the prime field element.
     fn into_repr(&self) -> Self::BigInt;
 
+    /// Reads bytes in big-endian, and converts them to a field element.
```
Pratyush, line 348: Why big-endian? AFAIK most APIs are usually little-endian (for example: `BigInteger` itself

Pratyush: (maybe we should add `from_bytes_be_mod_order` and `from_bytes_le_mod_order`.

CHANGELOG.md
ValarDragon: Removed the breaking change bit, since this is no longer breaking

ff/src/fields/mod.rs
ValarDragon: The two hash to field API's I looked at for reference both used big endian, but I applied the suggested change.

ff/src/fields/mod.rs
```diff
@@ -344,6 +345,40 @@ pub trait PrimeField:
...
+
+        // Update the result, byte by byte.
+        // We go through existing field arithmetic, which handles the reduction.
+        // TODO: If we need higher speeds, parse more bytes at once, or implement
+        // modular multiplication by a u64
+        let window_size = Self::from(256u64);
+        for i in num_bytes_to_directly_convert..bytes.len() {
+            res *= window_size;
+            res += Self::from(bytes[i]);
+        }
```
Pratyush, line 370:
```suggestion
        for byte in bytes[num_bytes_to_directly_convert..] {
            res *= window_size;
            res += Self::from(byte);
        }
```

ff/src/fields/mod.rs
```diff
@@ -614,4 +650,100 @@ mod no_std_tests {
             );
         }
     }
+
+    #[test]
+    fn test_from_be_bytes_mod_order() {
```
Pratyush, line 655: Should we also add tests for `le`?

Pratyush:
btw for the test, does it also make sense to test with random inputs?

ValarDragon:
Now that we have a reference, it makes sense

Pratyush:
This LGTM modulo the nit and the comment about tests.

PR #166, opened 2021-01-06
Add to_bytes to bigint

ValarDragon:
## Description

Adds a simple `to_bytes` method to `bigint`. Currently there is `to_bits` and `write`, and `write` is a bit awkward to use if you're not concerned with efficiency of the serialization.

---

write a little note why.

- [ ] Wrote unit tests

### DIFF
--- a/ec/src/msm/fixed_base.rs
+++ b/ec/src/msm/fixed_base.rs
@@ -65,7 +65,7 @@ impl FixedBaseMSM {
         scalar: &T::ScalarField,
     ) -> T {
         let modulus_size = <T::ScalarField as PrimeField>::Params::MODULUS_BITS as usize;
-        let mut scalar_val = scalar.into_repr().to_bits();
+        let mut scalar_val = scalar.into_repr().to_bits_be();
         scalar_val.reverse();
 
         let mut res = multiples_of_g[0][0].into_projective();
--- a/ec/src/msm/fixed_base.rs
+++ b/ec/src/msm/fixed_base.rs
@@ -65,8 +65,7 @@ impl FixedBaseMSM {
         scalar: &T::ScalarField,
     ) -> T {
         let modulus_size = <T::ScalarField as PrimeField>::Params::MODULUS_BITS as usize;
-        let mut scalar_val = scalar.into_repr().to_bits_be();
-        scalar_val.reverse();
+        let scalar_val = scalar.into_repr().to_bits_le();
 
         let mut res = multiples_of_g[0][0].into_projective();
         for outer in 0..outerc {
--- a/ff/src/biginteger/macros.rs
+++ b/ff/src/biginteger/macros.rs
@@ -168,11 +168,7 @@ macro_rules! bigint_impl {
 
             #[inline]
             fn to_bits(&self) -> Vec<bool> {
-                let mut res = Vec::with_capacity($num_limbs * 64);
-                for b in BitIteratorBE::new(self.0) {
-                    res.push(b);
-                }
-                res
+                BigIteratorBE::new(self.0).collect::<Vec<_>>()
             }
 
             #[inline]
--- a/ff/src/biginteger/macros.rs
+++ b/ff/src/biginteger/macros.rs
@@ -168,9 +168,28 @@ macro_rules! bigint_impl {
 
             #[inline]
             fn to_bits(&self) -> Vec<bool> {
-                let mut res = Vec::with_capacity(256);
+                let mut res = Vec::with_capacity($num_limbs * 64);
                 for b in BitIteratorBE::new(self.0) {
                     res.push(b);
+                }
+                res
+            }
+
+            #[inline]
+            fn to_bytes(&self) -> Vec<u8> {
+                let mut res = Vec::with_capacity($num_limbs * 8);
+                let mut cur_byte = 0u8;
+                let mut bit_num_mod_8 = 0;
+                for b in BitIteratorBE::new(self.0) {
+                    cur_byte = (cur_byte * 2) + (b as u8);
+                    bit_num_mod_8 = (bit_num_mod_8 + 1) % 8;
+                    if bit_num_mod_8 == 0 { 
+                        res.push(cur_byte);
+                        cur_byte = 0u8;
+                    }
+                }
+                if bit_num_mod_8 != 0 {
+                    res.push(cur_byte);
                 }
                 res
             }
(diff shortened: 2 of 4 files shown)


ff/src/biginteger/mod.rs
```diff
@@ -102,6 +102,10 @@ pub trait BigInteger:
     /// leading zeros.
     fn to_bits(&self) -> Vec<bool>;
 
+    /// Returns the byte representation in a big endian byte array, without
+    /// leading zeros.
```
Pratyush, line 106: Is there a reason to omit leading zeros? It makes the size dependent on the value, which can be confusing. Also, `to_bits` is not value-dependent, AFAIK

ValarDragon: Oh you're right, `to_bits` isn't value dependent, so this is also not value dependent. (Its using the same bit iterator as `to_bits`) I'll fix the comments accordingly.

ff/src/biginteger/macros.rs
```diff
@@ -168,13 +168,32 @@ macro_rules! bigint_impl {
 
             #[inline]
             fn to_bits(&self) -> Vec<bool> {
-                let mut res = Vec::with_capacity(256);
+                let mut res = Vec::with_capacity($num_limbs * 64);
```
Pratyush, line 171: Separately, I think we can replace this with `BigIteratorBE::new(self.0).collect::<Vec<_>>()`

ValarDragon: Oh true, I'll change it to that

ff/src/biginteger/macros.rs
```diff
@@ -168,13 +168,32 @@ macro_rules! bigint_impl {
...
+                    bit_num_mod_8 = (bit_num_mod_8 + 1) % 8;
+                    if bit_num_mod_8 == 0 {
+                        res.push(cur_byte);
+                        cur_byte = 0u8;
+                    }
+                }
+                if bit_num_mod_8 != 0 {
+                    res.push(cur_byte);
+                }
+                res
```
Pratyush: How about we rewrite this using `to_bits`?

```rust
                self
                	.to_bits()
                    .into_iter()
                    .chunks(8)
                    .map(|bits| /* logic to get byte from 8-bit chunks */)
                    .collect()
```

ff/src/biginteger/macros.rs
```diff
@@ -168,9 +168,24 @@ macro_rules! bigint_impl {
 
             #[inline]
             fn to_bits(&self) -> Vec<bool> {
-                let mut res = Vec::with_capacity(256);
+                BigIteratorBE::new(self.0).collect::<Vec<_>>()
```
Pratyush, line 171:
```suggestion
                BitIteratorBE::new(self.0).collect::<Vec<_>>()
```

ff/src/biginteger/macros.rs
```diff
@@ -166,11 +166,45 @@ macro_rules! bigint_impl {
...
+            }
+
             #[inline]
-            fn to_bits(&self) -> Vec<bool> {
-                let mut res = Vec::with_capacity(256);
-                for b in BitIteratorBE::new(self.0) {
-                    res.push(b);
+            fn to_bytes_le(&self) -> Vec<u8> {
+                let array_map = self.0.iter().map(|limb| limb.to_le_bytes());
+                let mut res = Vec::<u8>::new();
```
Pratyush, line 205:
```suggestion
                let mut res = Vec::<u8>::with_capacity($num_limbs * 8);
```

ff/src/biginteger/macros.rs
```diff
@@ -166,11 +166,45 @@ macro_rules! bigint_impl {
...
+            fn to_bits_le(&self) -> Vec<bool> {
+                BitIteratorLE::new(self.0).collect::<Vec<_>>()
+            }
+
+            #[inline]
+            fn to_bytes_be(&self) -> Vec<u8> {
+                let mut le_bytes = self.to_bytes_le();
+                le_bytes.reverse();
+                le_bytes
+            }
```
Pratyush, line 200: I think these can all be made default impls on the trait itself? For the bits, you can call `BitIteratorXX::new(self)...`, and `to_bytes_be` is implemented directly in terms of `to_bytes_le`.

ValarDragon: Sure, but do we want to support BigInteger as a trait for others to be implementing? I assumed BigInt was something internal to the library

ValarDragon: True, ok I'll do that

ValarDragon: I didn't realize the trait had `AsRef<[u64]>`, seems kind of word to be in the trait imo. Why would we not allow float back ends?

ValarDragon: Actually I think its weird to only implement one of `to_bytes_le` and `to_bytes_be` as the default, I'd prefer it in the macro. (The reason to put it in the default impl would be for users implementing these in an alternate fashion, but that then assumes that `to_bytes_le` is the native endianness to do it)

Pratyush: Actually I think we can move `to_bytes_le` also to the trait, as it can be implemented from `AsRef<[u64]>`.

(This is an example of why the `AsRef<[u64]>` trait bound is useful)

ValarDragon: I'll move it for now, but I think its worth revisiting if BigInt needs AsRef<[u64]>, I feel like we shouldn't make PrimeField / elliptic curves depend on that

ValarDragon: We lose capacity hints with the `to_bytes` being a default implementation, not really sure its worth making it default then?

ValarDragon: Ah no, I was wrong. The capacity hints are still available

ec/src/msm/fixed_base.rs
```diff
@@ -65,7 +65,7 @@ impl FixedBaseMSM {
         scalar: &T::ScalarField,
     ) -> T {
         let modulus_size = <T::ScalarField as PrimeField>::Params::MODULUS_BITS as usize;
-        let mut scalar_val = scalar.into_repr().to_bits();
+        let mut scalar_val = scalar.into_repr().to_bits_be();
         scalar_val.reverse();
```
Pratyush, line 69: This can just be `scalar.into_repr().to_bits_le()`, I believe

Pratyush (approved):

PR #169, opened 2021-01-07
Improve radix-2 FFTs

Pratyush:
## Description

Improves radix-2 FFTs by moving to the algorithm described in https://github.com/kwantam/fffft. The implementation matches the one there almost verbatim, except for tweaks specific to `ark-poly` APIs. The speedups provided by this new algorithm are *really* nice: between 25-50% on even moderate sizes.

---

write a little note why.

- [x] Wrote unit tests


weikengchen:
Should we add rsw to the `authors` field of `Cargo.toml` of `ff`? Another potential place is the LICENSE file, but we never wrote down the copyright owner field in the license file.

Pratyush:
@kwantam, how would you like to be credited? (fantastic work btw)

ValarDragon:
I think at minimum, there should be a header at the top of the file explaining it came from `kwantam/fffft`, and was written by kwantam, but just adapated to the arkworks traits

kwantam:
Hi folks! Glad the code is useful, and thanks for the kind words!

Feel free to credit however makes sense for your project. The suggested comment at the top of a file with a pointer to the fffft repo/crate seems like a nice idea. If you list names in an authors file, my real name and/or GitHub user-id would be cool.

Cool work, by the way!

weikengchen:
Just out of severe curiosity, any story on the GitHub user-id `kwantam`? @kwantam

kwantam:
> Just out of severe curiosity, any story on the GitHub user-id `kwantam`? @kwantam

Not much of one---I figured that the correct spelling would be unavailable as a username, whereas a misspelling might not (and indeed it seems like that's usually the case). :)

poly/src/domain/radix2/fft.rs
```diff
@@ -0,0 +1,200 @@
+// The code below is a port of the excellent library of https://github.com/kwantam/fffft
```
ValarDragon, line 1:
```suggestion
// The code below is a port of the excellent library of https://github.com/kwantam/fffft by Riad Wahby
```

poly/src/domain/radix2/mod.rs
```diff
@@ -54,7 +50,11 @@ impl<F: FftField> EvaluationDomain<F> for Radix2EvaluationDomain<F> {
     /// having `num_coeffs` coefficients.
     fn new(num_coeffs: usize) -> Option<Self> {
         // Compute the size of our evaluation domain
-        let size = num_coeffs.next_power_of_two() as u64;
+        let size = if num_coeffs.is_power_of_two() {
```
Pratyush, line 53: This avoids bumping up to the next power of two in an edge case.

poly-benches/benches/fft.rs
```diff
@@ -20,12 +20,6 @@ const BENCHMARK_LOG_INTERVAL_DEGREE: usize = 1;
 const ENABLE_RADIX2_BENCHES: bool = true;
 const ENABLE_MIXED_RADIX_BENCHES: bool = true;
 
-const ENABLE_SUBGROUP_FFT_BENCH: bool = true;
```
ValarDragon, line 23: i thought these flags were useful, so that I could just bench the function I was working on

Pratyush: You can still do that by passing an argument to cargo bench. For example, `cargo bench ifft` runs only the `ifft` benchmarks. (You can pass an arbitrary regex in place of `ifft`.)

weikengchen (approved):
The code looks good. But it is good if @ValarDragon takes a final pass.

ValarDragon (approved):
LGTM! Lets compare the methods for mitigating parallelization overhead in a second PR?

PR #171, opened 2021-01-08
FFT parallelization threshold

Pratyush:
## Description

Adds a threshold when parallelizing FFTs. Currently the threshold is derived empirically; it gives speedups of between 15-30% compared to the non-threshold version.

---

write a little note why.

- [x] Wrote unit tests


weikengchen (approved):
I think it looks good. But, a comment noting that this number is chosen empirically in the code would be helpful I think.

Pratyush:
Ok, done!

ValarDragon:
Thanks for integrating this! We should still benchmark this against hardcoding the base case tho

Issue #172, opened 2021-01-09
Benchmark radix2 FFT to determine performance overhead at small size


Pratyush:
We should benchmark the current behaviour of radix2 FFTs at small sizes (rayon's with_min_len) against just hardcoding the while loop for small sizes to use `iter` instead of `cfg_iter`.

_Originally posted by @ValarDragon in https://github.com/arkworks-rs/algebra/issues/171#issuecomment-757097625_

PR #173, opened 2021-01-10
Hardcode FFT inner loop

ValarDragon:
## Description

This hardcodes the inner loop to not use parallelism, which does in fact yield better speeds. Below is a summary of the speed difference this PR made on a fresh, 16 core, 32 GB RAM, CPU-optimized Digital Ocean server.

The summary of the speed differences is that at all sizes where the data vector fit into RAM (tested from log_degree = 15 to log_degree=20), this change improved speed by ~10%. 

When the polynomial size was bigger (tested from log_degree = 21, 22), this change improved speed by 5-6%.
At polynomial sizes of log_degree = 23,24,25, speed improved by 8%.

As the FFT is in the critical path, I believe we ought to go with this approach here.

closes: #172 

---

write a little note why.

- [x] Wrote unit tests

### DIFF
--- a/poly/src/domain/radix2/fft.rs
+++ b/poly/src/domain/radix2/fft.rs
@@ -78,7 +78,7 @@ impl<F: FftField> Radix2EvaluationDomain<F> {
             let chunk_size = 2 * gap;
             let nchunks = xi.len() / chunk_size;
 
-            let inner_fn = |(idx, (lo, hi)): (usize, (&mut T, &mut T))| {
+            let butterfly_fn = |(idx, (lo, hi)): (usize, (&mut T, &mut T))| {
                 let neg = *lo - *hi;
                 *lo += *hi;
 
@@ -92,9 +92,9 @@ impl<F: FftField> Radix2EvaluationDomain<F> {
             ark_std::cfg_chunks_mut!(xi, chunk_size).for_each(|cxi| {
                 let (lo, hi) = cxi.split_at_mut(gap);
                 if chunk_size > MIN_CHUNK_SIZE_FOR_PARALLELIZATION {
-                    cfg_iter_mut!(lo).zip(hi).enumerate().for_each(inner_fn);
+                    cfg_iter_mut!(lo).zip(hi).enumerate().for_each(butterfly_fn);
                 } else {
-                    lo.iter_mut().zip(hi).enumerate().for_each(inner_fn);
+                    lo.iter_mut().zip(hi).enumerate().for_each(butterfly_fn);
                 }
             });
             gap /= 2;
@@ -109,7 +109,7 @@ impl<F: FftField> Radix2EvaluationDomain<F> {
             let chunk_size = 2 * gap;
             let nchunks = xi.len() / chunk_size;
 
-            let inner_fn = |(idx, (lo, hi)): (usize, (&mut T, &mut T))| {
+            let butterfly_fn = |(idx, (lo, hi)): (usize, (&mut T, &mut T))| {
                 *hi *= roots[nchunks * idx];
                 let neg = *lo - *hi;
                 *lo += *hi;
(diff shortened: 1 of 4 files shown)


ValarDragon:
I'm having issues avoiding code re-use for the fft_core_loop. I tried to define a common variable as
```
let fft_core_loop: fn((usize, (&mut F, &mut F))) = 
                |(idx, (lo, hi)): (usize, (&mut F, &mut F))| {
                let neg = *lo - *hi;
                *lo += *hi;

                *hi = neg;
                *hi *= roots[nchunks * idx];
            };
```
and re-use it in both loops, but I got the error:

Screenshot 2021-01-10 at 2 56 40 AM

I tried playing around with lifetimes with no luck, any suggestions on how to avoid code duplication? (I'll benchmark any change)

Pratyush:
thanks for the PR

> When the polynomial size was bigger than RAM (tested from log_degree = 21, 22), this change improved speed by 5-6%.

Quick question: how is that bigger than memory? Degree 2^22 polynomials for 256-bit fields fit in less than 250 MB of memory

ValarDragon:
You're right. I divided by 1MB instead of 1GB in my math 😅 . I wonder why we get this non-linearity in speedup

Pratyush:
Caching effects? 2^20 is about 40 MB; maybe it doesn’t fit in cache any
more after that?

ValarDragon:
Oh yeah, that makes sense since small gap sizes have large stride

Pratyush:
> I'm having issues avoiding code re-use for the fft_core_loop

The latest commit should help with deduplication @ValarDragon, but I did move a conditional inside of a loop; you might want to check that and make sure it doesn't affect performance.

ValarDragon:
Thanks for the help in combining the code paths! Re-benchmarked after renames / code re-use improvements, and the performance improvements of the original post still stand.

I view this as good to merge.

poly/src/domain/radix2/fft.rs
```diff
@@ -75,19 +75,27 @@ impl<F: FftField> Radix2EvaluationDomain<F> {
...
-                    .zip(hi)
-                    .enumerate()
-                    .for_each(|(idx, (lo, hi))| {
-                        let neg = *lo - *hi;
-                        *lo += *hi;
-
-                        *hi = neg;
-                        *hi *= roots[nchunks * idx];
-                    });
+                if gap > MIN_CHUNK_SIZE_FOR_PARALLELIZATION / 2 {
```
Pratyush, line 94:
```suggestion
                if chunk_size > MIN_CHUNK_SIZE_FOR_PARALLELIZATION {
```

poly/src/domain/radix2/fft.rs
```diff
@@ -98,25 +106,39 @@ impl<F: FftField> Radix2EvaluationDomain<F> {
...
-                ark_std::cfg_iter_mut!(lo, 1000) // threshold of 1000 was determined empirically
-                    .zip(hi)
-                    .enumerate()
-                    .for_each(|(idx, (lo, hi))| {
-                        *hi *= roots[nchunks * idx];
-                        let neg = *lo - *hi;
-                        *lo += *hi;
-                        *hi = neg;
-                    });
+                if gap > MIN_CHUNK_SIZE_FOR_PARALLELIZATION / 2 {
```
Pratyush, line 124:
```suggestion
                if chunk_size > MIN_CHUNK_SIZE_FOR_PARALLELIZATION {
```

poly/src/domain/radix2/fft.rs
Pratyush: Oh I just saw that the latest change comment reverted this change; is there a reason why?

ValarDragon: yeah, really oddly the 2^25 degree FFT was saying it was 1% slower with the `chunk_size > ` form.

 I didn't compile in god bolt to check if `chunk_size` is really taking an extra register, or if its just system noise. In the case that it is taking an extra register, we could just un-nest the if statement.

Pratyush: In that case but we might want to just remove chunk_size, right?

ValarDragon: I prefer the constant being called `chunk_size` though. I'll add a comment about the if statement stating the condition in terms of chunk size?

Pratyush: Alright, sounds good then, this PR is good to go =)

Pratyush (approved):

PR #174, opened 2021-01-14
Fix inconsistency between parallel and serial coset ffts

ValarDragon:
## Description

The current coset FFT is inconsistent between parallel and serial versions, which led to hard to debug bugs. (We suspect due to bit reversal issues). This reverts to a simpler parallel coset compute powers method, add tests for parallel/serial consistency.

We should re-investigate how to apply the faster compute powers method without causing errors.

---

write a little note why.

- [x] Wrote unit tests
- [x] Added a relevant changelog entry to the `Pending` section in `CHANGELOG.md` - N/A since its a bug introduced this update


Pratyush:
This LGTM, but let's file an issue to follow up about the bit-reversal issue

Pratyush (approved):

Pratyush:
Ah one sec; doesn't this PR also affect the radix2 FFT?

ValarDragon:
Changing compute powers does affect the subgroup radix 2 FFT

poly/src/domain/radix2/fft.rs
```diff
@@ -61,12 +62,68 @@ impl<F: FftField> Radix2EvaluationDomain<F> {
...
     #[cfg(not(feature = "parallel"))]
     fn roots_of_unity(&self, root: F) -> Vec<F> {
-        crate::domain::utils::compute_powers_serial(self.size as usize, root)
+        compute_powers_serial(self.size as usize, root)
     }
 
     #[cfg(feature = "parallel")]
     fn roots_of_unity(&self, root: F) -> Vec<F> {
-        crate::domain::utils::compute_powers(self.size as usize, root)
+        // TODO: Understand why this functions output isn't domain.elements(),
```
Pratyush, line 70: The corresponding test fails? If so we should add it anyway, but mark it with an `#[ignore]` attribute

ValarDragon: added!

PR #177, opened 2021-01-14
Cache align FFT core loop code

ValarDragon:
## Description

This PR changes the core loop for the FFT to access the roots of unity in a cache aligned way, similar to what is implemented in libiop. Currently its set to minimize computation, which is actually pretty bad for performance given that its memory bottlenecked.

In libiop we saw that doing this was a huge speedup. So far, I've implemented this in this PR with an arbitrary set of parameters, and this is already a `10%` speedup in the serial case. I anticipate it will perform better. (Code is not yet  benchmarked for the parallel case)

---

write a little note why.

- [x] Wrote unit tests - n/a, covered by existing tests


poly/src/domain/radix2/fft.rs
```diff
@@ -138,20 +138,29 @@ impl<F: FftField> Radix2EvaluationDomain<F> {
...
 
         let mut gap = xi.len() / 2;
         while gap > 0 {
             // each butterfly cluster uses 2*gap positions
             let chunk_size = 2 * gap;
-            let nchunks = xi.len() / chunk_size;
+            let root_len = cache_aligned_roots.len();
+            if root_stride > MAX_ROOT_STRIDE && root_len > ROOT_STOP_RESIZING {
+                for i in 0..(root_len / root_stride) {
+                    cache_aligned_roots[i] = cache_aligned_roots[i * root_stride];
```
ValarDragon, line 153: Is there a way to do a loop like this with rayon, without additional memory allocations?

ValarDragon: Maybe not, since it can't really be split into segments nicely

Pratyush: To be in place, mustn't this be sequential? For example, because we must read `cache_aligned_roots[1 * 4]` before we overwrite `cache_aligned_roots[4]`

(also, the first iteration just copies 0 into itself, so maybe we want to start at 1 instead of 0)

poly/src/domain/radix2/fft.rs
```diff
@@ -213,6 +232,22 @@ const MIN_CHUNK_SIZE_FOR_PARALLELIZATION: usize = 2048;
...
 const LOG_ROOTS_OF_UNITY_PARALLEL_SIZE: u32 = 7;
 
+// Maximum stride allowed for the root vector, before re-cache aligning it.
+// This won't cache align if we've reached the ROOT_STOP_RESIZING threshold.
+// TODO: Better understand how this should be optimized
+// (according to cache lines, and hardware prefetchers)
+const MAX_ROOT_STRIDE: usize = 2;
+
+// Once the size of the cache aligned roots is below this number, stop re-aligning it.
+// TODO: Figure out how we can make this depend on field size & system cache size.
```
Pratyush, line 242: We can make it depend on field size by figuring out the field size as `F::characteristic() * F::degree()`

ValarDragon: But its a constant, so it would have to go in the FftField params right?

ValarDragon: Or can you have constants depend on a template parameter?

ValarDragon:
Prior to the "seperate cache align levels" commit, I was seeing a 5% slowdown to parallel FFTs, until a size `2 << 21` fft, in which case it was a `7.5%` speedup. 

Maybe this suggests that we need a parameter for when to start cache aligning as well? (Also I think we should just brute force over all the parameters to set the final version. Do we have scripts for brute forcing over constants, w/ benchmarks?)

Pratyush:
We don't have scripts to brute force over the parameters. One way could be to just make an inner function that accepts the constants as parameters, and then run standard criterion benchmarks over the range of parameters?

(also it would be nice if this code was cache-oblivious, so we didn't have to worry about these things lol)

ValarDragon:
Hrmm, another option could be a python script setup. (python regex the file to swap out the constants, run cargo bench, get the output and output a CSV of all the results at the end)

Agreed that making this cache oblivious would be ideal. In the serial case, it looks like that is fine, but its not looking like that so far for the parallel case =/

poly/src/domain/radix2/fft.rs
Pratyush: Ah you're right, they're not `const fn`... I'm not sure we need it to be actually const; the compiler should be able to perform constant folding

ValarDragon:
Should I just convert this to something that is faster in the serial case over most sizes, and leave as a TODO investigating the parallel case? (I'll be very surprised if the parallel case doesn't show a similar speed improvement with the right parameterization)

Pratyush:
> Prior to the "seperate cache align levels" commit, I was seeing a 5% slowdown to parallel FFTs, until a size 2 << 21 fft, in which case it was a 7.5% speedup.

So is that ^ the state of things atm? If so, sure, let's just merge it for the serial case, and leave a TODO for the parallel case.

ValarDragon:
Yeah that is the state of things atm

Pratyush:
Let's update this only for the serial case and then let's merge?

Pratyush:
ping @ValarDragon?

ValarDragon:
Closing this in favor of #242. I will write up learnings from this code into an issue for later optimizations to the parallel case & IFFT

PR #182, opened 2021-01-18
Preliminary commit for min const generics

jon-chuang:
## Description

To do: 
- match to select literal loop-unrolled impl for square/mul (= very verbose macro-expanded source)...
- Get rid of weird artifacts due to `[T; N * 2]` where N is generic being impossible

Tests should pass on nightly 1.51.

Thoughts:
- ~limitations still exist for min const generics, especially for our literal-dependent loop unrolling use case, and also const eval with generics.~ Currently, this has been worked around
- Not sure how much real gains there are, yes fewer macros, but many breaking changes for imports and for many copy-pasta like `BigInt::<4>([0, 0, ..])`
- 
----
Update:
TODO:
- Check for performance regressions

Thoughts:
- Currently, `mul_without_reduce` is instantiated using loops that are not forced to be unrolled, so it is unclear what the compilers behaviour will be. It is only used in runtime code when `NO_CARRY` is false however. So I think its not a big issue. Still.

Currently, this is how mul_assign looks macro expanded:
```
 impl<'a, P: FpParams<N>, const N: usize> MulAssign<&'a Self> for Fp<P, N> {
                #[inline]
                fn mul_assign(&mut self, other: &Self) {
                    if P::NO_CARRY {
                        let input = &mut (self.0).0;
                        let other_ = (other.0).0;
                        match N {
                            1 => mul_assign_id1::<P, N>(input, other_),
                            4 => mul_assign_id4::<P, N>(input, other_),
                            5 => mul_assign_id5::<P, N>(input, other_),
                            6 => mul_assign_id6::<P, N>(input, other_),
                            12 => mul_assign_id12::<P, N>(input, other_),
                            13 => mul_assign_id13::<P, N>(input, other_),
                            _ => {
                                ::core::panicking::panic("internal error: entered unreachable code")
                            }
                        };
                        self.reduce();
                    } else {
                        *self = self.mul_without_reduce(other, P::MODULUS, P::INV);
                        self.reduce();
                    }
                }
            }
```

So, rather neat. But notice we define functions for each n limbs. Since we can control exactly which numbers we instantiate for via  the `limb_instantiation!(1, 4, 5, 6, 12, 13);` macro, I think unsavoury code bloat is substantially reduced.

---

write a little note why.

- [ ] Wrote unit tests


ff/src/fields/arithmetic.rs
```diff
@@ -58,10 +58,10 @@ macro_rules! impl_field_mul_assign {
 }
 
 macro_rules! impl_field_into_repr {
-    ($limbs:expr, $BigIntegerType:ty) => {
+    ($limbs:expr) => {
         #[inline]
-        #[ark_ff_asm::unroll_for_loops]
-        fn into_repr(&self) -> $BigIntegerType {
+        // #[ark_ff_asm::unroll_for_loops]
```
Pratyush, line 63: Do you feel like the compiler can still optimize this well? Or should we try to do some manual unrolling inside the macro?

ff/src/biginteger/mod.rs
```diff
@@ -17,16 +17,343 @@ use zeroize::Zeroize;
...
-bigint_impl!(BigInteger384, 6);
-bigint_impl!(BigInteger768, 12);
-bigint_impl!(BigInteger832, 13);
+
+#[derive(Copy, Clone, PartialEq, Eq, Debug, Hash, Zeroize)]
+pub struct BigInt<const N: usize>(pub [u64; N]);
+
+impl<const N: usize> Default for BigInt<N> {
+    fn default() -> Self {
+        BigInt::<N>([0u64; N])
```
Pratyush, line 26:
```suggestion
        Self([0u64; N])
```

ff/src/biginteger/mod.rs
```diff
@@ -17,16 +17,343 @@ use zeroize::Zeroize;
...
+    }
+}
+
+impl<const N: usize> BigInt<N> {
+    pub const fn new(value: [u64; N]) -> Self {
+        BigInt::<N>(value)
```
Pratyush, line 32:
```suggestion
        Self(value)
```

ff/src/fields/arithmetic.rs
jon-chuang: Well, for now, since we can no longer take advantage of literals in macro instantiations, I've decided one should have match arms and do the loop unroll for each of the possible num limbs. 

This is extremely ugly, and I'm not sure if there's a workaround; the only plausible workaround for unrolling loops parametrised by const is proposal for compiler hints #[optimize(..)] that is not moving atm.

jon-chuang: So, basically, this is not the final form

jon-chuang: One can make the actual impl very compact, but under the hood, with macro expand etc it will be very ugly, just like our asm code selecting the num_limbs match arms

Pratyush:
Amazing, thanks for this!

>     * match to select literal loop-unrolled impl for square/mul (= very verbose macro-expanded source)...

Maybe we can use techniques from the C++ world for this? Eg: https://www.informit.com/articles/article.aspx?p=30667&seqNum=7

It could get complicated though.

>     * Get rid of weird artifacts due to `[T; N * 2]` where N is generic being impossible

Hmm that's unfortunate; it's due to this issue: https://github.com/rust-lang/rust/issues/43408


>     * Not sure how much real gains there are, yes fewer macros, but many breaking changes for imports and for many copy-pasta like `BigInt::<4>([0, 0, ..])`

Can't the breaking changes be worked around by still exporting `{BigInteger,Fp}{256/320/384...}`? Also, there would be nice gains in compile time, as the compiler is simply doing much less work.

ff/src/fields/models/fp.rs
```diff
@@ -0,0 +1,967 @@
...
+use num_traits::{One, Zero};
+
+use crate::{
+    biginteger::{arithmetic as fa, BigInt, BigInteger as _BigInteger},
+    bytes::{FromBytes, ToBytes},
+    fields::{FftField, Field, FpParameters, LegendreSymbol, PrimeField, SquareRootField},
+};
+use ark_serialize::*;
+
+pub trait FpParams<const N: usize>: FpParameters<BigInt = BigInt<N>> {}
```
Pratyush, line 18: I think we can get rid of this trait now, no?

ff/src/fields/arithmetic.rs
jon-chuang: I think I will simply modify the proc macro, exposing a new one that creates the match arms up to some limit, which can be a meta attr that can be parsed as usize for the match limit.

jon-chuang:
> Maybe we can use techniques from the C++ world for this? Eg: https://www.informit.com/articles/article.aspx?p=30667&seqNum=7
> 
> It could get complicated though.

Well for this, Im not sure if it will work; notice the metaprogramming examples use literals as the param. I think something a lot of people would like is the ability to do metaprogramming based on const. Unfortunately const is something evaluated at MIR level, so there is no longer any syntax to do metaprogramming with. Still, it would be really cool to have functionality to do metaprogramming at MIR level.

> Can't the breaking changes be worked around by still exporting `{BigInteger,Fp}{256/320/384...}`? Also, there would be nice gains in compile time, as the compiler is simply doing much less work.

ff/src/fields/arithmetic.rs
jon-chuang: I think, for a specific problem, which is copying the data back over in each of the match arms, since there would be a type mismatch, it might make sense to do an unsafe transmute of a pointer; however, this would violate our desire to have safe code.

ff/src/fields/models/fp.rs
jon-chuang: Well, for now, I use it to instantiate NO_CARRY as a const param. Since otherwise, it was troublesome to ensure it would evaluate at compile time.

jon-chuang:
So I would say at this point that Im quite happy with state of PR, except for having checked for performance regressions and careful review. Also the number of unresolved questions about API and so on. But the core of the code is looking pretty good if you ask me.

jon-chuang:
Here is some preliminary data for the performance regression tests.

There appears to be some problems with the serialization, and more pertinently, the `into_repr` implementations. 

Apart from that, things seem alright.
Screenshot from 2021-01-22 03-00-32

Issue #183, opened 2021-01-20
panic = "abort" on release


jon-chuang:
This can reduce binary size and improve instr cache performance. So maybe we should test this.

daira:
It's also completely necessary for security and correctness in mixed-language projects &mdash; because panics across an FFI boundary are undefined behaviour. By their nature it's impossible to prevent panics, and the only other alternative is to add a bunch of verbose and error-prone boilerplate to *all* code that crosses an FFI boundary (including in libraries you don't want to have to change). See also https://github.com/zcash/librustzcash/issues/78 and https://github.com/rust-lang/rust/issues/52652 .

Note that `panic = "abort"` only takes effect for a top-level crate, not dependencies.

PR #184, opened 2021-01-21
Switch to panic='abort' for safety across FFI boundaries

ValarDragon:
Thanks to @daira for pointing out the security impact, and @jon-chuang for raising the issue / pointing out perf improvements.

This still needs to be done in downstream repositories, as this change only applies to the top level crate.

closes: #183
---

write a little note why.

- [ ] Wrote unit tests - n/a?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,6 +19,7 @@ members = [
 opt-level = 3
 lto = "thin"
 incremental = true
+panic = 'abort'
 
 [profile.bench]
 opt-level = 3
@@ -30,6 +31,7 @@ debug-assertions = false
 
 [profile.dev]
 opt-level = 0
+panic = 'abort'
 
 [profile.test]
 opt-level = 3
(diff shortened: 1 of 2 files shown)


Pratyush:
Doesn’t this only affect our libraries, and not any downstream binaries?

ValarDragon:
Yes. my understanding of the [profile docs](https://doc.rust-lang.org/cargo/reference/profiles.html) is that this will only take precedence for FFI that is obtained from compiling `algebra` directly. (E.g. golang code using FFI for this curve arithmetic, and implementing BLS with it)

I don't think this is a big concern for algebra, but is much more likely to be one for groth16, and it doesn't really hurt to do it in all libraries.

Cargo.toml
```diff
@@ -30,6 +31,7 @@ debug-assertions = false
 
 [profile.dev]
 opt-level = 0
+panic = 'abort'
```
Pratyush, line 34: Is there a benefit to turning it to `abort` here?

ValarDragon: Not to my knowledge, but it feels like a safer default imo

Pratyush (approved):

PR #186, opened 2021-01-22
min const generics

jon-chuang:
Replacement for https://github.com/arkworks-rs/algebra/pull/182
Mostly done, what's left is to review/decide on API issues/integrate with rest of codebase (like curves)

---

write a little note why.

- [ ] Wrote unit tests


ff/src/fields/arithmetic.rs
```diff
@@ -6,240 +6,191 @@
...
+                BigInt::<N>(r)
             }
-            tmp.0 = r;
-            tmp
         }
     };
 }
-
-macro_rules! impl_field_square_in_place {
+// macro_rules! impl_deserialize_flags {
```
Pratyush, line 62:
```suggestion

```

ff/src/fields/arithmetic.rs
```diff
@@ -6,240 +6,191 @@
...
-                if i < $limbs - 1 {
-                    for j in 0..$limbs {
-                        if j > i {
-                            r[i + j] =
-                                mac_with_carry!(r[i + j], (self.0).0[i], (self.0).0[j], &mut carry);
-                        }
-                    }
-                    r[$limbs + i] = carry;
-                    carry = 0;
+// macro_rules! impl_deserialize_flags {
```
Pratyush, line 93:
```suggestion

```

Pratyush:
The macro infrastructure is kinda difficult to follow; is there a way to simplify it? Also, I think merging the `additive_ops` and `multiplicative_ops` macro complicates the macro impl  for only a small benefit. Could we revert that change?

ff/src/fields/models/fp.rs
```diff
@@ -0,0 +1,833 @@
...
+        r
+    }
+}
+
+impl<P: FpParams<N>, const N: usize> FftField for Fp<P, N> {
+    type FftParams = P;
+
+    #[inline]
+    fn two_adic_root_of_unity() -> Self {
+        Fp::<P, N>(P::TWO_ADIC_ROOT_OF_UNITY, PhantomData)
```
Pratyush, line 467:
```suggestion
        Self::new(P::TWO_ADIC_ROOT_OF_UNITY)
```

ff/src/fields/models/fp.rs
```diff
@@ -0,0 +1,833 @@
...
+    }
+
+    #[inline]
+    fn large_subgroup_root_of_unity() -> Option<Self> {
+        Some(Fp::<P, N>(P::LARGE_SUBGROUP_ROOT_OF_UNITY?, PhantomData))
```
Pratyush, line 472:
```suggestion
        Some(Self::new(P::LARGE_SUBGROUP_ROOT_OF_UNITY?))
```

ff/src/fields/models/fp.rs
```diff
@@ -0,0 +1,833 @@
...
+    }
+
+    #[inline]
+    fn multiplicative_generator() -> Self {
+        Fp::<P, N>(P::GENERATOR, PhantomData)
```
Pratyush, line 477:
```suggestion
        Self::new(P::GENERATOR)
```

ff/src/fields/models/fp.rs
```diff
@@ -0,0 +1,833 @@
...
+
+impl<P: FpParams<N>, const N: usize> Neg for Fp<P, N> {
+    type Output = Self;
+    #[inline]
+    #[must_use]
+    fn neg(self) -> Self {
+        if !self.is_zero() {
+            let mut tmp = P::MODULUS.clone();
+            tmp.sub_noborrow(&self.0);
+            Fp::<P, N>(tmp, PhantomData)
```
Pratyush, line 703:
```suggestion
            Self::new(tmp)
```

jon-chuang:
> The macro infrastructure is kinda difficult to follow; is there a way to simplify it? Also, I think merging the `additive_ops` and `multiplicative_ops` macro complicates the macro impl for only a small benefit. Could we revert that change?

I've added some comments. 

I guess that given our usecase, I can simplify by making more defaults

jon-chuang:
To be quite honest, the limitations of const means that const generics at present buys little imo, under the hood, one is more or less instantiating concretely over limb types within ff anyway. So it's unclear what one should have expected or might hope is yet to be improved.

jon-chuang:
I would say that I did prefer the previous method of just implementing concretely via macros. Using const generics while requiring literals for unrolls means some unpleasant workarounds that make it harder to modify the code.

huitseeker:
I would consider adding a check for the minimal rustc version supporting `min_const_generics.

In fact, I have this in a commit you can grab here:
https://github.com/huitseeker/algebra/commit/c69b1ed6a0505911e0335d63fc46f2eab204a057

Here's a sample output on an older compiler:
https://gist.github.com/cd63c76a4e53c2f4bbf19f9ca29f4eba

Pratyush:
Superseded by #372 and #379

Issue #118, opened 2020-12-07
Implement smaller test field for mixed radix


ValarDragon:
Currently the field we use for mixed radix tests is `MNT6Fr`, whose radix 2 component is relatively large (15 or 16). This forces the mixed radix FFT tests to do a 2**17-sized mixed radix FFT, which is the slowest test in the repository by a large margin.

We should implement a field to test with, that has a much smaller radix-2 component, so we can test the mixed radix FFT on smaller polynomial sizes.

weikengchen:
There is a shortcut---implementing a field that is basically `MNT4Fr`, but we cheat that its radix-2 component is small.

Pratyush:
> There is a shortcut---implementing a field that is basically MNT4Fr, but we cheat that its radix-2 component is small.

Sure, we can put this in `test-curves`

weikengchen:
(Just a correction that I previously said `MNT4Fr`, it is `MNT6Fr`, in case someone is heading in the wrong direction...)

weikengchen:
Given the challenges Nick and I were facing in the PR, I might start to find a curve that has low 2-arity (which I assume would be easy?)

Pratyush:
Hmm just saw the comment on the PR; yeah I think it might be easier to just start with a different mixed-arity curve.

Pratyush:
(It actually just needs to be a prime field, not even a curve, and sampling such field should be easy)

PR #187, opened 2021-01-26
added version of MNT4-753 Fq with small 2-adicity

npwardberkeley:
## Description

closes: #118 

---

write a little note why.

- [ ] Wrote unit tests


Pratyush:
Thanks for this! since this is just for testing, we can just make the small-adicity version the default (and remove the other one). We already test the high 2-adicity case via the BLS12 curve.

weikengchen:
Looking at the current code, we may need to bump up the generator from 17 to 17 ^ {2^??}. I am looking into this.

weikengchen:
Update: @Pratyush  we need some help.

Basically, it is hard to compute `T_MINUS_ONE_DIV_TWO` for our fake curve.

So, originally, T = (MODULUS - 1) / 2^S where S is the two-arity. Therefore, T is always an odd number.

But now, when we cheat that S = 3 (instead of S = 15), T is even, and T_MINUS_ONE_DIV_TWO is problematic.

-------

Is it possible for us to not modify T but still get the power down? Should we keep the T unchanged?

weikengchen:
Close this PR, for two reasons:

1. It seems that we need to keep the current MNT-753 because in the `r1cs_std` we use it to test UInt64.
2. Likely what we are going to do---we add a new curve, probably BN384, since it is easier to find parameters---which would be used both for #118 and UInt64.
3. After the changes are done and are propagated to r1cs_std (and also the `succinct_r1cs` repo), we drop the MNT-753 curve from here.

Pratyush:
Is the reason for using a pairing-friendly curve (BN384) just that it enables testing the Marlin gadget? If the curve doesn't need to be pairing friendly, you can just choose an arbitrary prime `p` that satisfies the smoothness requirements, and then generate a prime-order curve whose order is `p` via these tools: https://github.com/J08nY/ecgen or https://github.com/kwantam/eccons

weikengchen:
Nope. Just appear that `ecfactory` has code that generates a curve for BN384.

(Currently debugging the generator, but should be done soon)

PR #188, opened 2021-01-27
Make Short Weierstrass random sampling result in an element with unknown discrete log

kobigurk:
## Description

Short Weierstrass random sampling is currently not safe as a method for sampling random elements with unknown discrete log. The PR duplicates the Twisted Edwards random sampling behavior which is safe.

I have not added new tests nor documentation.
---

write a little note why.

- [ ] ~~Wrote unit tests~~
- [ ] ~~Updated relevant documentation in the code~~


Pratyush (approved):
Thanks for the PR!

weikengchen:
My main concern is a probably stupid question: for SW curves, for a random x, is it guaranteed that we can find a point on the curve? I have this concern since the `get_point_from_x` for SW curves is `Optional`.

weikengchen:
(this same stupid question also goes to the TE one)

Pratyush:
That's why the loop is there, no? If the x coordinate is invalid, we just keep sampling until we find a valid one.

weikengchen:
oh got it. I did not notice that loop.

PR #190, opened 2021-01-30
Add simple curve cycle trait for curve cycles

drewstone:
## Description
The purpose of this PR is to define a trait for standard (potentially non-pairing based) and pairing based curve cycles.

No unit tests as this is just new trait def.


drewstone:
Ngl I just wanna contribute to this awesome project. I am still wearing my training wheels when looking around. Let me know if this isn't the right place or implementation for something like this. I have yet to see a PCD or cycle engine defined for non-pairing based curves so I figured there ought to be one.

daira:
Looks good. Can you impl this for Pallas and Vesta?

Pratyush:
This makes me think we need to have a `CurveCycle` trait, like so:

```rust
pub trait CurveCycle 
where
    <Self::E1 as AffineCurve>::Projective: MulAssign<<Self::E2 as AffineCurve>::BaseField>,
    <Self::E2 as AffineCurve>::Projective: MulAssign<<Self::E1 as AffineCurve>::BaseField>,
{
	// Maybe these constraints are too tight and could cause rust's constraint solver
	// error out due to cycles (heh); we remove the
	// bounds on the associated types if that's the case.
	type E1: AffineCurve<
		BaseField = <Self::E2 as AffineCurve>::ScalarField,
		ScalarField = <Self::E2 as AffineCurve>::BaseField,
	>;
	type E2: AffineCurve<
		BaseField = <Self::E1 as AffineCurve>::ScalarField,
		ScalarField = <Self::E1 as AffineCurve>::BaseField,
	>;
}

pub trait PairingFriendlyCycle: CurveCycle {
	// add whatever extra bounds are necessary.
	type Engine1: PairingEngine<
		G1Affine = Self::E1,
		G1Projective = <Self::E1 as AffineCurve>::Projective,
		Fq = <Self::E1 as AffineCurve>::BaseField,
		Fr = <Self::E1 as AffineCurve>::ScalarField,
	>;
	
	type Engine2: PairingEngine<
		G2Affine = Self::E2,
		G2Projective = <Self::E2 as AffineCurve>::Projective,
		Fq = <Self::E2 as AffineCurve>::BaseField,
		Fr = <Self::E2 as AffineCurve>::ScalarField,
	>;
}
```

drewstone:
@daira where's the right place to implement that?

drewstone:
I've updated the trait with your feedback @Pratyush. Getting rid of the strict def on `E2` managed to get it to compile without issues. Let me know your thoughts!

ec/src/lib.rs
```diff
@@ -313,6 +313,8 @@ pub fn prepare_g2<E: PairingEngine>(g: impl Into<E::G2Affine>) -> E::G2Prepared
 }
 
 /// A cycle of pairing-friendly elliptic curves.
+/// TODO: remove this once dependencies are
+///       updated with new curve cycle traits
```
Pratyush, line 317: Maybe we can add a `#[deprecated("Please use `PairingFriendlyCycle` instead")]` marker here?

ec/src/lib.rs
```diff
@@ -313,6 +313,7 @@ pub fn prepare_g2<E: PairingEngine>(g: impl Into<E::G2Affine>) -> E::G2Prepared
 }
 
 /// A cycle of pairing-friendly elliptic curves.
+#[deprecated("Please use `PairingFriendlyCycle` instead")]
```
Pratyush, line 316: oops, this should be

```suggestion
#[deprecated(note = "Please use `PairingFriendlyCycle` instead")]
```

Pratyush:
Thanks for the PR! could you add a relevant entry in CHANGELOG.md under `Improvements`? Oh and a final format.

drewstone:
All set @Pratyush

Pratyush (approved):

PR #192, opened 2021-01-30
Fix assembly backend

Pratyush:
## Description

Fix assembly backend to actually reduce the result of squaring/multiplication. Depending on the available CI runner, also runs tests with the assembly backend.

The bug arises because the assembly backend (before this PR) did something like the following:
```rust
fn square_in_place(&mut self) {
	let mut a = (self.0).0; // `a` is a *copy* of the data in `self`.
	// do assembly stuff with `a`
	// this is where the bug arose: all assembly operations happened on `a`, 
	// but we didn't write the result back to `self`.
}
```
In this PR, we fix the above code so that operations do indeed happen *in* place.
```rust
fn square_in_place(&mut self) {
	let a = &mut (self.0).0; // Now `a` is a mutable reference to the contents of `self`.
	// do assembly stuff with `a`
}
```

---

write a little note why.

- [x] Wrote unit tests

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -108,6 +108,17 @@ jobs:
                    --exclude ark-poly-benches \
                    --exclude ark-test-templates"
 
+      - name: Test assembly on nightly
+        env:
+          RUSTFLAGS: -C target-feature=+bmi2,+adx
+        uses: actions-rs/cargo@v1
+        with:
+            command: test
+            args: "--workspace \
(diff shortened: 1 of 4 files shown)


weikengchen (approved):
Looks good.

jon-chuang:
Possible alternate fix: `ctx.add_declaration("a", "r", "a");` -> `ctx.add_declaration("a", "r", "&mut a");

PR #194, opened 2021-01-31
Made ProjectiveCurve CanonicalSerialize and CanonicalDeserialize

rozbb:
This makes serializing `GIPAProof` (from the `ripp` crate) much easier, for example. This serializing mechanism is not as fast as batch normalization, but we can document that detail elsewhere and let people benefit from the new ergonomics.


weikengchen (approved):
Looks good.

weikengchen:
Thanks for the PR.

PR #197, opened 2021-02-04
Add a BN384 curve for testing mixed-radix (and UInt128)

weikengchen:
## Description

Currently, the test curve in this repo is MNT4-753, which has two problems: 
- it is too big.
- when using it to test mixed-radix, it is slow since we need to use big polynomials because MNT4-753's two-adicity is still not low.

closes: #118 

We still keep MNT4-753 here because `r1cs-std` uses it. When `r1cs-std` is updated to not use it, we might consider deleting it (note: this might be a breaking change).

---

- [x] Wrote unit tests


test-curves/src/bn384_small_two_adicity/g1.rs
```diff
@@ -0,0 +1,47 @@
...
+impl ModelParameters for Parameters {
+    type BaseField = Fq;
+    type ScalarField = Fr;
+}
+
+impl SWModelParameters for Parameters {
+    /// COEFF_A = 0
+    const COEFF_A: Fq = field_new!(Fq, "0");
+
+    /// COEFF_B = 3
```
Pratyush, line 24:
```suggestion
    /// COEFF_B = 17
```

test-curves/src/bn384_small_two_adicity/g1.rs
```diff
@@ -0,0 +1,47 @@
...
+    fn mul_by_a(_: &Self::BaseField) -> Self::BaseField {
+        Self::BaseField::zero()
+    }
+}
+
+/// G1_GENERATOR_X = 1
+pub const G1_GENERATOR_X: Fq = field_new!(Fq, "1");
+
+/// G1_GENERATOR_Y = 2
+pub const G1_GENERATOR_Y: Fq = field_new!(Fq, "1564103007291248495620802981106927637056875497719363782670466283602827039808131416274891289504983049472454474542799");
```
Pratyush, line 47: Why is this generator's y-coordinate so large?

weikengchen: So this is because b=17 is the first value that satisfies the criteria (note: if b=3 satisfies we would have a simple generator, but nope). And the y coordinate here is the sqrt root of 18.

weikengchen: And I think clearly I should fix the comment which says 2 and also add a note that this is sqrt 18

weikengchen: updated

Pratyush (approved):

test-curves/src/bn384_small_two_adicity/g1.rs
Pratyush: Maybe we can set x = -1? That way y = 4

test-curves/src/bn384_small_two_adicity/g1.rs
```diff
@@ -0,0 +1,47 @@
...
+
+#[derive(Clone, Default, PartialEq, Eq)]
+pub struct Parameters;
+
+impl ModelParameters for Parameters {
+    type BaseField = Fq;
+    type ScalarField = Fr;
+}
+
+impl SWModelParameters for Parameters {
```
ValarDragon, line 20: Do we have any plans for the elliptic curve? Or is it included just in case we want integration tests with mixed radix fields

ValarDragon (approved):
LGTM, looking forward to faster tests!

test-curves/src/bn384_small_two_adicity/g1.rs
weikengchen: I just dropped it in for convenience.

test-curves/src/bn384_small_two_adicity/g1.rs
weikengchen: Yes. You are right. It is valid.

weikengchen: Wait...let me check the order of the point more carefully :)

weikengchen: Yes.

PR #199, opened 2021-02-04
Unroll loops and use in-place ops for faster `ff` and `ec` arithmetic

Pratyush:
## Description

* *BigInteger*: unroll loops in `add_nocarry`, `sub_noborrow`, `mul2`, `div2`, and `cmp`
* **SW** & **TE**: use in-place ops in mixed addition and doubling
* Extension fields: use in-place ops in mul and square

Overall, these changes provide a ~10% speedup to the relevant ops.

---

write a little note why.

- [x] Wrote unit tests

### DIFF
--- a/ec/src/models/short_weierstrass_jacobian.rs
+++ b/ec/src/models/short_weierstrass_jacobian.rs
@@ -260,7 +260,6 @@ impl<P: Parameters> Default for GroupAffine<P> {
 #[derivative(
     Copy(bound = "P: Parameters"),
     Clone(bound = "P: Parameters"),
-    Eq(bound = "P: Parameters"),
     Debug(bound = "P: Parameters"),
     Hash(bound = "P: Parameters")
 )]
@@ -279,6 +278,7 @@ impl<P: Parameters> Display for GroupProjective<P> {
     }
 }
 
+impl<P: Parameters> Eq for GroupProjective<P> {}
 impl<P: Parameters> PartialEq for GroupProjective<P> {
     fn eq(&self, other: &Self) -> bool {
         if self.is_zero() {
@@ -581,10 +581,9 @@ impl<'a, P: Parameters> Add<&'a Self> for GroupProjective<P> {
     type Output = Self;
 
     #[inline]
-    fn add(self, other: &'a Self) -> Self {
-        let mut copy = self;
-        copy += other;
-        copy
+    fn add(mut self, other: &'a Self) -> Self {
+        self += other;
+        self
     }
 }
 
@@ -657,10 +656,9 @@ impl<'a, P: Parameters> Sub<&'a Self> for GroupProjective<P> {
     type Output = Self;
 
     #[inline]
-    fn sub(self, other: &'a Self) -> Self {
-        let mut copy = self;
-        copy -= other;
-        copy
+    fn sub(mut self, other: &'a Self) -> Self {
+        self -= other;
+        self
     }
 }
 
--- a/ec/src/models/short_weierstrass_jacobian.rs
+++ b/ec/src/models/short_weierstrass_jacobian.rs
@@ -121,8 +121,15 @@ impl<P: Parameters> GroupAffine<P> {
             true
         } else {
             // Check that the point is on the curve
-            let y2 = self.y.square();
-            let x3b = P::add_b(&((self.x.square() * &self.x) + &P::mul_by_a(&self.x)));
+            // y2 = y^2
+            let mut y2 = self.y;
+            y2.square_in_place();
+            // x3b = x^3 + Ax + b
+            let mut x3b = self.x;
+            x3b.square_in_place();
+            x3b *= &self.x;
+            x3b += P::mul_by_a(&self.x);
+            x3b = P::add_b(&x3b);
             y2 == x3b
         }
     }
@@ -292,13 +299,22 @@ impl<P: Parameters> PartialEq for GroupProjective<P> {
         // The points (X, Y, Z) and (X', Y', Z')
         // are equal when (X * Z^2) = (X' * Z'^2)
         // and (Y * Z^3) = (Y' * Z'^3).
-        let z1z1 = self.z.square();
-        let z2z2 = other.z.square();
+        let mut z1z1 = self.z;
+        z1z1.square_in_place();
+        let mut z2z2 = other.z;
+        z2z2.square_in_place();
 
         if self.x * &z2z2 != other.x * &z1z1 {
             false
         } else {
-            self.y * &(z2z2 * &other.z) == other.y * &(z1z1 * &self.z)
+            let mut lhs = self.y;
+            lhs *= z2z2;
+            lhs *= &other.z;
+            let mut rhs = other.y;
+            rhs *= z1z1;
+            rhs *= &self.z;
+
+            lhs == rhs
         }
     }
 }
@@ -430,63 +446,98 @@ impl<P: Parameters> ProjectiveCurve for GroupProjective<P> {
 
         if P::COEFF_A.is_zero() {
             // A = X1^2
-            let mut a = self.x.square();
+            let mut a = self.x;
+            a.square_in_place();
 
             // B = Y1^2
-            let b = self.y.square();
+            let mut b = self.y;
+            b.square_in_place();
 
             // C = B^2
-            let mut c = b.square();
+            let mut c = b;
+            c.square_in_place();
 
             // D = 2*((X1+B)2-A-C)
-            let d = ((self.x + &b).square() - &a - &c).double();
+            let mut d = self.x;
+            d += &b;
+            d.square_in_place();
+            d -= &a;
+            d -= &c;
+            d.double_in_place();
 
             // E = 3*A
-            let e = a + &*a.double_in_place();
+            let mut e = a;
+            e += &*a.double_in_place();
 
             // F = E^2
-            let f = e.square();
+            let mut f = e;
+            f.square_in_place();
 
             // Z3 = 2*Y1*Z1
             self.z *= &self.y;
             self.z.double_in_place();
 
             // X3 = F-2*D
-            self.x = f - &d - &d;
+            self.x = -d;
+            self.x.double_in_place();
+            self.x += f;
 
             // Y3 = E*(D-X3)-8*C
-            self.y = (d - &self.x) * &e - &*c.double_in_place().double_in_place().double_in_place();
+            self.y = d;
+            self.y -= &self.x;
+            self.y *= e;
+            self.y -= &*c.double_in_place().double_in_place().double_in_place();
             self
         } else {
             // http://www.hyperelliptic.org/EFD/g1p/auto-shortw-jacobian-0.html#doubling-dbl-2009-l
             // XX = X1^2
-            let xx = self.x.square();
+            let mut xx = self.x;
+            xx.square_in_place();
 
             // YY = Y1^2
-            let yy = self.y.square();
+            let mut yy = self.y;
+            yy.square_in_place();
 
             // YYYY = YY^2
-            let mut yyyy = yy.square();
+            let mut yyyy = yy;
+            yyyy.square_in_place();
 
             // ZZ = Z1^2
-            let zz = self.z.square();
+            let mut zz = self.z;
+            zz.square_in_place();
 
             // S = 2*((X1+YY)^2-XX-YYYY)
-            let s = ((self.x + &yy).square() - &xx - &yyyy).double();
+            let mut s = self.x;
+            s += &yy;
+            s.square_in_place();
+            s -= &xx;
+            s -= &yyyy;
+            s.double_in_place();
 
             // M = 3*XX+a*ZZ^2
-            let m = xx + &xx + &xx + &P::mul_by_a(&zz.square());
+            let mut m = xx;
+            m.double_in_place();
+            m += xx;
+            m += &P::mul_by_a(&zz.square());
 
             // T = M^2-2*S
-            let t = m.square() - &s.double();
+            let mut t = m;
+            t.square_in_place();
+            t -= &s.double();
 
             // X3 = T
             self.x = t;
-            // Y3 = M*(S-T)-8*YYYY
-            let old_y = self.y;
-            self.y = m * &(s - &t) - &*yyyy.double_in_place().double_in_place().double_in_place();
             // Z3 = (Y1+Z1)^2-YY-ZZ
-            self.z = (old_y + &self.z).square() - &yy - &zz;
+            self.z += &self.y;
+            self.z.square_in_place();
+            self.z -= &yy;
+            self.z -= &zz;
+            // Y3 = M*(S-T)-8*YYYY
+            self.y = s;
+            self.y -= &t;
+            self.y *= m;
+            self.y -= &*yyyy.double_in_place().double_in_place().double_in_place();
+
             self
         }
     }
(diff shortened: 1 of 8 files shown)


ec/src/models/short_weierstrass_jacobian.rs
```diff
@@ -430,63 +430,98 @@ impl<P: Parameters> ProjectiveCurve for GroupProjective<P> {
...
+            f.square_in_place();
 
             // Z3 = 2*Y1*Z1
             self.z *= &self.y;
             self.z.double_in_place();
 
             // X3 = F-2*D
-            self.x = f - &d - &d;
+            self.x = -d;
+            self.x.double_in_place();
```
ValarDragon, line 466: Perhaps we should eventually bench `double_in_place_neg` ala https://github.com/ConsenSys/gurvy/blob/develop/bls381/g1.go#L659 (it just changes the last instruction of `double`). Its not clear to me that its worth it though, since it hurts the instruction cache.

ValarDragon: moved to new issue

ff/src/fields/macros.rs
```diff
@@ -618,43 +618,39 @@ macro_rules! impl_Fp {
             type Output = Self;
 
             #[inline]
-            fn add(self, other: &Self) -> Self {
-                let mut result = self.clone();
-                result.add_assign(other);
-                result
+            fn add(mut self, other: &Self) -> Self {
```
ValarDragon, line 621: I'm confused as to why this is valid, I thought `Add` isn't supposed to mutate the base

Pratyush: Here `self` is passed by value (so either the input gets moved, or copied):
```rust
let a = 2;
let c = a.add(&b); // here `a` is copied
print!(a); // will print 2
```

ff/src/fields/models/cubic_extension.rs
```diff
@@ -468,16 +488,36 @@ impl<'a, P: CubicExtParameters> MulAssign<&'a Self> for CubicExtField<P> {
...
-        self.c0 = ad + &P::mul_base_field_by_nonresidue(&x);
-        self.c1 = y + &P::mul_base_field_by_nonresidue(&cf);
+        let mut ad = d;
+        ad *= &a;
+        let mut be = e;
+        be *= &b;
+        let mut cf = f;
+        cf *= &c;
+
+        let mut x = e;
```
ValarDragon, line 494:
```suggestion
         // x = (e + f) * (b + c) - be - cf;
         let mut x = e;
```

ff/src/fields/models/cubic_extension.rs
```diff
@@ -468,16 +488,36 @@ impl<'a, P: CubicExtParameters> MulAssign<&'a Self> for CubicExtField<P> {
...
+        x += &f;
+        x *= &(b + &c);
+        x -= &be;
+        x -= &cf;
+
+        let mut y = d;
```
ValarDragon, line 500:
```suggestion
         // y = (d + e) * (a + b) - ad - be;
         let mut y = d;
```

ff/src/fields/models/cubic_extension.rs
```diff
@@ -468,16 +488,36 @@ impl<'a, P: CubicExtParameters> MulAssign<&'a Self> for CubicExtField<P> {
...
+        y += &e;
+        y *= &(a + &b);
+        y -= &ad;
+        y -= &be;
+
+        let mut z = d;
```
ValarDragon, line 506:
```suggestion
         // z = (d + f) * (a + c) - ad + be - cf;
         let mut z = d;
```

ff/src/fields/macros.rs
ValarDragon: Oh cool, that makes sense!

ValarDragon:
LGTM, for the extension fields. Though `inverse_in_place` could have this optimization applied to it, if the `inverse` logic is moved to `inverse_in_place`.

Pratyush:
This is ready for review.

ec/src/models/short_weierstrass_jacobian.rs
```diff
@@ -121,8 +121,13 @@ impl<P: Parameters> GroupAffine<P> {
             true
         } else {
             // Check that the point is on the curve
-            let y2 = self.y.square();
-            let x3b = P::add_b(&((self.x.square() * &self.x) + &P::mul_by_a(&self.x)));
+            let mut y2 = self.y;
```
ValarDragon, line 124:
```suggestion
            // y2 = y^2
            let mut y2 = self.y;
```

ec/src/models/short_weierstrass_jacobian.rs
```diff
@@ -121,8 +121,13 @@ impl<P: Parameters> GroupAffine<P> {
...
+            y2.square_in_place();
+            let mut x3b = self.x;
```
ValarDragon, line 126:
```suggestion
            // x3b = x^3 + Ax + b
            let mut x3b = self.x;
```

ec/src/models/short_weierstrass_jacobian.rs
```diff
@@ -260,7 +265,6 @@ impl<P: Parameters> Default for GroupAffine<P> {
 #[derivative(
     Copy(bound = "P: Parameters"),
     Clone(bound = "P: Parameters"),
-    Eq(bound = "P: Parameters"),
```
ValarDragon, line 263: Why did this get removed?

ff/src/fields/models/cubic_extension.rs
```diff
@@ -372,8 +389,11 @@ impl<P: CubicExtParameters> FromBytes for CubicExtField<P> {
 impl<P: CubicExtParameters> Neg for CubicExtField<P> {
     type Output = Self;
     #[inline]
-    fn neg(self) -> Self {
-        Self::new(self.c0.neg(), self.c1.neg(), self.c2.neg())
+    fn neg(mut self) -> Self {
+        self.c0 = -self.c0;
```
ValarDragon, line 393: should this instead be calling neg on the base field?

jon-chuang:
Hi @Pratyush , thanks for this quick follow up.

I did a bench of this against simply implementing the unroll to big integer, here are the results:
Screenshot from 2021-02-05 08-55-09


Thanks for catching cmp as being a helpful change. This improved things further over my original changes.

However, I question the changes to the semantics of the non-assigning versions of the ops. At the most, I would remove `.clone()`, as was done for modulus. Further, it appears, according to the benchmark, that changing the formulas to use only assigning versions of the ops has little or possibly even negative effect.

ff/src/fields/models/cubic_extension.rs
Pratyush: `neg` is just the method name corresponding to `-`.

ec/src/models/short_weierstrass_jacobian.rs
Pratyush: This got moved to a proper impl, because you shouldn't have a manual impl of `PartialEq` and a derived impl of `Eq`.

Pratyush:
> However, I question the changes to the semantics of the non-assigning versions of the ops. At the most, I would remove .clone(), as was done for modulus.

The semantics didn't change; I only removed unnecessary copies. In particular, the non-assigning versions take `self` by value, not by reference.

> Further, it appears, according to the benchmark, that changing the formulas to use only assigning versions of the ops has little or possibly even negative effect.

That's surprising, I definitely benchmarking both the `assign` versions against the old versions.

> @Pratyush Would you mind giving me partial access to this repo so I could push commits to a branch on the repo? It would simplify things for me so I wouldn't have to keep switching my git remote to target different urls.

Let me figure that out

jon-chuang:
> Let me figure that out. In the mean time a simpler way to handle it would be to just change the patch location
Hmm actually this is orthogonal to that issue (since deps target master, and one still has to make changes to target branches). I'm just thinking in terms of friction in PR workflow. It's a small thing. 

Wrt the cross-dependency issue, I think patch doesn't fix the problem, I'll investigate if there is a good way to fix it.

> That's surprising, I definitely benchmarked the assign versions against the old versions, and found a non-negligible difference.
 
Are you saying that the assign versions without the biginteger unroll changes produced an improvement?

> The semantics didn't change; I only removed unnecessary copies. In particular, the non-assigning versions take self by value, not by reference.

Could you clarify this? I meant that the semantics have changed due to mutating the underlying variable now. 

To clarify: 
> I question the changes to the semantics of the non-assigning versions of the ops, which are changed to mutate the underlying variable. From an API standpoint, I think this is confusing, further, it achieves nothing over just `x.op_assign(); let y = x;`

My point essentially is that the old non assigning APIs don't have to be changed, even if one were to convert all group formulas to use assigning versions. Which is orthogonal to my point that it seems those conversions don't appear to help, at least according to the benchmarks I performed.

Pratyush:
> Hmm actually this is orthogonal to that issue (since deps target master, and one still has to make changes to target branches). I'm just thinking in terms of friction in PR workflow. It's a small thing.

Ah I just point the `patch` to a local clone of the relevant repos, so that I don't have to change branches and such. That way I can switch out the paths easily.

ValarDragon:
Here is my patch in curves, which works quite nice for me:
```
[patch.'https://github.com/arkworks-rs/algebra']
ark-ff = { path = '../algebra/ff' }
ark-ec = { path = '../algebra/ec' }
ark-serialize = { path = '../algebra/serialize' }
```

ValarDragon:
Also I re-benchmarked this, and confirmed that I only saw operation times go down for all operations benchmarked. (Between 2 and 10% reduction depending on the operation)

ValarDragon (approved):
LGTM sans comment requests. Thanks for updating this!

ValarDragon:
wait what happened in this force push

Pratyush:
Sorry, I just rebased on master, though I will be making a clean up of the history also soon, to separate out commits that change just the unrolling and commits that additionally use `in_place` ops.

Pratyush:
Ok @jon-chuang @ValarDragon seems like the in place ops don't actually get you a big benefit; I can't reproduce whatever I had earlier (which might well have been wrong). In light of that, I think it makes sense keep only the first three commits (loop unrolling and reducing copies), as the in place ops make the code more difficult to read.

<details>
<summary>Fields ops</summary>
```
 name                                    only_unroll ns/iter  in_place_ops ns/iter  diff ns/iter   diff %  speedup 
 bls12_381::fq12::add_assign             144                  142                             -2   -1.39%   x 1.01 
 bls12_381::fq12::deser                  835                  859                             24    2.87%   x 0.97 
 bls12_381::fq12::deser_unchecked        835                  859                             24    2.87%   x 0.97 
 bls12_381::fq12::double                 123                  128                              5    4.07%   x 0.96 
 bls12_381::fq12::inverse                17,497               18,103                         606    3.46%   x 0.97 
 bls12_381::fq12::mul_assign             3,960                4,039                           79    1.99%   x 0.98 
 bls12_381::fq12::negate                 141                  137                             -4   -2.84%   x 1.03 
 bls12_381::fq12::ser                    506                  504                             -2   -0.40%   x 1.00 
 bls12_381::fq12::ser_unchecked          520                  509                            -11   -2.12%   x 1.02 
 bls12_381::fq12::square                 2,788                2,744                          -44   -1.58%   x 1.02 
 bls12_381::fq12::sub_assign             146                  140                             -6   -4.11%   x 1.04 
 bls12_381::fq2::add_assign              13                   13                               0    0.00%   x 1.00 
 bls12_381::fq2::deser                   127                  127                              0    0.00%   x 1.00 
 bls12_381::fq2::deser_unchecked         126                  124                             -2   -1.59%   x 1.02 
 bls12_381::fq2::double                  12                   13                               1    8.33%   x 0.92 
 bls12_381::fq2::inverse                 11,427               11,210                        -217   -1.90%   x 1.02 
 bls12_381::fq2::mul_assign              132                  131                             -1   -0.76%   x 1.01 
 bls12_381::fq2::negate                  15                   13                              -2  -13.33%   x 1.15 
 bls12_381::fq2::ser                     87                   84                              -3   -3.45%   x 1.04 
 bls12_381::fq2::ser_unchecked           85                   84                              -1   -1.18%   x 1.01 
 bls12_381::fq2::sqrt                    78,645               77,164                      -1,481   -1.88%   x 1.02 
 bls12_381::fq2::square                  105                  107                              2    1.90%   x 0.98 
 bls12_381::fq2::sub_assign              15                   15                               0    0.00%   x 1.00 
 bls12_381::fq::add_assign               8                    8                                0    0.00%   x 1.00 
 bls12_381::fq::deser                    61                   61                               0    0.00%   x 1.00 
 bls12_381::fq::deser_unchecked          61                   61                               0    0.00%   x 1.00 
 bls12_381::fq::double                   6                    6                                0    0.00%   x 1.00 
 bls12_381::fq::from_repr                40                   39                              -1   -2.50%   x 1.03 
 bls12_381::fq::into_repr                29                   29                               0    0.00%   x 1.00 
 bls12_381::fq::inverse                  10,880               10,880                           0    0.00%   x 1.00 
 bls12_381::fq::mul_assign               37                   37                               0    0.00%   x 1.00 
 bls12_381::fq::negate                   7                    7                                0    0.00%   x 1.00 
 bls12_381::fq::repr_add_nocarry         5                    5                                0    0.00%   x 1.00 
 bls12_381::fq::repr_div2                2                    2                                0    0.00%   x 1.00 
 bls12_381::fq::repr_mul2                2                    2                                0    0.00%   x 1.00 
 bls12_381::fq::repr_num_bits            2                    2                                0    0.00%   x 1.00 
 bls12_381::fq::repr_sub_noborrow        2                    3                                1   50.00%   x 0.67 
 bls12_381::fq::ser                      43                   43                               0    0.00%   x 1.00 
 bls12_381::fq::ser_unchecked            43                   43                               0    0.00%   x 1.00 
 bls12_381::fq::sqrt                     19,079               19,263                         184    0.96%   x 0.99 
 bls12_381::fq::square                   37                   37                               0    0.00%   x 1.00 
 bls12_381::fq::sub_assign               9                    8                               -1  -11.11%   x 1.12 
 ed_on_bls12_381::fq::add_assign         4                    4                                0    0.00%   x 1.00 
 ed_on_bls12_381::fq::deser              30                   31                               1    3.33%   x 0.97 
 ed_on_bls12_381::fq::deser_unchecked    30                   31                               1    3.33%   x 0.97 
 ed_on_bls12_381::fq::double             4                    4                                0    0.00%   x 1.00 
 ed_on_bls12_381::fq::from_repr          24                   24                               0    0.00%   x 1.00 
 ed_on_bls12_381::fq::into_repr          4                    4                                0    0.00%   x 1.00 
 ed_on_bls12_381::fq::inverse            5,405                5,098                         -307   -5.68%   x 1.06 
 ed_on_bls12_381::fq::mul_assign         21                   21                               0    0.00%   x 1.00 
 ed_on_bls12_381::fq::negate             5                    5                                0    0.00%   x 1.00 
 ed_on_bls12_381::fq::repr_add_nocarry   4                    4                                0    0.00%   x 1.00 
 ed_on_bls12_381::fq::repr_div2          2                    2                                0    0.00%   x 1.00 
 ed_on_bls12_381::fq::repr_mul2          2                    3                                1   50.00%   x 0.67 
 ed_on_bls12_381::fq::repr_num_bits      2                    3                                1   50.00%   x 0.67 
 ed_on_bls12_381::fq::repr_sub_noborrow  2                    2                                0    0.00%   x 1.00 
 ed_on_bls12_381::fq::ser                18                   18                               0    0.00%   x 1.00 
 ed_on_bls12_381::fq::ser_unchecked      19                   18                              -1   -5.26%   x 1.06 
 ed_on_bls12_381::fq::sqrt               12,672               12,640                         -32   -0.25%   x 1.00 
 ed_on_bls12_381::fq::square             21                   20                              -1   -4.76%   x 1.05 
 ed_on_bls12_381::fq::sub_assign         6                    5                               -1  -16.67%   x 1.20 
```
</details>

<details>
<summary>Group ops</summary>
```
name                             only_unroll_g1 ns/iter  in_place_ops_g1 ns/iter  diff ns/iter  diff %  speedup 
 bls12_381::g1::add_assign        655                     661                                 6   0.92%   x 0.99 
 bls12_381::g1::add_assign_mixed  486                     481                                -5  -1.03%   x 1.01 
 bls12_381::g1::deser             148,522                 149,175                           653   0.44%   x 1.00 
 bls12_381::g1::deser_unchecked   131                     132                                 1   0.76%   x 0.99 
 bls12_381::g1::double            333                     343                                10   3.00%   x 0.97 
 bls12_381::g1::msm_131072        1,342,187,320           1,334,606,002              -7,581,318  -0.56%   x 1.01 
 bls12_381::g1::mul_assign        166,342                 175,056                         8,714   5.24%   x 0.95 
 bls12_381::g1::rand              103,346                 105,790                         2,444   2.36%   x 0.98 
 bls12_381::g1::ser               105                     108                                 3   2.86%   x 0.97 
 bls12_381::g1::ser_unchecked     84                      86                                  2   2.38%   x 0.98 
```
</details>

Issue #198, opened 2021-02-04
Investigate ways to improve basic primitive speed


jon-chuang:
With some simple changes converting biginteger primitive iter ops to their unrolled counterparts, one obtains some decent results.

Hence, the hypothesis is that assembly could improve these results further.

Here are the results (with some anomalies I believe due to overly aggressive optimisations in the benchmark code):
![Screenshot from 2021-02-04 18-51-34](https://user-images.githubusercontent.com/9093549/106882917-41761380-671a-11eb-9c85-aa75cacd8cb4.png)
![Screenshot from 2021-02-04 18-49-26](https://user-images.githubusercontent.com/9093549/106882922-42a74080-671a-11eb-9869-5a52d69eec9c.png)

Pratyush:
See https://github.com/arkworks-rs/algebra/pull/199

I think https://github.com/rust-lang/stdarch/issues/666 also affects addition and subtraction, so we end up with non-optimal assembly:

![`add_nocarry` assembly](https://user-images.githubusercontent.com/3220730/106938907-68a9f200-66d4-11eb-96d6-7a062115a0c4.png)

jon-chuang:
Hmm, although this (adcx, adox etc) should affect mul due to requiring two carry chains, in theory, llvm can deal with single carry chains (as with adc). 

However, there are many questionable instructions here, such as `movzx` and `setb`. This could be artifacts from using u128. Hence, to obtain a cleaner assembly, I will investigate simply implementing assembly versions of the key functions.

I may also investigate if it is worth implementing cmp_less and cmp_greater functions to reduce the generality bloat required to define cmp.

Pratyush:
In that case, maybe we can just directly use the intrinsic? https://doc.rust-lang.org/core/arch/x86_64/fn._addcarryx_u64.html

jon-chuang:
Possibly. 

However, since we are already using assembly, it may pay to have greater control over the assembly that is produced.

For instance for `sub_noborrow` as well.

jon-chuang:
I think I will revamp the assembly framework so there are less weird artifacts from the time I went a little crazy with the proc macro stuff and trying to have a DSL syntax.

Overall, I think it's an ok alternative to generating .S, but now it's rather chaotic and ad hoc.

Pratyush:
For sub_noborrow, we have https://doc.rust-lang.org/core/arch/x86_64/fn._subborrow_u64.html

jon-chuang:
[...]

PR #204, opened 2021-02-05
Improve primitive biginteger operations

jon-chuang:
## Description

Details and benchmarks discussed here: https://github.com/arkworks-rs/algebra/issues/198
Related PR: https://github.com/arkworks-rs/algebra/pull/199, now #205 

---

write a little note why.

- [x] Wrote unit tests


weikengchen:
@Pratyush I think #205 and this one is solving the same issue and could be reconciled together.

ff/src/fields/arithmetic.rs
```diff
@@ -26,7 +26,6 @@ macro_rules! impl_field_mul_assign {
                 {
                     // Tentatively avoid using assembly for `$limbs == 1`.
                     if $limbs <= 6 && $limbs > 1 {
-                        assert!($limbs <= 6);
```
ValarDragon, line 29: should these just be `debug_assert`

jon-chuang: Well, it doesn't really achieve anything either way since the code will never execute and the assert will never fail.

I think pratyush add these things sometimes hoping that it would help rid of bounds checks that rustc introduces when indexing into slices. But, I have a feeling that indexing into arrays these bounds checks don't happen. 

Further still, in our case we are implementing assembly, so no such bounds checks exist by default.

So they should be removed.

ValarDragon: Alot of users do compile without assembly, since its inconvenient to do so. This isn't getting called in the assembly code path right?

ValarDragon: Edit: Sorry you're right this should definitely be removed. Didn't see the use_asm tag right above it. Great catch

ff/src/biginteger/macros.rs
```diff
@@ -119,8 +164,14 @@ macro_rules! bigint_impl {
             }
 
             #[inline]
+            #[ark_ff_asm::unroll_for_loops]
```
ValarDragon, line 167: I think we need to benchmark this change on larger fields as well?

jon-chuang: Yes, you are probably right. This will currently involve changing a lot of Cargo.toml files to point to my local repo. I'll try to fix some of the issues I find in this process.

jon-chuang:
Things I am not sure about: 
- necessity of removing derive(Eq)?

ValarDragon (commented):
All of these changes LGTM. Holding off on approving until we have confirmed performance improvements of always unrolling on varying field sizes.

Pratyush:
@ValarDragon this PR is subsumed by #205, minus the `is_zero` change (assembly is more compact for the iterator version at higher number of limbs, and at lower number of limbs it's unrolled)

ValarDragon:
I thought this PR also removed more `assert`s? I saw two removed here, but only one in the other PR. Let me recount though.

ValarDragon:
Recounted, it is indeed not fully subsumed. The assert for `mul_in_place` removed here is not removed in #205 atm.

jon-chuang:
closed in favour of #205

jon-chuang:
closed in favour of #205

PR #205, opened 2021-02-05
Unroll biginteger loops and reduce copies

Pratyush:
## Description

Replaces #199, motivation from #198

---

write a little note why.

- [x] Wrote unit tests

### DIFF
--- a/ff/src/biginteger/macros.rs
+++ b/ff/src/biginteger/macros.rs
@@ -12,35 +12,38 @@ macro_rules! bigint_impl {
         impl BigInteger for $name {
             const NUM_LIMBS: usize = $num_limbs;
 
-            #[inline]
+            #[ark_ff_asm::unroll_for_loops]
             fn add_nocarry(&mut self, other: &Self) -> bool {
                 let mut carry = 0;
 
-                for (a, b) in self.0.iter_mut().zip(other.0.iter()) {
-                    *a = adc!(*a, *b, &mut carry);
+                for i in 0..$num_limbs {
+                    self.0[i] = adc!(self.0[i], other.0[i], &mut carry);
                 }
 
                 carry != 0
             }
 
-            #[inline]
+            #[ark_ff_asm::unroll_for_loops]
             fn sub_noborrow(&mut self, other: &Self) -> bool {
                 let mut borrow = 0;
 
-                for (a, b) in self.0.iter_mut().zip(other.0.iter()) {
-                    *a = sbb!(*a, *b, &mut borrow);
+                for i in 0..$num_limbs {
+                    self.0[i] = sbb!(self.0[i], other.0[i], &mut borrow);
                 }
 
                 borrow != 0
             }
 
             #[inline]
+            #[ark_ff_asm::unroll_for_loops]
+            #[allow(unused)]
             fn mul2(&mut self) {
                 let mut last = 0;
-                for i in &mut self.0 {
-                    let tmp = *i >> 63;
-                    *i <<= 1;
-                    *i |= last;
+                for i in 0..$num_limbs {
+                    let a = &mut self.0[i];
+                    let tmp = *a >> 63;
+                    *a <<= 1;
+                    *a |= last;
                     last = tmp;
                 }
             }
@@ -72,12 +75,15 @@ macro_rules! bigint_impl {
             }
 
             #[inline]
+            #[ark_ff_asm::unroll_for_loops]
+            #[allow(unused)]
             fn div2(&mut self) {
                 let mut t = 0;
-                for i in self.0.iter_mut().rev() {
-                    let t2 = *i << 63;
-                    *i >>= 1;
-                    *i |= t;
+                for i in 0..$num_limbs {
+                    let a = &mut self.0[$num_limbs - i - 1];
+                    let t2 = *a << 63;
+                    *a >>= 1;
+                    *a |= t;
                     t = t2;
                 }
             }
@@ -270,16 +276,19 @@ macro_rules! bigint_impl {
 
         impl Ord for $name {
             #[inline]
+            #[ark_ff_asm::unroll_for_loops]
             fn cmp(&self, other: &Self) -> ::core::cmp::Ordering {
-                for (a, b) in self.0.iter().rev().zip(other.0.iter().rev()) {
+                use core::cmp::Ordering;
+                for i in 0..$num_limbs {
+                    let a = &self.0[$num_limbs - i - 1];
+                    let b = &other.0[$num_limbs - i - 1];
                     if a < b {
-                        return core::cmp::Ordering::Less;
+                        return Ordering::Less;
                     } else if a > b {
-                        return core::cmp::Ordering::Greater;
+                        return Ordering::Greater;
                     }
                 }
-
-                core::cmp::Ordering::Equal
+                Ordering::Equal
             }
         }
 
--- a/ff/src/biginteger/macros.rs
+++ b/ff/src/biginteger/macros.rs
@@ -12,13 +12,14 @@ macro_rules! bigint_impl {
         impl BigInteger for $name {
             const NUM_LIMBS: usize = $num_limbs;
 
+            #[inline]
             #[ark_ff_asm::unroll_for_loops]
             fn add_nocarry(&mut self, other: &Self) -> bool {
                 let mut carry = 0;
 
                 for i in 0..$num_limbs {
                     #[cfg(all(target_arch = "x86_64", feature = "asm"))]
-                    #[cfg_attr(all(target_arch = "x86_64", feature = "asm"), allow(unsafe_code))]
+                    #[allow(unsafe_code)]
                     unsafe {
                         carry = core::arch::x86_64::_addcarry_u64(
                             carry,
@@ -28,7 +29,7 @@ macro_rules! bigint_impl {
                         )
                     };
 
-                    #[cfg(not(feature = "asm"))]
+                    #[cfg(not(all(target_arch = "x86_64", feature = "asm")))]
                     {
                         self.0[i] = adc!(self.0[i], other.0[i], &mut carry);
                     }
@@ -37,13 +38,14 @@ macro_rules! bigint_impl {
                 carry != 0
             }
 
+            #[inline]
             #[ark_ff_asm::unroll_for_loops]
             fn sub_noborrow(&mut self, other: &Self) -> bool {
                 let mut borrow = 0;
 
                 for i in 0..$num_limbs {
                     #[cfg(all(target_arch = "x86_64", feature = "asm"))]
-                    #[cfg_attr(all(target_arch = "x86_64", feature = "asm"), allow(unsafe_code))]
+                    #[allow(unsafe_code)]
                     unsafe {
                         borrow = core::arch::x86_64::_subborrow_u64(
                             borrow,
@@ -53,7 +55,7 @@ macro_rules! bigint_impl {
                         )
                     };
 
-                    #[cfg(not(feature = "asm"))]
+                    #[cfg(not(all(target_arch = "x86_64", feature = "asm")))]
                     {
                         self.0[i] = sbb!(self.0[i], other.0[i], &mut borrow);
                     }
@@ -66,6 +68,24 @@ macro_rules! bigint_impl {
             #[ark_ff_asm::unroll_for_loops]
             #[allow(unused)]
             fn mul2(&mut self) {
+                #[cfg(all(target_arch = "x86_64", feature = "asm"))]
+                #[allow(unsafe_code)]
+                {
+                    let mut carry = 0;
+
+                    for i in 0..$num_limbs {
+                        unsafe {
+                            carry = core::arch::x86_64::_addcarry_u64(
+                                carry,
+                                self.0[i],
+                                self.0[i],
+                                &mut self.0[i],
+                            );
+                        }
+                    }
+                    return;
+                }
+
                 let mut last = 0;
                 for i in 0..$num_limbs {
                     let a = &mut self.0[i];
@@ -76,6 +96,7 @@ macro_rules! bigint_impl {
                 }
             }
 
+            #[inline]
             #[ark_ff_asm::unroll_for_loops]
             fn muln(&mut self, mut n: u32) {
                 if n >= 64 * $num_limbs {
@@ -118,6 +139,7 @@ macro_rules! bigint_impl {
                 }
             }
 
+            #[inline]
             #[ark_ff_asm::unroll_for_loops]
             fn divn(&mut self, mut n: u32) {
                 if n >= 64 * $num_limbs {
@@ -158,7 +180,12 @@ macro_rules! bigint_impl {
 
             #[inline]
             fn is_zero(&self) -> bool {
-                self.0.iter().all(|&e| e == 0)
+                for i in 0..$num_limbs {
+                    if self.0[i] != 0 {
+                        return false;
+                    }
+                }
+                true
             }
 
             #[inline]
--- a/ff/src/biginteger/macros.rs
+++ b/ff/src/biginteger/macros.rs
@@ -21,12 +21,8 @@ macro_rules! bigint_impl {
                     #[cfg(all(target_arch = "x86_64", feature = "asm"))]
                     #[allow(unsafe_code)]
                     unsafe {
-                        carry = core::arch::x86_64::_addcarry_u64(
-                            carry,
-                            self.0[i],
-                            other.0[i],
-                            &mut self.0[i],
-                        )
+                        use core::arch::x86_64::_addcarry_u64;
+                        carry = _addcarry_u64(carry, self.0[i], other.0[i], &mut self.0[i])
                     };
 
                     #[cfg(not(all(target_arch = "x86_64", feature = "asm")))]
@@ -47,12 +43,8 @@ macro_rules! bigint_impl {
                     #[cfg(all(target_arch = "x86_64", feature = "asm"))]
                     #[allow(unsafe_code)]
                     unsafe {
-                        borrow = core::arch::x86_64::_subborrow_u64(
-                            borrow,
-                            self.0[i],
-                            other.0[i],
-                            &mut self.0[i],
-                        )
+                        use core::arch::x86_64::_subborrow_u64;
+                        borrow = _subborrow_u64(borrow, self.0[i], other.0[i], &mut self.0[i])
                     };
 
                     #[cfg(not(all(target_arch = "x86_64", feature = "asm")))]
@@ -75,24 +67,22 @@ macro_rules! bigint_impl {
 
                     for i in 0..$num_limbs {
                         unsafe {
-                            carry = core::arch::x86_64::_addcarry_u64(
-                                carry,
-                                self.0[i],
-                                self.0[i],
-                                &mut self.0[i],
-                            );
-                        }
+                            use core::arch::x86_64::_addcarry_u64;
+                            carry = _addcarry_u64(carry, self.0[i], self.0[i], &mut self.0[i])
+                        };
                     }
-                    return;
                 }
 
-                let mut last = 0;
-                for i in 0..$num_limbs {
-                    let a = &mut self.0[i];
-                    let tmp = *a >> 63;
-                    *a <<= 1;
-                    *a |= last;
-                    last = tmp;
+                #[cfg(not(all(target_arch = "x86_64", feature = "asm")))]
+                {
+                    let mut last = 0;
+                    for i in 0..$num_limbs {
+                        let a = &mut self.0[i];
+                        let tmp = *a >> 63;
+                        *a <<= 1;
+                        *a |= last;
+                        last = tmp;
+                    }
                 }
             }
 
--- a/ff/src/biginteger/macros.rs
+++ b/ff/src/biginteger/macros.rs
@@ -19,8 +19,13 @@ macro_rules! bigint_impl {
                 for i in 0..$num_limbs {
                     #[cfg(all(target_arch = "x86_64", feature = "asm"))]
                     #[cfg_attr(all(target_arch = "x86_64", feature = "asm"), allow(unsafe_code))]
-                    unsafe { 
-                        carry = core::arch::x86_64::_addcarry_u64(carry, self.0[i], other.0[i], &mut self.0[i]) 
+                    unsafe {
+                        carry = core::arch::x86_64::_addcarry_u64(
+                            carry,
+                            self.0[i],
+                            other.0[i],
+                            &mut self.0[i],
+                        )
                     };
 
                     #[cfg(not(feature = "asm"))]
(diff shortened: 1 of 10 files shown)


weikengchen (dismissed):
I skimmed the changes. Most of them look reasonable. I will wait for CI to do the careful check.

jon-chuang:
I still don't understand the purpose of this PR. Why are the changes to the non assign ops made? I was trying to establish that these changes have no worth and are in fact detractive, however, these complaints have not been addressed.

One can already achieve everything one would like in the new versions of the non-assigning ops with the assigning versions. Previously, non-assigning ops acted as syntactic sugar. Now, I am unconvinced one should ever use them.

jon-chuang:
Why would you need to assign the result of sub to a new variable, when you have the old variable denoted by self that holds that very value? That makes absolutely no sense. And yet this got an approving review, whereas a PR that I made that did not have such absurd changes and had been thoroughly benched did not get any review.

jon-chuang:
I'm starting to feel that Zexe is not very open source at all. Increasingly, I feel less and less like it is worth contributing to.

weikengchen:
I apologize that I was more focusing on the changes that clean up nonnative, poly-commit, ... I will surely take a look at the min const generics PR soon.

It is also related to the number of changes---small PRs are easier to reconcile. Our PRs in poly-commit and marlin, which add the constraints for Marlin verifier, are still there---because we want to avoid conflicts with other PRs and projects.

jon-chuang:
Hi ok sorry, anw I wasn't referring to min-const-generics, I think that one will be binned, and it's low priority. 

It's just there seems no reason why some people have more authority than others to push to branches, approve reviews, put labels. Makes little sense.

weikengchen:
It has one thing to do with the challenging balance between an open-sourced platform and an academic prototype---sometimes we need to wrap up something within a time limit---and there is probably why breaking changes are happening more often than a stable library should. It is a slightly challenging balance that we are maintaining.

I want to assure you that when I approve this one, I was just being requested for a review, and I just did a sanity check. You may also request a review or @ us. 

We are sort of overwhelmed with many projects---and for me, beyond arkworks, I am also working on other non-zero-knowledge stuff, so pardon me for being only able to maintain a slight balance.

weikengchen:
I think there is something that we can improve---for example, as you point out, the ability to request a review should be open.

Again please pardon that it is also a preliminary step, and many things can be changed to improve it!

jon-chuang:
Hi yes, I definitely understand that, anw I think Ive not been too communicative, I will try to communicate more. 

Actually I do appreciate the work from the arkworks team of course. 

Anw I should chat more and figure out some solutions. Anw I think this is a case of just pratyush and I developing parallel work, which is confusing, I should chat more with pratyush in real time on telegram let's say since exchanging comments/PRs on issues can be a bit stilted and a bit too formalised. Since there are some fundamental differences.

weikengchen:
Yes. I feel it is related to the situation that Pratyush is unaware that you will be working on it. 

On the one hand, we are trying to get more people involved via `good first issue` because we still have to do many things ourselves. On the other hand, we are very fortunate to have you.

It is fortunate that we are working on zero-knowledge proofs that have a larger community. 

-----

I do receive emails for all the PRs but I usually focus on those that I am more familiar with (or bug reports). So, I am not perfect.

jon-chuang:
Hmm ok, but if it hasn't been reviewed carefully, I don't think it should be approved, no?

weikengchen:
I was just doing a sanity check---I am not an expert in asm, and the review process is mainly to prevent errors, add comments on things that are unclear. 

I agree that we should have let you review this and reconcile with #204.

weikengchen (commented):
TODO: Reconcile with #204 as there seem to be some subtle improvements.

ValarDragon:
> Why would you need to assign the result of sub to a new variable, when you have the old variable denoted by self that holds that very value? That makes absolutely no sense. And yet this got an approving review, whereas a PR that I made that did not have such absurd changes and had been thoroughly benched did not get any review.
> Hi ok sorry, anw I wasn't referring to min-const-generics, I think that one will be binned, and it's low priority. I'm referring to #204 which is a very small PR.

FWIW, both this PR and 204 were made while I was asleep lol. I don't think sub-24 review turnarounds are reasonable as people are strewn across many timezones and have multiple projects in parallel...

ValarDragon:
> Why would you need to assign the result of sub to a new variable, when you have the old variable denoted by self that holds that very value? That makes absolutely no sense.

Are you talking about 
```
    fn sub(mut self, other: &'a Self) -> Self {
        self -= other;
        self
    }
```

This seems like a great optimization given that the base is already copied when this function is called?

jon-chuang:
> > Why would you need to assign the result of sub to a new variable, when you have the old variable denoted by self that holds that very value? That makes absolutely no sense. And yet this got an approving review, whereas a PR that I made that did not have such absurd changes and had been thoroughly benched did not get any review.
> > Hi ok sorry, anw I wasn't referring to min-const-generics, I think that one will be binned, and it's low priority. I'm referring to #204 which is a very small PR.
> 
> FWIW, both this PR and 204 were made while I was asleep lol. I don't think sub-24 review turnarounds are a reasonable expectation to hold as people are strewn across many timezones and have multiple projects in parallel... (Nor do other open source projects give any such guarantee. Projects I've worked on in the past such as https://github.com/tendermint/tendermint/ definitely don't)

Well that's definitely not what I was expected, what really irked me is that this PR had an immediate approving review, even though my overlapping PR was posted earlier, further, there were many unexplained choices like the ones I am disputing.

So it seemed to me like a bit of group-think and disregard for evaluating PRs on their merits, and also neglect for others' work in favour of those of the maintainers.

jon-chuang:
> > Why would you need to assign the result of sub to a new variable, when you have the old variable denoted by self that holds that very value? That makes absolutely no sense.
> 
> Are you talking about
> 
> ```
>     fn sub(mut self, other: &'a Self) -> Self {
>         self -= other;
>         self
>     }
> ```
> 
> This seems like a great optimization given that the base is already copied when this function is called?

Well, in that case, one would simply do:
```
p.sub_assign(&q);
let y = p;
```
No one would ever write such code.
But this is the only use case I can think of for which the new implementations are syntactic sugar. Further, pratyush already confirmed there is no performance reason to change the way the group addition formulas are written and in fact I show that they are near the performance of gurvy, wherein the remaining discrepancies I believe lie in things light uneccessary data movement at function boundaries (whereas gurvy generates much more optimal assembly for these via scripts).

Hence, my conclusion is that there has been 0 argument for why these changes make any sense at all.

jon-chuang:
If one truly wanted to extend the current syntactic sugar, the appropriate function to modify is to append `self` to every `_assign` function. Think about it and tell me I'm right.

ValarDragon:
> Well that's definitely not what I was expected, what really irked me is that this PR had an immediate approving review, even though my overlapping PR was posted earlier, further, there were many unexplained choices like the ones I am disputing.
> 
> So it seemed to me like a bit of group-think and disregard for evaluating PRs on their merits, and also neglect for others' work in favour of those of the maintainers.

Can you assume some more positive intent on our part? We're talking about a time difference in review of 5 hours between two PRs, and where for one of the PRs the reviewer was explicitly requested for review. Otherwise, the flow is that for a PR /issue not made by a maintainer, a maintainer has to add tags & request relevant people for review.

I'm glad your disputing the unexplained choice, but there were two explanations in the prior PR https://github.com/arkworks-rs/algebra/pull/199#discussion_r570510180, https://github.com/arkworks-rs/algebra/pull/199#issuecomment-773718298 .  My understanding of your last comment is that you think there is an error in these explanations, I have not as of yet looked into the docs to understand this. But saying it was approved with unexplained changes is not true given there were two explanations.

jon-chuang:
> > Well that's definitely not what I was expected, what really irked me is that this PR had an immediate approving review, even though my overlapping PR was posted earlier, further, there were many unexplained choices like the ones I am disputing.
> > So it seemed to me like a bit of group-think and disregard for evaluating PRs on their merits, and also neglect for others' work in favour of those of the maintainers.
> 
> Can you assume some more positive intent on our part? We're talking about a time difference in review of 5 hours between two PRs, and where for one of the PRs the reviewer was explicitly requested for review. Otherwise, the flow is that for a PR /issue not made by a maintainer, a maintainer has to add tags & request relevant people for review.
> 
> Thank you for disputing the unexplained choice, but there were two explanations in the prior PR [#199 (comment)](https://github.com/arkworks-rs/algebra/pull/199#discussion_r570510180), [#199 (comment)](https://github.com/arkworks-rs/algebra/pull/199#issuecomment-773718298) . My understanding of your last comment is that you think there is an error in these explanations, I have not as of yet looked into the docs / read the latest edits of the last two comment to understand this. But saying it was approved with unexplained changes is not true given there were two explanations that both seemed sound to me.

I see, I guess I am wrong about the semantics, and I missed Pratyush's point as there was no explaining comment, and at the time a bigger overarching issue in discussion. The explanation in your reviewing comment is sound but easy to miss. 

So I would say that efforts at exposition, both in this PR and #204 should be improved by describing potentially contentious points of the PR. I think it is bad to rely on voluminous earlier discussion and courteous to provide a succint description.

I also stand behind my complaint that the approving reviewer did not do due process for the PR by asking the same clarifying questions.

jon-chuang:
@Pratyush ,while these changes only help to reduce a simple impl by one line, firstly, it's potentially confusing (requiring myself and dev to ask clarifying questions), and second, it doesn't appear to actually make any performance difference, hence I am not in favour of them.

If the self term were not to be used further, I believe either rustc or llvm will optimize the copy away in the current impl anyway, so there's no improvement from move semantics.

Pratyush:
> It's just there seems no reason why some people have more authority than others to push to branches, approve reviews, assign reviewers/request reviews, put labels. Makes little sense.

For the authority to push to branches etc, I think that's a standard practice where maintainers have more permissions, no? 

> it's potentially confusing (requiring myself and dev to ask clarifying questions), and second, it doesn't appear to actually make any performance difference, hence I am not in favour of them.

IMO the code is clearer and shorter, and also the compiler has less work to do in optimizing the code; move semantics are a core and inherent part of Rust, so I don't think it's unreasonable to expect future readers to know that. I can add some comments in line to clarify that.

As for this PR vs #204, I wanted to separate out the unrolling changes from the intrinsics changes, just to make the PRs and performance effects smaller. Does that seem reasonable?

jon-chuang:
Actually looking at the old impls, I'm more convinced. These changes are very natural.

How about we just combine all the changes here, and I can bench if there are performance regressions? I'm going to sleep now though. But I can submit the combined changes when I wake up. Or if you'd like, you can extract them from #204.

Pratyush:
I can combine the changes here (I'll cherry-pick your commits)

Pratyush:
Regarding the `reduce copies` commits, I'll keep them around for now, but it should be easy to drop those commits before merging the PR.

ff/src/biginteger/macros.rs
```diff
@@ -17,7 +17,16 @@ macro_rules! bigint_impl {
                 let mut carry = 0;
 
                 for i in 0..$num_limbs {
-                    self.0[i] = adc!(self.0[i], other.0[i], &mut carry);
+                    #[cfg(all(target_arch = "x86_64", feature = "asm"))]
```
Pratyush, line 20: Note that this is a weaker requirement than for the assembly backend; it works on stable, and on all `x86_64` processors, as long as the `asm` feature is enabled. You don't need to have the `bmi2` and `adx` target features enabled.

This means that even stable code can take advantage of faster additions, subtractions, and negations.

ff/src/biginteger/macros.rs
```diff
@@ -12,40 +12,71 @@ macro_rules! bigint_impl {
         impl BigInteger for $name {
             const NUM_LIMBS: usize = $num_limbs;
 
-            #[inline]
+            #[ark_ff_asm::unroll_for_loops]
```
ValarDragon, line 15: Can there be a before/after benchmark for field arithmetic on a 256, 384 and 768 bit fields?

ff/src/fields/models/cubic_extension.rs
```diff
@@ -372,8 +372,11 @@ impl<P: CubicExtParameters> FromBytes for CubicExtField<P> {
 impl<P: CubicExtParameters> Neg for CubicExtField<P> {
     type Output = Self;
     #[inline]
-    fn neg(self) -> Self {
-        Self::new(self.c0.neg(), self.c1.neg(), self.c2.neg())
+    fn neg(mut self) -> Self {
```
ValarDragon, line 375: I still don't understand why this change is useful, does it show up in benchmarks or is it aesthetic?

Pratyush: it modifies the elements in place, instead of making a new one. I don't think it shows up in benchmarks, at least not enough to affect things greatly

ValarDragon (commented):
This LGTM, holding off on an approval until all changes are benchmarked on the varying field sizes.

I don't share the reservations around `sub(mut self, ...` and the like. Per godbolt, it seems that its just what looks more like idiomatic rust, not a performance question. I don't see how the one line shorter variant isn't idiomatic, but I'm not particularly familiar with this.

ff/src/biginteger/macros.rs
Pratyush: Adding one now.

ff/src/fields/models/cubic_extension.rs
ValarDragon: Sorry, I didn't mean the `mut` bit, I mean changing the `self.c0.neg` to `self.c0 = - self.c0`. I commented the wrong line 😅

ValarDragon: Ah I understand this now though, neg is not in place,

Pratyush:
Benchmarks:

<details>
<summary>Details</summary>

```
name                          bn254_fq_master ns/iter  bn254_fq ns/iter  diff ns/iter   diff %  speedup 
 bn254::fq::add_assign         5                        4                           -1  -20.00%   x 1.25 
 bn254::fq::deser              24                       29                           5   20.83%   x 0.83 
 bn254::fq::deser_unchecked    25                       29                           4   16.00%   x 0.86 
 bn254::fq::double             4                        3                           -1  -25.00%   x 1.33 
 bn254::fq::from_repr          24                       24                           0    0.00%   x 1.00 
 bn254::fq::into_repr          4                        4                            0    0.00%   x 1.00 
 bn254::fq::inverse            5,703                    5,103                     -600  -10.52%   x 1.12 
 bn254::fq::mul_assign         20                       21                           1    5.00%   x 0.95 
 bn254::fq::negate             6                        3                           -3  -50.00%   x 2.00 
 bn254::fq::repr_add_nocarry   4                        4                            0    0.00%   x 1.00 
 bn254::fq::repr_div2          3                        2                           -1  -33.33%   x 1.50 
 bn254::fq::repr_mul2          2                        2                            0    0.00%   x 1.00 
 bn254::fq::repr_num_bits      3                        3                            0    0.00%   x 1.00 
 bn254::fq::repr_sub_noborrow  5                        2                           -3  -60.00%   x 2.50 
 bn254::fq::ser                19                       19                           0    0.00%   x 1.00 
 bn254::fq::ser_unchecked      19                       18                          -1   -5.26%   x 1.06 
 bn254::fq::sqrt               6,527                    6,438                      -89   -1.36%   x 1.01 
 bn254::fq::square             20                       21                           1    5.00%   x 0.95 
 bn254::fq::sub_assign         6                        8                            2   33.33%   x 0.75 

```

</details>

<details>
<summary>BLS12-381</summary>

```
 name                                    bls12_381_fq_master ns/iter  bls12_381_fq ns/iter  diff ns/iter   diff %  speedup 
 bls12_381::fq::add_assign               8                            5                               -3  -37.50%   x 1.60 
 bls12_381::fq::deser                    54                           69                              15   27.78%   x 0.78 
 bls12_381::fq::deser_unchecked          54                           69                              15   27.78%   x 0.78 
 bls12_381::fq::double                   6                            5                               -1  -16.67%   x 1.20 
 bls12_381::fq::from_repr                40                           40                               0    0.00%   x 1.00 
 bls12_381::fq::into_repr                30                           30                               0    0.00%   x 1.00 
 bls12_381::fq::inverse                  11,642                       9,313                       -2,329  -20.01%   x 1.25 
 bls12_381::fq::mul_assign               38                           37                              -1   -2.63%   x 1.03 
 bls12_381::fq::negate                   9                            5                               -4  -44.44%   x 1.80 
 bls12_381::fq::repr_add_nocarry         6                            5                               -1  -16.67%   x 1.20 
 bls12_381::fq::repr_div2                2                            2                                0    0.00%   x 1.00 
 bls12_381::fq::repr_mul2                2                            2                                0    0.00%   x 1.00 
 bls12_381::fq::repr_num_bits            3                            2                               -1  -33.33%   x 1.50 
 bls12_381::fq::repr_sub_noborrow        7                            3                               -4  -57.14%   x 2.33 
 bls12_381::fq::ser                      43                           43                               0    0.00%   x 1.00 
 bls12_381::fq::ser_unchecked            44                           43                              -1   -2.27%   x 1.02 
 bls12_381::fq::sqrt                     19,697                       19,106                        -591   -3.00%   x 1.03 
 bls12_381::fq::square                   37                           37                               0    0.00%   x 1.00 
 bls12_381::fq::sub_assign               9                            6                               -3  -33.33%   x 1.50
```

</details>

<details>
<summary>Details</summary>

```
 name                            bw6_761_fq_master ns/iter  bw6_761_fq ns/iter  diff ns/iter   diff %  speedup
 bw6_761::fq::add_assign         16                         10                            -6  -37.50%   x 1.60 
 bw6_761::fq::deser              213                        212                           -1   -0.47%   x 1.00 
 bw6_761::fq::deser_unchecked    213                        212                           -1   -0.47%   x 1.00 
 bw6_761::fq::double             12                         9                             -3  -25.00%   x 1.33 
 bw6_761::fq::from_repr          182                        184                            2    1.10%   x 0.99 
 bw6_761::fq::into_repr          100                        97                            -3   -3.00%   x 1.03 
 bw6_761::fq::inverse            35,976                     26,473                    -9,503  -26.41%   x 1.36 
 bw6_761::fq::mul_assign         179                        178                           -1   -0.56%   x 1.01 
 bw6_761::fq::negate             22                         10                           -12  -54.55%   x 2.20 
 bw6_761::fq::repr_add_nocarry   11                         8                             -3  -27.27%   x 1.38 
 bw6_761::fq::repr_div2          4                          2                             -2  -50.00%   x 2.00 
 bw6_761::fq::repr_mul2          2                          3                              1   50.00%   x 0.67 
 bw6_761::fq::repr_num_bits      3                          3                              0    0.00%   x 1.00 
 bw6_761::fq::repr_sub_noborrow  13                         3                            -10  -76.92%   x 4.33 
 bw6_761::fq::ser                127                        131                            4    3.15%   x 0.97 
 bw6_761::fq::ser_unchecked      127                        131                            4    3.15%   x 0.97 
 bw6_761::fq::sqrt               172,372                    175,702                    3,330    1.93%   x 0.98 
 bw6_761::fq::square             161                        160                           -1   -0.62%   x 1.01 
 bw6_761::fq::sub_assign         16                         13                            -3  -18.75%   x 1.23 

```

</details>

The only slow down is in `deser`, and I can't figure out why (I tried various tricks to speed it up). But it's not on the critical path for any code, so I'm okay with merging as is.

ValarDragon (approved):
This LGTM. Base field Inverse speeding up seems like a solid reason to assume this is not just noise.

Pratyush:
ok @jon-chuang it seems the last thing that we need to account for is the speed of doubling; can we just implement it as `self + self` there (i.e., no need to change `mul2`)? (also, at least on our benchmark machine, the speed of addition and doubling is the same right now)

ValarDragon:
This line wasnt converted from #205, potentially more werent as well: https://github.com/arkworks-rs/algebra/pull/204/files#diff-26dabf22b47ebe1097e18d25b10c8252331b1255b142bf0f29cdc9f2ae28a27fL29

Pratyush:
I added the assert; the primary thing left is to decide whether to switch doubling to use addition as well.

ff/src/biginteger/macros.rs
jon-chuang: I think you can change the addcarryx to addcarry with no perf penalty. Addcarryx I think technically needs bmi and adx too

jon-chuang: Realistically though most x64 processors have both, so I've always wondered whether checking for them is a bit pedantic... But maybe it's fine to be a bit on the safe side. It's just very verbose to compile the asm version, I noticed that for many libraries, the asm versions are the defaults...

jon-chuang: > I think you can change the addcarryx to addcarry with no perf penalty.

I'll double check this

jon-chuang:
> ok @jon-chuang it seems the last thing that we need to account for is the speed of doubling; can we just implement it as `self + self` there (i.e., no need to change `mul2`)? (also, at least on our benchmark machine, the speed of addition and doubling is the same right now). Basically, the following:
> 
> ```rust
> fn double_in_place(&mut self) {
> 	#[cfg(all(target_arch = "x86_64", feature = "asm"))]
> 	{ *self += &*self; }
> 	#[cfg(not(feature = "asm"))]
> 	{
> 		//existing code
> 	}
> }
> ```

Well, I know it's faster when using the intrinsics, potentially it's slower when not. 

Btw does that code compile? I had to do some unsafe magic to try to get the mut and non-mut reference to work together?

Well I decided not to trust the Fq benchmarks on this, when I looked at the g1 benchmarks there was about a 3-4% improvement overall.

ff/src/biginteger/macros.rs
```diff
@@ -12,40 +12,71 @@ macro_rules! bigint_impl {
         impl BigInteger for $name {
             const NUM_LIMBS: usize = $num_limbs;
 
-            #[inline]
```
jon-chuang, line 15: Why are we removing inline? I think this should be given an inline hint no?

ff/src/biginteger/macros.rs
```diff
@@ -12,40 +12,71 @@ macro_rules! bigint_impl {
...
+            #[ark_ff_asm::unroll_for_loops]
             fn sub_noborrow(&mut self, other: &Self) -> bool {
                 let mut borrow = 0;
 
-                for (a, b) in self.0.iter_mut().zip(other.0.iter()) {
-                    *a = sbb!(*a, *b, &mut borrow);
+                for i in 0..$num_limbs {
+                    #[cfg(all(target_arch = "x86_64", feature = "asm"))]
+                    #[cfg_attr(all(target_arch = "x86_64", feature = "asm"), allow(unsafe_code))]
```
jon-chuang, line 46: I think this is unnecessarily verbose since it only compiles if the cfg is true. So previous version is fine.

jon-chuang:
I've pushed some changes, which help the PR reach the previous baseline perf. and address some of the review comments I made.

jon-chuang:
Here are the final results from before and after this PR:
Screenshot from 2021-02-06 10-37-47

We achieve a 15-25% speedup across the board.

Pratyush:
Ok I think this is ready to merge!

PR #207, opened 2021-02-05
Specialization to save additions in Fp2 for negative non-residue

ValarDragon:
In particular, this adds add_and_mul_fp_by_nonresidue, sub_and_mul_fp_by_nonresidue
and utilizes this in fp2 multiplication and squaring, saving additions in each.

This is because when the non-residue is negative, there is an extra operation for negating it. However every usage is of the form `x + mul_by_non_residue(y)` or `x - mul_by_non_residue(y)`, so that final negation can be absorbed.

I've benchmarked this on bls12_377, and this is an 8% savings to square and a 4% savings to mul_assign.

```
 name                             master.txt ns/iter  fp2_optimization.txt ns/iter  diff ns/iter  diff %  speedup
 bls12_377::fq2::add_assign       15                       15                                   0   0.00%   x 1.00
 bls12_377::fq2::deser            155                      176                                 21  13.55%   x 0.88
 bls12_377::fq2::deser_unchecked  155                      176                                 21  13.55%   x 0.88
 bls12_377::fq2::double           13                       12                                  -1  -7.69%   x 1.08
 bls12_377::fq2::inverse          11,891                   11,650                            -241  -2.03%   x 1.02
 bls12_377::fq2::mul_assign       232                      223                                 -9  -3.88%   x 1.04
 bls12_377::fq2::negate           24                       24                                   0   0.00%   x 1.00
 bls12_377::fq2::ser              90                       91                                   1   1.11%   x 0.99
 bls12_377::fq2::ser_unchecked    90                       94                                   4   4.44%   x 0.96
 bls12_377::fq2::sqrt             149,598                  150,679                          1,081   0.72%   x 0.99
 bls12_377::fq2::square           222                      203                                -19  -8.56%   x 1.09
 bls12_377::fq2::sub_assign       16                       16                                   0   0.00%   x 1.00
```
(I believe the serialization and sqrt speed differences are noise, since they were untouched. However it is slightly suspicious to claim this, since deser took a significant overhead. But I don't see why else deser_unchecked would have slowed down as it doesn't touch any of the involved methods)

Benchmarks for g2 are currently running, thought I'd open the PR in case anyone wanted to start reviewing it sooner.

/cc @jon-chuang

---

Before we can merge this PR, please make sure that all the following items have been
checked off. If any of the checklist items are not applicable, please leave them but
write a little note why.

- [x] Targeted PR against correct branch (master)
- [x] Linked to Github issue with discussion and accepted design OR have an explanation in the PR that describes this work.
[...]

PR #212, opened 2021-02-06
Generalised version of improved fp2

jon-chuang:
Related to #207 

The main contribution of this PR is to get rid of multiple redundant functions, and to unify all mul by small const methods into default implementations.

Results vis a vis pre #207:
Screenshot from 2021-02-07 07-51-02

Often, we get speedups for free (for instance for bw6). I'm still investigating why this is.

mnt4_298's residue is G2 basefield non-residue is 17, so it's not worthwhile.

---

write a little note why.

- [ ] Wrote unit tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -84,6 +84,7 @@ The main features of this release are:
 - #190 (ark-ec) Add curve cycle trait and extended pairing cycle trait for all types of ec cycles.
 - #201 (ark-ec, ark-ff, ark-test-curves, ark-test-templates) Remove the dependency on `rand_xorshift`
 - #205 (ark-ec, ark-ff) Unroll loops and conditionally use intrinsics in `biginteger` arithmetic, and reduce copies in `ff` and `ec` arithmetic.
+- #207 (ark-ff) Improve performance of extension fields when the non-residue is negative. (Improves fq2, fq12, and g2 speed on bls12 and bn curves)
 
 ### Bug fixes
 - #36 (ark-ec) In Short-Weierstrass curves, include an infinity bit in `ToConstraintField`.
--- a/ff/src/fields/mod.rs
+++ b/ff/src/fields/mod.rs
@@ -339,6 +339,33 @@ pub trait PrimeField:
     type Params: FpParameters<BigInt = Self::BigInt>;
     type BigInt: BigInteger;
 
+    #[inline(always)]
+    fn mul_by_i8(mut self, i: i8) -> Self {
+        self.mul_assign_by_i8(i);
+        self
+    }
+
+    #[inline(always)]
+    #[ark_ff_asm::unroll_for_loops]
+    fn mul_assign_by_i8(&mut self, i: i8) {
+        if i < 0 {
+            *self = self.neg();
+        }
+        let i = i.abs();
+
+        let tmp = *self;
+
+        for j in 1..7 {
+            if i >= 1 << (7 - j) {
+                self.double_in_place();
+                // if the next bit is 1
+                if (i >> (6 - j)) % 2 == 1 {
+                    *self += tmp;
+                }
+            }
+        }
+    }
+
     /// Returns a prime field element from its underlying representation.
     fn from_repr(repr: Self::BigInt) -> Option<Self>;
 
--- a/ff/src/fields/models/fp2.rs
+++ b/ff/src/fields/models/fp2.rs
@@ -1,5 +1,5 @@
 use super::quadratic_extension::*;
-use crate::{fields::PrimeField, Field};
+use crate::fields::PrimeField;
 use core::marker::PhantomData;
 use num_traits::Zero;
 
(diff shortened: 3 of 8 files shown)


jon-chuang:
@ValarDragon @Pratyush

jon-chuang:
I think to really generalise this, we need to define some way to define multiplications with integers for fields. Or, we should have some way to indicate that a field element is really small.

jon-chuang:
Another way to achieve this is by defining a new type parameter.

Pratyush:
What about adding a `SmallValue` associated type to `Field`?

```rust
pub trait Field: Mul<<Self as Field>::SmallValue, Output = Self> {
	...
	type SmallValue: Neg + Mul<Self, Output = Self> + Into<Self>;
}
```

We can use it as follows:
```rust

// Impl for prime fields FpN:
impl Field for FpN {
	type Smallvalue = i8;
}

// Impl for quadratic extension fields:
impl Field for QuadExtField {
	type Smallvalue = Pair<Self::BaseField>;
}

pub trait QuadExtParameters {
	const NONRESIDUE: Pair<Self::BaseField>; // No need for separate NONRESIDUE_SMALL
	// No need for mul_by_nonresidue, because it's subsumed by the trait bounds above
}

pub struct Pair<F: Field> {
	// The coordinates can be either small or full-sized.
	c0: Either<F, F::SmallValue>, 
	c1: Either<F, F::SmallValue>,
}
```

(Similarly for cubic ext field)

jon-chuang:
Hmm, that was something that crossed my mind, but I wanted something that could be dropped in so that one wouldn't have to rewrite any formulas.

I just resorted in the end to defining a specific method, `mul_by_i8`, where the integer is meant to be constant so that llvm can optimise out the branches and loops.

jon-chuang:
I should probably write some tests for `mul_by_i8`.

Pratyush:
The benefit of the foregoing approach is primarily that it enables multiplication by arbitrary small field elements cheaply. For example, one could use it also to replace the `mul_by_a` methods in SWParameters, and similarly for TEParameters. 

It would also reduce some complexity in writing new curves and fields. For example, right now, when implementing a new extension field, one could forget to specialize the `mul_by_nonresidue` method, resulting in slowdowns that are difficult to trace. By going this way, there's no way to forget. (similarly for `mul_by_a`)

jon-chuang:
Hmm, but do we have the same idea of what mul_by_small_element entails? 

There are two possibilities essentially:
- the double and add method used in this PR
- an improvement over montgomery mul that nevertheless can multiply by small numbers which are too large for method 1.

In my mind, it makes sense to define these as two separate functionalities, and one can already create these functions generically in the PrimeField trait or Fp struct, for instance.

Pratyush:
> There are two possibilities essentially:
> * the double and add method used in this PR
> * an improvement over montgomery mul can multiply field elements by small numbers which are too large for method 1.

Why not have a threshold? For small enough numbers, we use custom addition chains, and then after that we use the specialized montgomery mul (which we can add at a later date)

jon-chuang:
Well, determining that threshold is probably not that straightforward. I would rather let the downstream user decide which method they'd like to deploy, or if they'd like to set a threshold.

Btw, regarding mul_by_a, I noticed that 
```
#[inline(always)]
pub fn sub_noborrow(x: &mut [u64; 6], other: &[u64; 6]) -> bool {
    let mut borrow = 0u8;

    unsafe {
        borrow = core::arch::x86_64::_subborrow_u64(borrow, x[0], other[0], &mut x[0]);
        borrow = core::arch::x86_64::_subborrow_u64(borrow, x[1], other[1], &mut x[1]);
        borrow = core::arch::x86_64::_subborrow_u64(borrow, x[2], other[2], &mut x[2]);
        borrow = core::arch::x86_64::_subborrow_u64(borrow, x[3], other[3], &mut x[3]);
        borrow = core::arch::x86_64::_subborrow_u64(borrow, x[4], other[4], &mut x[4]);
        borrow = core::arch::x86_64::_subborrow_u64(borrow, x[5], other[5], &mut x[5]);
    }

    borrow != 0
}

pub fn zero() -> [u64; 6] {[0u64; 6]}

pub fn f(x: &mut [u64; 6]) {
    sub_noborrow(x, &zero());
}
```
results in non-trivial code for f. This means that rust probably can't optimise the zero additions away, sadly.

Hence I will also submit a PR to rewrite some formulas in adding results of mul_by_a, when a is zero.

Pratyush:
> Well, determining that threshold is probably not that straightforward. I would rather let the downstream user decide which method they'd like to deploy, or if they'd like to set a threshold.

For the case of PrimeFields we can support that use case by also requiring `Mul<i64, Output = Self>`, where which uses the montgomery mul, while `Mul<i8, Output = Self>` uses addition chains, and maybe delegates to the i64 impl at some threshold.

jon-chuang:
You are indeed right that having a faster mod mul for u64, let's say, would already be of great value. However, it unfortunately does not fit within the Montgomery representation framework, for there even small values in biginteger space are large values in montgomery rep space.

Hence, I do not have any straightforward idea on how to take advantage of small values that are not suitable for the double and add chain via mul_by_i8.

I have confirmed some positive results for converting some mul_by_a to use mul_by_i8:
Screenshot from 2021-02-07 13-59-55

jon-chuang:
Turns out I made a mistake in implementation, by default, I used naive mul_assign with Self::Residue rather than mul_base_field_by_residue. Now, it should be back to before for tower fields.

Issue #123, opened 2020-12-09
Poly: Investigate Faster sparse polynomial evaluation methods


ValarDragon:
At the moment, the logic for sparse polynomial evaluation is:
```
let total = self.coeffs.iter()
            .map(|(i, c)| (*c * point.pow(&[*i as u64])))
            .sum();
```

This recomputes the power of point for each term, which is more costly than need be.

Sparse polynomials for the most part used in vanishing polynomials in our codebase. I suggest we write a fast path for sparse polynomial evaluation, when all terms are powers of 2. (This captures vanishing polynomials for unions of cosets, and vanishing polynomials for subspaces of binary fields)
 
Then for general sparse polynomial evaluation, I suggest we compare the current method against computing all of the squares, and then make the per-term work just computing the correct product of the squares. (As this removes any duplicate squarings computed)

_Originally posted by @ValarDragon in https://github.com/arkworks-rs/algebra/pull/117#issuecomment-741568583_

jon-chuang:
well, naively, one solution is just to precompute the powers of 2 of the point, and implement the right addition chain for the given exponent i.

I agree this current method is quite ugly.

PR #214, opened 2021-02-08
Sparse poly eval

jon-chuang:
## Description

closes: https://github.com/arkworks-rs/algebra/issues/123
Results:
2.5x improvement.

New
Screenshot from 2021-02-08 13-50-46

Old
Screenshot from 2021-02-08 13-52-53

---

write a little note why.

- [x] Wrote unit tests

### DIFF
--- a/ff/src/fields/mod.rs
+++ b/ff/src/fields/mod.rs
@@ -189,8 +189,10 @@ pub trait Field:
         res
     }
 
-    /// Exponentiates this element by a number represented with `u64` limbs,
-    /// least significant limb first.
+    /// Exponentiates a field element by a number represented with `u64` limbs,
+    /// via a table containing as many powers of 2 of the given element
+    /// as the ceil of log2 of the exp, starting from the 1st power,
+    /// i.e. pows_2 = [p, p^2, p^4, ..., p^(2^n)] where exp has <= n bits
     #[inline]
     fn pow_with_table<S: AsRef<[u64]>>(pows_2: &[Self], exp: S) -> Self {
         let mut res = Self::one();
--- a/ff/src/fields/mod.rs
+++ b/ff/src/fields/mod.rs
@@ -189,19 +189,22 @@ pub trait Field:
         res
     }
 
-    /// Exponentiates a field element by a number represented with `u64` limbs,
-    /// via a table containing as many powers of 2 of the given element
-    /// as the ceil of log2 of the exp, starting from the 1st power,
-    /// i.e. pows_2 = [p, p^2, p^4, ..., p^(2^n)] where exp has <= n bits
+    /// Exponentiates a field element `f` by a number represented with `u64` limbs,
+    /// using a precomputed table containing as many powers of 2 of `f`
+    /// as the 1 + the floor of log2 of the exponent `exp`, starting from the 1st power.
+    /// That is, `powers_of_2` should equal `&[p, p^2, p^4, ..., p^(2^n)]`
+    /// when `exp` has at most `n` bits.
+    ///
+    /// This returns `None` when a power is missing from the table.
     #[inline]
-    fn pow_with_table<S: AsRef<[u64]>>(pows_2: &[Self], exp: S) -> Self {
+    fn pow_with_table<S: AsRef<[u64]>>(powers_of_2: &[Self], exp: S) -> Option<Self> {
         let mut res = Self::one();
         for (pow, bit) in BitIteratorLE::without_trailing_zeros(exp).enumerate() {
             if bit {
-                res *= &pows_2[pow];
+                res *= powers_of_2.get(pow)?;
             }
         }
-        res
+        Some(res)
     }
 }
 
--- a/ff/src/fields/mod.rs
+++ b/ff/src/fields/mod.rs
@@ -191,7 +191,7 @@ pub trait Field:
 
     /// Exponentiates this element by a number represented with `u64` limbs,
     /// least significant limb first.
-    #[inline(always)]
+    #[inline]
     fn pow_with_table<S: AsRef<[u64]>>(&self, exp: S, pows_2: &[Self]) -> Self {
         let exp = exp.as_ref();
         let len = exp.len();
@@ -199,7 +199,9 @@ pub trait Field:
         for i in 0..len {
             let lz = exp[len - 1 - i].leading_zeros() as usize;
             max -= lz;
-            if lz == 64 { break; }
+            if lz == 64 {
+                break;
+            }
         }
 
         let mut res = Self::zero();
--- a/ff/src/fields/mod.rs
+++ b/ff/src/fields/mod.rs
@@ -188,6 +188,30 @@ pub trait Field:
         }
         res
     }
+
+    /// Exponentiates this element by a number represented with `u64` limbs,
+    /// least significant limb first.
+    #[inline(always)]
+    fn pow_with_table<S: AsRef<[u64]>>(&self, exp: S, pows_2: &[Self]) -> Self {
+        let exp = exp.as_ref();
+        let len = exp.len();
+        let mut max = len * 64;
+        for i in 0..len {
+            let lz = exp[len - 1 - i].leading_zeros() as usize;
+            max -= lz;
+            if lz == 64 { break; }
+        }
+
+        let mut res = Self::zero();
+        for i in 0..max {
+            let part = i / 64;
+            let bit = i - (64 * part);
+            if exp[part] & (1 << bit) > 0 {
+                res *= pows_2[i]
+            }
+        }
+        res
+    }
 }
 
 /// A trait that defines parameters for a field that can be used for FFTs.
--- a/ff/src/fields/mod.rs
+++ b/ff/src/fields/mod.rs
@@ -192,24 +192,11 @@ pub trait Field:
     /// Exponentiates this element by a number represented with `u64` limbs,
     /// least significant limb first.
     #[inline]
-    fn pow_with_table<S: AsRef<[u64]>>(&self, exp: S, pows_2: &[Self]) -> Self {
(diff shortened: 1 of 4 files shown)


poly/src/polynomial/univariate/sparse.rs
```diff
@@ -62,12 +62,26 @@ impl<F: Field> Polynomial<F> for SparsePolynomial<F> {
...
+            pows_2.push(point);
+        }
         // compute all coeff * point^{i} and then sum the results
         let total = self
             .coeffs
             .iter()
-            .map(|(i, c)| (*c * point.pow(&[*i as u64])))
-            .sum();
+            .map(|(i, c)| {
+                debug_assert_eq!(point.pow_with_table(&[*i as u64], &pows_2[..]), point.pow(&[*i as u64]));
```
Pratyush, line 82: Do we need the debug assert?

ff/src/fields/mod.rs
```diff
@@ -188,6 +188,32 @@ pub trait Field:
...
+        let exp = exp.as_ref();
+        let len = exp.len();
+        let mut max = len * 64;
+        for i in 0..len {
+            let lz = exp[len - 1 - i].leading_zeros() as usize;
+            max -= lz;
+            if lz == 64 {
+                break;
+            }
+        }
```
Pratyush, line 205: Can you add a comment describing what this is doing?

jon-chuang: I changed it to use the BitIteratorLE

poly/src/polynomial/univariate/sparse.rs
```diff
@@ -62,11 +62,30 @@ impl<F: Field> Polynomial<F> for SparsePolynomial<F> {
         if self.is_zero() {
             return F::zero();
         }
+
+        let total_bits = core::mem::size_of::<usize>() * 8;
+        let max_pow_2 = total_bits - self.degree().leading_zeros() as usize;
+
+        let mut pows_2 = Vec::<F>::with_capacity(max_pow_2);
```
Pratyush, line 69: This is computing `log2(total_bits)`, right? You can replace it with `let max_pow_2 = ark_std::log2(total_bits);`.

Pratyush:
Thanks for this! This LGTM once the comment is addressed

poly/src/polynomial/univariate/sparse.rs
jon-chuang: fair enough

jon-chuang: imo this should be a usize default function

jon-chuang: btw, I am uncomfortable with the current name, imo it ought to be named `log2_then_ceil`, or `log2_of_next_pow_of_2` or something.

jon-chuang:
Ok, it should be good to merge now

poly/src/polynomial/univariate/sparse.rs
```diff
@@ -62,11 +62,29 @@ impl<F: Field> Polynomial<F> for SparsePolynomial<F> {
         if self.is_zero() {
             return F::zero();
         }
+
+        // Add one for good luck/avoid corner case
```
Pratyush, line 66: Why the `+ 1`? What's the corner case here?

jon-chuang: The corner case is when degree is a power of two.

jon-chuang: It's not important to work it out since the efficiency difference is logarithmic, so it only barely matters even when nonzero terms are logarithmic in the degree.

jon-chuang: Ok, I've used the exact formula now but better written

Pratyush: So you want to calculate `log2_floor`? (As opposed to `log2_ceil`, which is what `ark_std::log2` implements)

jon-chuang: Well, no, it's actually log2_plus_one_floor

Pratyush: Er how so? This playground shows that it's just `log2_ceil`: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=90a1770db078df16538f7e1a304524d7

Pratyush: Oh nvm, I get it now.

Pratyush: Hmm then I'm confused; why not use `ark_std::log2`? Why do you need to round up 2^{k} to 2^{k + 1}?

poly-benches/benches/dense_uv_polynomial.rs
```diff
@@ -35,6 +40,21 @@ fn setup_bench<F: Field>(c: &mut Criterion, name: &str, bench_fn: fn(&mut Benche
     group.finish();
 }
 
+fn bench_sparse_poly_evaluate<F: Field>(b: &mut Bencher, non_zero_entries: &usize) {
+    const MAX_DEGREE: usize = 1 << 15;
+    // Per benchmark setup
+    let mut rng = &mut ark_std::test_rng();
+    let mut inner: Vec<(usize, F)> = Vec::with_capacity(*non_zero_entries);
+    (0..*non_zero_entries)
```
Pratyush, line 48:
```suggestion
fn bench_sparse_poly_evaluate<F: Field>(b: &mut Bencher, non_zero_entries: usize) {
    const MAX_DEGREE: usize = 1 << 15;
    // Per benchmark setup
    let mut rng = &mut ark_std::test_rng();
    let mut inner: Vec<(usize, F)> = Vec::with_capacity(non_zero_entries);
    (0..non_zero_entries)
```

poly/src/polynomial/univariate/sparse.rs
```diff
@@ -62,11 +62,28 @@ impl<F: Field> Polynomial<F> for SparsePolynomial<F> {
...
+
+        let max_pow_2 = 0usize.leading_zeros() - self.degree().leading_zeros();
+        let mut pows_2 = Vec::<F>::with_capacity(max_pow_2 as usize);
+
+        let mut p = *point;
+        pows_2.push(p);
+        for _ in 1..max_pow_2 {
+            p.square_in_place();
+            pows_2.push(p);
+        }
```
Pratyush, line 74: I think we can replace it with

```suggestion
		let num_powers = ark_std::log2(self.degree());
        let mut powers_of_2 = Vec::with_capacity(num_powers as usize);

        let mut p = *point;
        pows_2.push(p);
        for _ in 1..=num_powers {
            p.square_in_place();
            powers_of_2.push(p);
        }
```

poly/src/polynomial/univariate/sparse.rs
```diff
@@ -62,11 +62,28 @@ impl<F: Field> Polynomial<F> for SparsePolynomial<F> {
...
             .coeffs
             .iter()
-            .map(|(i, c)| (*c * point.pow(&[*i as u64])))
+            .map(|(i, c)| {
+                debug_assert_eq!(
+                    F::pow_with_table(&pows_2[..], &[*i as u64]),
+                    point.pow(&[*i as u64]),
+                    "pows not equal"
+                );
+                *c * F::pow_with_table(&pows_2[..], &[*i as u64])
```
Pratyush, line 85:
```suggestion
                    F::pow_with_table(&powers_of_2[..], &[*i as u64]),
                    point.pow(&[*i as u64]),
                    "pows not equal"
                );
                *c * F::pow_with_table(&powers_of_2[..], &[*i as u64])
```

poly/src/polynomial/univariate/sparse.rs
Pratyush: I see, it's because of the loop which goes up to `max_pows - 1`

poly/src/polynomial/univariate/sparse.rs
jon-chuang: I mean, that's the same as adding 1. and it's not the exact formula as I said...

jon-chuang: In principle, I am ok with this but I prefer my way as it's more transparent, whereas its not completely clear what log2 does unless you read the impl.

ff/src/fields/mod.rs
```diff
@@ -188,6 +188,21 @@ pub trait Field:
...
+    #[inline]
+    fn pow_with_table<S: AsRef<[u64]>>(pows_2: &[Self], exp: S) -> Self {
+        let mut res = Self::one();
+        for (pow, bit) in BitIteratorLE::without_trailing_zeros(exp).enumerate() {
+            if bit {
+                res *= &pows_2[pow];
+            }
+        }
+        res
+    }
```
Pratyush, line 205: Let's make it return an option, so that it return `None` when the powers are insufficient

```suggestion
    /// Exponentiates a field element `f` by a number represented with `u64` limbs,
    /// using a precomputed table containing as many 
    /// powers of 2 of `f`
    /// as the ceiling of log2 of the exponent `exp`, starting from the 1st power.
    /// That is, `powers_of_2` should equal `&[p, p^2, p^4, ..., p^(2^n)]` 
    /// when `exp` has at most `n` bits.
    ///
    /// This returns `None` when a power is missing from the
    /// table.
    #[inline]
    fn pow_with_table<S: AsRef<[u64]>>(powers_of_2: &[Self], exp: S) -> Option<Self> {
        let mut res = Self::one();
        for (pow, bit) in BitIteratorLE::without_trailing_zeros(exp).enumerate() {
            if bit {
                res *= &powers_of_2.get(&pow)?;
            }
        }
        Some(res)
```

Pratyush (approved):
Thanks, this looks good now!

PR #220, opened 2021-03-05
Fix a typo in `variable_base.rs` and proof-reading

weikengchen:
## Description

Michele (@mmaker) found the typo of `backwords`. I, therefore, do a quick pass of this file, adding periods that we can add.

I use `backward` instead of `backwards` because the M-W dictionary prefers the first spelling; both are correct though.

---

- [ ] Wrote unit tests _(do not apply)_
- [ ] Added a relevant changelog entry to the `Pending` section in `CHANGELOG.md` _(see below)_

This seems to be a very tiny change---so I did not add an entry in the changelog.


ValarDragon (approved):
Nice catch 😂

PR #221, opened 2021-03-07
Add CI job to test against `curves` repo

ashutoshvarma:
## Description

closes: #34

Add new CI job `test_against_curves` which does the following steps (as per issue description) 
- Clone `curves` repo
- Clone `algebra` repo
- Patch curves `Cargo.toml` to force the dependency tree to use the local version of the algebra crates.
- Run all tests in `curves`

### Additional Notes
Running all the tests for curves taken around 22 min which leads to an increase of CI runtime from 10 min to 25 min on average.
But if an additional 15min can prevent merging changes that break downstream (curves), I'd say it's well spent

---

write a little note why.

- [ ] Wrote unit tests 

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -159,3 +159,60 @@ jobs:
         with:
             command: build
             args: --workspace --exclude ark-algebra-test-templates --exclude ark-poly-benches --target thumbv6m-none-eabi
+
+  test_against_curves:
+    name: Test against curves
+    runs-on: ubuntu-latest
+    env:
+      RUSTFLAGS: -Dwarnings
+    strategy:
+      matrix:
+        rust:
+          - stable
+          - nightly
+    steps:
+      - name: Checkout curves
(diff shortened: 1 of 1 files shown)


jon-chuang:
Hi, so is the idea to make sure that new PRs don't break curves? But how about the ones that do?

ashutoshvarma:
> Hi, so is the idea to make sure that new PRs in algebra don't break curves? But how about the ones that do?
I guess the idea is that at least it would be caught? I think this makes sense given that it's a separate job.

Yeah, that's the idea, ones that break curves should ideally cause CI to fail.

> Although this makes some sense, do note that many other downstream components make use of generic APIs in algebra, so this doesn't rule out downstream breakage outside of curves.

The job in PR only concerns with detecting breakage in curves. For other components, we can try to add similar jobs if the resulting CI runtime is within the respectable range.

Pratyush:
This is great, thanks!

Pratyush (approved):

PR #228, opened 2021-03-09
Explicitly specify the associated type of the trait & Minimum version.

sunhuachuang:
## Description

Checked on stable versions.

If explicitly specify the associated type of the trait `CurveCycle` & `PairingFriendlyCycle`.
`<=1.45` is not ok, because [`const fn`](https://github.com/rust-lang/rust/issues/49146).
`1.46 ~ 1.50` is ok. `cargo test` is ok.

If not.
`1.49 +` is ok.

---

write a little note why.

- [ ] Wrote unit tests

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@ In addition, the [`curves`](https://github.com/arkworks-rs/curves) repository co
 
 ## Build guide
 
-The library compiles on the `stable` toolchain of the Rust compiler. To install the latest version of Rust, first install `rustup` by following the instructions [here](https://rustup.rs/), or via your platform's package manager. Once `rustup` is installed, install the Rust toolchain by invoking:
+The library compiles on the `stable (1.46+)` toolchain of the Rust compiler. To install the latest version of Rust, first install `rustup` by following the instructions [here](https://rustup.rs/), or via your platform's package manager. Once `rustup` is installed, install the Rust toolchain by invoking:
 ```bash
 rustup install stable
 ```
--- a/ec/src/lib.rs
+++ b/ec/src/lib.rs
@@ -328,6 +328,8 @@ where
 
 pub trait CurveCycle
 where
+    <Self::E1 as AffineCurve>::BaseField: PrimeField + SquareRootField,
+    <Self::E2 as AffineCurve>::BaseField: PrimeField + SquareRootField,
     <Self::E1 as AffineCurve>::Projective: MulAssign<<Self::E2 as AffineCurve>::BaseField>,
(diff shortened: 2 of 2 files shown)


weikengchen:
I need to verify...but one of the reasons why we change the representation of `CurveCycle` and so on is to reduce the needs for the downstream to include those checks (but I could be wrong, it might be unnecessary).

One quick question: the arkworks-rs has traditionally been rush to use newer versions of Rust (since some language features we have been waiting for being stable can simplify the code a lot). In your use case, is there a situation where you would need to use a previous version?

sunhuachuang:
I also agree to use the latest version to provide performance and optimized code. And the library is evolving rapidly.

But unfortunately, our production environment is still using version 1.48, looking for stability, not the latest, so the default compile environment in docker cannot upgrade.

Maybe another way is to put some tags on the library to mark version upgrade and incompatibility. Otherwise, the only way is to lock commit, but it's not elegant.

Pratyush:
Is it not possible to install the latest rust version in your docker environment via a new docker layer? Or even just installing rustup from source?

sunhuachuang:
Thinks, I am going to persuade everyone to upgrade to the latest version. And then it will be easily solved.

And if no others need this PR, I will close it.

weikengchen:
Keep us posted. I think this PR is still helpful---we can say what is the minimal version needed to compile.

ValarDragon:
FWIW, you can now pin to v0.2.0, which will at least work for `1.49+`. If you have thoughts on how up-to-date of a rust version is ok for you're environment, please leave them in #255 as well!

ValarDragon:
I think this PR is good to merge

weikengchen:
Just to collect a last round of feedback: @sunhuachuang do you still need this PR so that `algebra` works for 1.46? @ValarDragon what do you think about this, since future versions of arkworks-rs may bring the version dependency even higher?

Based on the discussion above, what is most needed in this PR is that we specify the minimal version to use.

sunhuachuang:
I'm OK, I have upgraded the company environment to the latest 1.51+.

Rust is developing very fast. Although some projects continue to use lower versions, if no one else encounters this problem, we can continue to upgrade with Rust, as long as it is compatible with the 6-month (2 versions).

ValarDragon:
Oh cool, if this change isn't needed anymore, then I think we should close this. @huitseeker added an MSRV check in the https://github.com/arkworks-rs/algebra/tree/compiler_1.51 branch, so I think we should close this for now then

weikengchen:
Got it. Then let me close this PR for now.

PR #230, opened 2021-03-13
Adding wNAF multiplication functionality

ggkitsas:
## Description

Adds wNAF multiplication functionality for `Group` elements. It doesn't change the way scalar multiplication is currently performed but rather provides an option to use wNAF.

Two functions are added:
* `wnaf_table`: generates a lookup table for a given `Group` element and `window` size
* `wnaf_mul`: performs the scalar multiplication using a lookup table and a scalar in NAF (can be obtained by `find_wnaf`)
---

write a little note why.

- [x] Wrote unit tests

### DIFF
--- a/ec/src/group/wnaf.rs
+++ b/ec/src/group/wnaf.rs
@@ -26,7 +26,8 @@ pub fn wnaf_mul<G: Group>(table: &[G], scalar_wnaf: &[i64]) -> G {
     result
 }
 
-pub fn wnaf_table<G: Group>(table: &mut Vec<G>, mut base: G, window: usize) {
+pub fn wnaf_table<G: Group>(mut base: G, window: usize) -> Vec<G> {
+    let mut table: Vec<G> = vec![];
     table.truncate(0);
     table.reserve(1 << (window - 1));
 
@@ -36,4 +37,5 @@ pub fn wnaf_table<G: Group>(table: &mut Vec<G>, mut base: G, window: usize) {
         table.push(base);
         base.add_assign(&dbl);
     }
+    table
 }
--- a/ec/src/group/wnaf.rs
+++ b/ec/src/group/wnaf.rs
@@ -1,0 +1,39 @@
+extern crate alloc;
+use alloc::vec::Vec;
+use crate::Group;
+
+pub fn wnaf_mul<G: Group>(table: &[G], scalar_wnaf: &[i64]) -> G {
+    let mut result = G::zero();
+
+    let mut found_one = false;
+
+    for n in scalar_wnaf.iter().rev() {
+        if found_one {
+            result = result.double();
+        }
+
+        if *n != 0 {
+            found_one = true;
+
+            if *n > 0 {
+                result += &table[(n / 2) as usize];
+            } else {
+                result -= &table[((-n) / 2) as usize];
+            }
+        }
+    }
+
+    result
+}
+
+pub fn wnaf_table<G: Group>(table: &mut Vec<G>, mut base: G, window: usize) {
+    table.truncate(0);
+    table.reserve(1 << (window - 1));
+
+    let dbl = base.double();
+
+    for _ in 0..(1 << (window - 1)) {
+        table.push(base);
+        base.add_assign(&dbl);
+    }
+}
--- a/ec/src/group/wnaf.rs
+++ b/ec/src/group/wnaf.rs
@@ -2,7 +2,7 @@ extern crate alloc;
 use crate::Group;
 use alloc::vec::Vec;
 
-// Computes scalar multiplication of a scalar_wnaf and a group element G. It uses the wNAF method.
+// Computes scalar multiplication of a preprocessed group element by a scalar in windowed non-adjacent form (wNAF).
 // `scalar_wnaf` is expected in its NAF form which can be obtained by using `find_wnaf` on the underlying BigInt.
 // `table` holds pre-computed multiples of the group element; it can be calculated using `wnaf_table`.
 pub fn wnaf_mul<G: Group>(table: &[G], scalar_wnaf: &[i64]) -> G {
--- a/ec/src/group/wnaf.rs
+++ b/ec/src/group/wnaf.rs
@@ -1,6 +1,6 @@
 extern crate alloc;
-use alloc::vec::Vec;
 use crate::Group;
+use alloc::vec::Vec;
 
 pub fn wnaf_mul<G: Group>(table: &[G], scalar_wnaf: &[i64]) -> G {
     let mut result = G::zero();
--- a/ec/src/group/wnaf.rs
+++ b/ec/src/group/wnaf.rs
@@ -2,6 +2,9 @@ extern crate alloc;
 use crate::Group;
 use alloc::vec::Vec;
 
+// Computes scalar multiplication of a scalar_wnaf and a group element G. It uses the wNAF method.
+// `scalar_wnaf` is expected in its NAF form which can be obtained by using `find_wnaf` on the underlying BigInt.
+// `table` holds pre-computed multiples of the group element; it can be calculated using `wnaf_table`.
 pub fn wnaf_mul<G: Group>(table: &[G], scalar_wnaf: &[i64]) -> G {
     let mut result = G::zero();
 
@@ -28,7 +31,6 @@ pub fn wnaf_mul<G: Group>(table: &[G], scalar_wnaf: &[i64]) -> G {
 
 pub fn wnaf_table<G: Group>(mut base: G, window: usize) -> Vec<G> {
     let mut table: Vec<G> = vec![];
-    table.truncate(0);
     table.reserve(1 << (window - 1));
 
     let dbl = base.double();
--- a/ec/src/wnaf.rs
+++ b/ec/src/wnaf.rs
@@ -11,6 +11,8 @@ pub struct WnafContext {
 impl WnafContext {
     /// Construct a new context for a window of size `window_size`.
     pub fn new(window_size: usize) -> Self {
+        assert!(window_size >= 2);
+        assert!(window_size < 64);
         Self { window_size }
     }
 
@@ -48,7 +50,7 @@ impl WnafContext {
         if 1 << (self.window_size - 1) > base_table.len() {
             return None;
         }
-        let scalar_wnaf = scalar.into_repr().find_wnaf(self.window_size);
+        let scalar_wnaf = scalar.into_repr().find_wnaf(self.window_size).unwrap();
 
         let mut result = G::zero();
 
--- a/ec/src/wnaf.rs
+++ b/ec/src/wnaf.rs
@@ -1,41 +1,75 @@
 use crate::ProjectiveCurve;
+use ark_ff::{BigInteger, PrimeField};
 use ark_std::vec::Vec;
 
-// Computes scalar multiplication of a preprocessed group element by a scalar in windowed non-adjacent form (wNAF).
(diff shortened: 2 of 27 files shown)


ec/src/group/wnaf.rs
```diff
@@ -0,0 +1,39 @@
...
+            } else {
+                result -= &table[((-n) / 2) as usize];
+            }
+        }
+    }
+
+    result
+}
+
+pub fn wnaf_table<G: Group>(table: &mut Vec<G>, mut base: G, window: usize) {
```
ValarDragon, line 29: Should we change this method to just output the table, and not take in a table as input. I'm not seeing any benefit from having it be a mutative argument?

ec/src/group/wnaf.rs
```diff
@@ -0,0 +1,39 @@
+extern crate alloc;
+use crate::Group;
+use alloc::vec::Vec;
+
+pub fn wnaf_mul<G: Group>(table: &[G], scalar_wnaf: &[i64]) -> G {
```
ValarDragon, line 5: Can this function have some comments specifying the algorithm being used for future readers. e.g. just writing out the expected form of the scalar wnaf, and the equation its computing.

test-templates/src/groups.rs
```diff
@@ -70,4 +74,14 @@ pub fn group_test<G: Group>(a: G, mut b: G) {
         a.mul(&(fr_rand1 * &fr_rand2)),
         "(a * r1) * r2 != a * (r1 * r2)"
     );
+
+    // Check that mul and wnaf_mul give the same results for several window values
+    for w in 2..=22 {
```
ValarDragon, line 79: Can we lower the max tested table size, I think building a 2^22 table is way too much computation for CI

ec/src/group/wnaf.rs
```diff
@@ -0,0 +1,39 @@
...
+    result
+}
+
+pub fn wnaf_table<G: Group>(table: &mut Vec<G>, mut base: G, window: usize) {
+    table.truncate(0);
+    table.reserve(1 << (window - 1));
+
+    let dbl = base.double();
+
+    for _ in 0..(1 << (window - 1)) {
```
ValarDragon, line 35: Once this is merged, lets make an issue to add parallelism to this

ValarDragon (changes requested):
Thanks for PR'ing this! LGTM, left some requests for comments & minor changes

ec/src/group/wnaf.rs
ggkitsas: Makes sense, modified :+1:

ValarDragon (approved):
LGTM!

Minor note, can a changelog entry be added to ### features.
perhaps:
`- #230 (ark-ec) Add `wnaf_mul` implementation for groups`

ec/src/group/wnaf.rs
jon-chuang: not sure you can do this. Base is mutated sequentially

jon-chuang:
I will be making some PRs from celo-org merge which include a wNAF mul combined with GLV. I think what would be nice would be to have the GLV-compatible curve use wNAF mul with a default window size specified in parameters. 

I think we ought to make wNAF (+GLV if applicable) the default mul method, if it is faster.

ec/src/group/wnaf.rs
ValarDragon: You can, its similar to computing all powers of a generator in parallel. You chunk which multiples are to be computed in each thread, and then just combine them together when building the final window.

jon-chuang: Fair enough, but typically, w =5 or 6 at most for wNAF

Pratyush:
@jon-chuang how does this PR interact with your upcoming wNAF PR?

jon-chuang:
@Pratyush it can be used as an impl in a different file, which I am for as it separates duties.

ec/src/group/wnaf.rs
jon-chuang: Furthermore, for your idea, where it applies, one is probably still better off doubling sequentially to create a table and then running the exponentiation in parallel on the back of that table

ec/src/group/wnaf.rs
```diff
@@ -0,0 +1,43 @@
+extern crate alloc;
+use crate::Group;
+use alloc::vec::Vec;
+
+// Computes scalar multiplication of a scalar_wnaf and a group element G. It uses the wNAF method.
```
Pratyush, line 5:
```suggestion
// Computes scalar multiplication of a preprocessed group element by a scalar in windowed non-adjacent form (wNAF).
```

Pratyush:
Thanks for the PR! This looks mostly good to me, barring the concern @jon-chuang raised about using `Group`; it's better to implement it directly for `ProjectiveCurve`, so that you can take advantage of things like mixed addition.

ec/src/wnaf.rs
```diff
@@ -0,0 +1,43 @@
...
+            }
+        }
+    }
+
+    result
+}
+
+pub fn wnaf_table<G: ProjectiveCurve>(mut base: G, window: usize) -> Vec<G> {
+    let mut table: Vec<G> = vec![];
+    table.reserve(1 << (window - 1));
```
Pratyush, line 34:
```suggestion
    let mut table = Vec::with_capacity(1 << (window - 1));
```

ec/src/wnaf.rs
```diff
@@ -0,0 +1,43 @@
...
+
+    let dbl = base.double();
+
+    for _ in 0..(1 << (window - 1)) {
+        table.push(base);
+        base.add_assign(&dbl);
```
Pratyush, line 40:
```suggestion
        base += &dbl;
```

ec/src/wnaf.rs
```diff
@@ -0,0 +1,43 @@
...
+use crate::ProjectiveCurve;
+use alloc::vec::Vec;
+
+// Computes scalar multiplication of a preprocessed group element by a scalar in windowed non-adjacent form (wNAF).
+// `scalar_wnaf` is expected in its NAF form which can be obtained by using `find_wnaf` on the underlying BigInt.
+// `table` holds pre-computed multiples of the group element; it can be calculated using `wnaf_table`.
+pub fn wnaf_mul<G: ProjectiveCurve>(table: &[G], scalar_wnaf: &[i64]) -> G {
+    let mut result = G::zero();
+
+    let mut found_one = false;
```
Pratyush, line 11:
```suggestion
    let mut found_non_zero = false;
```

ec/src/wnaf.rs
```diff
@@ -0,0 +1,43 @@
...
+
+    for n in scalar_wnaf.iter().rev() {
+        if found_one {
+            result.double_in_place();
+        }
+
+        if *n != 0 {
+            found_one = true;
```
Pratyush, line 19:
```suggestion
            found_non_zero = true;
```

ec/src/wnaf.rs
```diff
@@ -0,0 +1,43 @@
...
+
+            if *n > 0 {
+                result += &table[(n / 2) as usize];
```
Pratyush, line 22: Should we change the signature to return an `Option<G>`, and return `None` if the table does not contain enough elements?

ec/src/wnaf.rs
```diff
@@ -0,0 +1,43 @@
+extern crate alloc;
+use crate::ProjectiveCurve;
+use alloc::vec::Vec;
+
+// Computes scalar multiplication of a preprocessed group element by a scalar in windowed non-adjacent form (wNAF).
+// `scalar_wnaf` is expected in its NAF form which can be obtained by using `find_wnaf` on the underlying BigInt.
+// `table` holds pre-computed multiples of the group element; it can be calculated using `wnaf_table`.
+pub fn wnaf_mul<G: ProjectiveCurve>(table: &[G], scalar_wnaf: &[i64]) -> G {
```
Pratyush, line 8: Should we rename this to `wnaf_mul_with_table`, and add a function `wnaf_mul` that works roughly like:
```rust
pub fn wnaf_mul<G: ProjectiveCurve>(element: &G, scalar_wnaf: &[i64]) -> G {
	// calculate table size
	let window_size = something();
	let table = wnaf_table(*element, window_size);
	wnaf_mul_with_table(&table, scalar_wnaf)
}
```

ec/src/wnaf.rs
```diff
@@ -0,0 +1,42 @@
...
+    let mut result = G::zero();
+
+    let mut found_non_zero = false;
+
+    for n in scalar_wnaf.iter().rev() {
+        if found_one {
```
Pratyush, line 14:
```suggestion
        if found_non_zero {
```

ec/src/wnaf.rs
ggkitsas: I don't know any particular way of deciding the window size. Do you have anything specific in mind?

ec/src/wnaf.rs
```diff
@@ -0,0 +1,42 @@
+extern crate alloc;
+use crate::ProjectiveCurve;
+use alloc::vec::Vec;
```
Pratyush, line 3:
```suggestion
use crate::ProjectiveCurve;
use ark_std::vec::Vec;
```

ec/src/wnaf.rs
Pratyush: Hmm it seems that right now the `wnaf_mul` passes tests because the output of `BigInteger::find_wnaf` assume a window size of just 2, and so we never access something that's not in the table. However, if we change the table size to 1, the tests should fail because `wnaf_mul` accesses a multiple that's not in the table. 

@ggkitsas Do you mind giving me access to your branch so that I can push some fixes for this?

ggkitsas: Ahh I see, @Pratyush invitation sent

Pratyush:
@ggkitsas I pushed some updates that should now enable working with arbitrary window sizes. Please take a look and let me know if it makes sense.

ggkitsas:
@Pratyush LGTM, thanks!

ec/src/wnaf.rs
```diff
@@ -0,0 +1,77 @@
...
+    /// Computes scalar multiplication of a group element by `scalar`.
+    /// `base_table` holds precomputed multiples of the group element; it can be generated using `Self::table`.
+    /// `scalar` is an element of `G::ScalarField`.
+    ///
+    /// Returns `None` if the table is too small.
+    pub fn mul_with_table<G: ProjectiveCurve>(
+        &self,
+        base_table: &[G],
+        scalar: &G::ScalarField,
+    ) -> Option<G> {
```
Pratyush, line 49: @ValarDragon @jon-chuang should we normalize the table here?

Also, should we make `fixed_base_msm` use wNAF also?

jon-chuang: I think normalizing the table could be done only for fixedbaseMSM? 

Unless mixed addition is significantly faster than the ordinary projective add? Still, for fixed base MSM, one can potentially use the infrastructure created for the batch ops. 

Is it used for something important? I can't recall.

Pratyush: Yeah, I was proposing using mixed addition instead of projective add. It's not always much faster, for example it's not worth it for Edwards curves. Another thing that we can do is to make two different methods, one that uses a projective table, and one that use an affine table. (Maybe we can abstract over these?)

Pratyush: Let's revisit this later; there's no need to block this PR on this decision.

Pratyush:
This LGTM, so unless folks would like further changes, I'll merge this today.

ValarDragon (approved):
LGTM!

PR #235, opened 2021-03-17
Miscellaneous Changes: Serialization, additional helper methods, reorg

jon-chuang:
## Description

Miscelaneous changes associated with celo-org merge.
Ranging from harmless to API changes. 
closes: https://github.com/arkworks-rs/algebra/pull/127

---

write a little note why.

- [x] Wrote unit tests


jon-chuang:
To reviewers:
- Not sure if you'd like for me to refactor further. This is the least cohesive PRs of those to come since I've staged them to build atop one another. However, if you think it would be cleaner to extract the diff of one of the branches `jonch/celo-org-...` feel free to do so.

ec/src/models/short_weierstrass_jacobian.rs
```diff
@@ -560,6 +589,18 @@ impl<P: Parameters> ProjectiveCurve for GroupProjective<P> {
...
+    fn mul<S: AsRef<[u64]>>(mut self, other: S) -> Self {
+        let mut res = Self::zero();
+        for b in BitIteratorBE::without_leading_zeros(other.as_ref()) {
+            res.double_in_place();
+            if b {
+                res += self;
+            }
+        }
+        self = res;
+        self
```
Pratyush, line 602:
```suggestion
    fn mul<S: AsRef<[u64]>>(self, other: S) -> Self {
        let mut res = Self::zero();
        for b in BitIteratorBE::without_leading_zeros(other.as_ref()) {
            res.double_in_place();
            if b {
                res += &self;
            }
        }
        res
```

ff/src/biginteger/mod.rs
```diff
@@ -25,6 +25,7 @@ bigint_impl!(BigInteger128, 2);
 bigint_impl!(BigInteger256, 4);
 bigint_impl!(BigInteger320, 5);
 bigint_impl!(BigInteger384, 6);
+bigint_impl!(BigInteger512, 6);
```
Pratyush, line 28:
```suggestion
bigint_impl!(BigInteger512, 8);
```

ff/src/biginteger/mod.rs
```diff
@@ -137,4 +138,18 @@ pub trait BigInteger:
         *self = Self::read(reader)?;
         Ok(())
     }
+
+    /// Takes two slices of u64 representing big integers and returns a bigger
+    /// BigInteger of type Self representing their product. Preferably used
+    /// only for even NUM_LIMBS. We require the invariant that this.len() ==
+    /// other.len() == NUM_LIMBS / 2
+    fn mul_no_reduce(this: &[u64], other: &[u64]) -> Self;
```
Pratyush, line 146: Does this need to be a method on bigint? Can't it just be a free-standing function that is generic over `B: BigInteger`?

So something like `fn mul_no_reduce<B: BigInteger>(this: &[u64], other: &[u64]) -> B`?

ff/src/biginteger/mod.rs
```diff
@@ -137,4 +138,18 @@ pub trait BigInteger:
...
+
+    /// Similar to `mul_no_reduce` but accepts slices of len == NUM_LIMBS and
+    /// only returns lower half of the result
+    fn mul_no_reduce_lo(this: &[u64], other: &[u64]) -> Self;
```
Pratyush, line 150: Same for this

ff/src/biginteger/mod.rs
```diff
@@ -137,4 +138,18 @@ pub trait BigInteger:
...
+
+    /// Copies data from a slice to Self in a len agnostic way,
+    // based on whichever of the two is shorter.
+    fn from_slice(slice: &[u64]) -> Self;
```
Pratyush, line 154: Same for this.

ff/src/biginteger/mod.rs
jon-chuang: Hmm definitely this is possible. But this way, one doesn't need to import additional methods

ff/src/fields/mod.rs
```diff
@@ -418,6 +443,11 @@ pub trait PrimeField:
         Self::Params::T_MINUS_ONE_DIV_TWO
     }
 
+    /// Returns the trace minus one divided by two.
+    fn modulus() -> Self::BigInt {
+        Self::Params::MODULUS
+    }
```
Pratyush, line 449: This already exists as the `characteristic` method on `Field`.
```suggestion
```

ff/src/biginteger/mod.rs
jon-chuang: Is your concern that there would be too many methods?

test-curves/Cargo.toml
```diff
@@ -9,10 +9,15 @@ edition = "2018"
 publish = false
 
 [dependencies]
+paste = "0.1"
```
Pratyush, line 12: I think this is not needed in this PR?

test-curves/Cargo.toml
```diff
@@ -9,10 +9,15 @@ edition = "2018"
...
 ark-std = { git = "https://github.com/arkworks-rs/utils", default-features = false }
 ark-ff = { path = "../ff", default-features = false }
 ark-ec = { path = "../ec", default-features = false }
 
+
+lazy_static = { version = "1.4.0", optional = true }
+accel = { git = "https://github.com/jon-chuang/accel", package = "accel", optional = true }
+
```
Pratyush, line 20: Same for this

ff/src/biginteger/mod.rs
Pratyush: Yeah, I think I want to reduce the API surface of our existing traits, because implementing them is kind of a pain right now.

ec/Cargo.toml
```diff
@@ -19,8 +19,13 @@ ark-ff = { path = "../ff", default-features = false }
 derivative = { version = "2", features = ["use_core"] }
 num-traits = { version = "0.2", default-features = false }
 rayon = { version = "1", optional = true }
+serde_json = { version = "1.0.58", optional = true }
+paste = "0.1"
 zeroize = { version = "1", default-features = false, features = ["zeroize_derive"] }
 
+[dev-dependencies]
+rand_xorshift = "0.2"
```
Pratyush, line 27: We can use `ark_std::rand::XorShiftRng` directly, no need to import this.
```suggestion
```

ec/src/lib.rs
```diff
@@ -214,6 +214,8 @@ pub trait ProjectiveCurve:
         self = res;
         self
     }
+
+    fn get_x(&mut self) -> &mut Self::BaseField;
```
Pratyush, line 218: I don't think this is well defined for projective curve representations, as the same curve point can have different projective coordinates, which means that the X coordinate can differ. 

(Also no code in this PR uses it, but I guess your later PRs will?)

ff/src/fields/mod.rs
jon-chuang: I think in my use case I needed it to be in the bigint format. I think `&[u64]` is problematic because then it requires a copy... which is not terribly expensive but...

ec/Cargo.toml
jon-chuang: yeah I'll have to change this...

ff/src/fields/mod.rs
Pratyush: I see, since it's going to be a constant maybe it's not copied at all? Is this mainly meant to be syntactic sugar for `<F as PrimeField>::Params::MODULUS`?

ff/src/biginteger/mod.rs
Pratyush: Maybe we can change the parameters here? `fn mul_no_reduce_lo(&self, other: &Self) -> Self`? Does your later code rely heavily on slices?

Another approach would be to use something like `num-bigint` for those places, because that explicitly supports these kinds of operations, and I imagine that the bigint ops won't dominate the GLV ops?

ff/src/biginteger/mod.rs
Pratyush: How do you use this API in your later PRs? Maybe it's possible to come up with an API that fits slightly better into the rest of the methods, as the current design API works with slices and only for for even-limb BigInts.

ec/Cargo.toml
```diff
@@ -19,8 +19,13 @@ ark-ff = { path = "../ff", default-features = false }
 derivative = { version = "2", features = ["use_core"] }
 num-traits = { version = "0.2", default-features = false }
 rayon = { version = "1", optional = true }
+serde_json = { version = "1.0.58", optional = true }
+paste = "0.1"
```
Pratyush, line 23: Are these used anywhere?

ec/src/lib.rs
jon-chuang: Hmm yes, I can remove it for now, it will be used only for a generic script to generate parameters. 

The purpose is to mutate x, in the use case for the purpose of applying an endomorphism, for which the particular Z coordinate does not matter.

ff/src/biginteger/mod.rs
jon-chuang: This (impl pain) isn't really an issue for the BigInteger trait, however? 

The question is whether one sees potential additional use elsewhere.

But, as the impact is small, I can move it to a separate method now, under an empty struct `BigIntegerHelper`

ec/src/lib.rs
Pratyush: I see, maybe it would be better to add `xyz_endo_in_place`/`xyz_endo` methods in the appropriate location (in the ModelParameters trait, perhaps?)

ff/src/fields/mod.rs
jon-chuang: Hmm, I'm not entirely clear on whether or not its copied. 
> Is this mainly meant to be syntactic sugar for <F as PrimeField>::Params::MODULUS?
Yes, I don't think this particular method should be an issue given we also define 
```
/// Returns the modulus minus one divided by two.
    fn modulus_minus_one_div_two() -> Self::BigInt {
        Self::Params::MODULUS_MINUS_ONE_DIV_TWO
    }
```
etc

ec/src/lib.rs
jon-chuang: possibly, but the method already defined later on would apply multiplication by a constant endo param (`P::glv_endomorphism_in_place(&mut p.x);`). 

This more generic method allows you to do whatever you want with x. I don't think this is an issue in itself if you know what you're doing... If you're worried about clutter, well, unfortunately I can't think of a workaround.

Further, as far as the curves we have are concerned, `get_x` and `get_y` for the ProjectiveCurve trait seem fairly reasonable...

jon-chuang: But we can discuss it further in the PR for which it is required.

Pratyush:
For the `serialize` changes, we should update the `derive` macro to additionally incorporate `uncompressed_unchecked`.

Also, maybe in a follow-up PR, let's make it so that the `Serialize` and `Deserialize` traits both have a `(de)serialize_with_mode` method:

```rust
trait CanonicalSerialize {
...
	fn serialize_with_mode<W: Write>(&self, compression: ShouldCompress, writer: W) {
		match compression {
			ShouldCompress::Yes => {
				// do stuff
			},
			ShouldCompress::No => {
				// do other stuff
			}
		}
	}

	// all other methods invoke `serialize_with_mode` with appropriate modes.
}

trait CanonicalDeserialize {
	fn deserialize_with_mode<R: Read>(&self, compression: ShouldCompress, check: ShouldCheck reader: R) -> Result<Self> {
		match compression {
			ShouldCompress::Yes => {
				// do stuff
			},
			ShouldCompress::No => {
				// do other stuff
			}
		}
		if let ShouldCheck::Yes = check {
			// perform checks
		}
		Ok(result)
	}
	
	// all other methods invoke `deserialize_with_mode` with appropriate modes.
}
```

(I omitted the check from serialize, because AFAIK there are no checks performed during serialization? This means that we remove `serialize_unchecked` and `serialize_uncompressed_unchecked` too.)

ff/src/fields/mod.rs
Pratyush: Ok yeah, that makes sense.

ec/src/lib.rs
Pratyush: Ah I meant that instead one would do something like `P::glv_endomorphism_in_place(&mut p);`, which fully encapsulates the details of the endomorphism; the signature would be something like

```rust
trait SWModelParameters {
	/// Existing stuff
	
	fn glv_endomorphism_in_place(p: &mut GroupProjective<Self>) {
		// do whatever
	}
}
```

Pratyush: But yeah, we can discuss it in the GLV PR

ff/src/fields/mod.rs
```diff
@@ -67,6 +67,31 @@ macro_rules! field_new {
...
+    };
+    ($name:ident, $c0:expr, $c1:expr, $c2:expr $(,)?) => {
+        $name {
+            c0: $c0,
+            c1: $c1,
+            c2: $c2,
+            _parameters: core::marker::PhantomData,
+        }
+    };
+}
```
Pratyush, line 93: Where is this used? Also, how is it different from the existing `field_new`

Pratyush: Ah I see, you don't convert to montgomery form here. Is there a reason the original constants can't be in normal (not Montgomery) form?

ff/src/fields/mod.rs
```diff
@@ -418,6 +443,11 @@ pub trait PrimeField:
         Self::Params::T_MINUS_ONE_DIV_TWO
     }
 
+    /// Returns the trace minus one divided by two.
```
Pratyush, line 446:
```suggestion
    /// Returns the prime modulus.
```

Pratyush:
@jon-chuang is there anything I can do to help push this forward?

ec/src/lib.rs
ValarDragon: Just checking, the plan is to remove it from this PR right?

ec/src/models/mod.rs
```diff
@@ -8,53 +8,15 @@ pub mod mnt6;
 pub mod short_weierstrass_jacobian;
 pub mod twisted_edwards_extended;
 
+pub use {
```
ValarDragon, line 11: Glad this code was moved to the relevant locations!

ec/src/models/mod.rs
```diff
@@ -8,53 +8,15 @@ pub mod mnt6;
...
-}
-
-pub trait MontgomeryModelParameters: ModelParameters {
-    const COEFF_A: Self::BaseField;
-    const COEFF_B: Self::BaseField;
-
-    type TEModelParameters: TEModelParameters<BaseField = Self::BaseField>;
+    type ScalarField: PrimeField
+        + SquareRootField
+        + From<<Self::ScalarField as PrimeField>::BigInt>
```
ValarDragon, line 20: Do we need this From? Its a breaking change, so if so it will need to be noted in changelog

serialize/Cargo.toml
```diff
@@ -15,6 +15,7 @@ edition = "2018"
 [dependencies]
 ark-serialize-derive = { path = "../serialize-derive", optional = true }
 ark-std = { git = "https://github.com/arkworks-rs/utils", default-features = false }
+paste = "0.1"
```
ValarDragon, line 18: Is this needed?

Pratyush (commented):
We've implemented some of the serialization changes in #463. For the GLV related changes, we should probably flesh out a design before merging.

PR #241, opened 2021-03-23
Add tests for msm::VariableBase; cleanup tests for bls,mnt::g1.

mmaker:
* Change the name and function signature from group_test(a, b) to
   group_tests::<G>().
* Add multi-scalar multiplications tests into bn384_small_two_adicity
   and bls12_381.
* Remove un-necessary small functions and test everywithing into test_g1.
  (Maybe a macro in test-templates would allow us to remove the entire folder
   test-curves?)

 > No issue related
- [x] Wrote unit tests
> Actually, added some!
> not relevant
> not important enough I believe

### DIFF
--- a/test-curves/src/bn384_small_two_adicity/tests.rs
+++ b/test-curves/src/bn384_small_two_adicity/tests.rs
@@ -4,7 +4,7 @@ use ark_ff::{One, UniformRand, Zero};
 use ark_std::rand::Rng;
 
 use crate::bn384_small_two_adicity::{g1, Fq, FqParameters, Fr, G1Affine, G1Projective};
-use ark_algebra_test_templates::{curves::*, fields::*, groups::*};
+use ark_algebra_test_templates::{curves::*, fields::*, groups::*, msm::test_var_base_msm};
 
 pub(crate) const ITERATIONS: usize = 5;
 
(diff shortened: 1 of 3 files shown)


mmaker:
The failing test is on another package ( https://github.com/arkworks-rs/curves/blob/master/bn254/src/curves/tests.rs#L46 ). 
I'm a bit puzzled why the same (incomplete) test was included in two different packages, and I'm not sure what's the best way around this error. 
Please let me know how you'd prefer me to proceed!

test-curves/src/bn384_small_two_adicity/tests.rs
```diff
@@ -32,23 +32,17 @@ fn test_fq() {
...
-
 #[test]
 fn test_g1_generator() {
     let generator = G1Affine::prime_subgroup_generator();
     assert!(generator.is_on_curve());
     assert!(generator.is_in_correct_subgroup_assuming_on_curve());
 }
+
+#[test]
+fn test_g1() {
```
ValarDragon, line 43: Do all other curves follow this test format? (`test_g1`, not `test_g1_projective_curve` and `test_g1_projective_group`) If so, PR looks good to merge to me

Pratyush:
@mmaker The reason tests are failing in curves is because `test-templates` is used for tests in that repo. If you could make a PR there as well, that would be great!

mmaker:
Hey, unfortunately I don't have time to fix this properly. I'm therefore closing this pull request and opening an issue #256 that would be great for a beginner contributor! :) I also added some hints on the macro formulation, as we envisaged.

If you could flag #256 as _easy_ (e.g. with a label) it'd be more easy to spot!

PR #242, opened 2021-03-23
Simple cache alignment for serial FFT

ValarDragon:
## Description

This PR adds a simple cache alignment for sequential FFTs. See #177 for more context, but essentially in the FFT at the moment, the roots are accessed in a manner that is cache unfriendly, which causes lots of latency.

It takes the approach of ensuring in the sequential case that the roots are laid out in memory in the exact same form that they are used in, in every iteration.

This provides pretty significant speedups to the algorithm. (They are super-linear in polynomial size, as they have increasingly notable affects at the later round of the FFT).
Here is a list of speedup percentages just at varying sizes, as measured on our benchmark server.
* 2^15 - 3.8%
* 2^16 - 5.8%
* 2^17 - 9.6%
* 2^18 - 11.4%
* 2^19 - 12.8%
* 2^20 - 15.74%
* 2^21 - 19.7%

It is left as a TODO to implement this for the IFFT loop, and to implement this in the parallel case. It is likely that benefits at smaller instance sizes will be more profound on commodity hardware with smaller cache sizes than our benchmark server.

---

write a little note why.

- [x] Wrote unit tests - covered by existing tests

### DIFF
--- a/poly/src/domain/radix2/fft.rs
+++ b/poly/src/domain/radix2/fft.rs
@@ -143,7 +143,7 @@ impl<F: FftField> Radix2EvaluationDomain<F> {
         // It is left as a TODO to implement this for the parallel case
         #[allow(unused_mut)]
         let mut roots = self.roots_of_unity(root);
-
+        
         #[cfg(not(feature = "parallel"))]
         let (mut root_len, mut first_iteration) = (roots.len(), true);
 
@@ -157,17 +157,19 @@ impl<F: FftField> Radix2EvaluationDomain<F> {
             // Cache align the FFT roots in the sequential case.
             // In this case, we are aiming to make every root that is accessed one after another,
             // appear one after another in the list of roots.
-            // Roots are already cache aligned in the first iteration, so we don't need to do anything.
             #[cfg(not(feature = "parallel"))]
-            if !first_iteration {
-                // if the roots are cache aligned in iteration i, then in iteration i+1,
-                // cache alignment requires selecting every other root.
-                // (The even powers relative to the current iterations generator)
-                for i in 1..(root_len / 2) {
-                    roots[i] = roots[i * 2];
+            {
+                // Roots are already cache aligned in the first iteration, so we don't need to do anything.
+                if !first_iteration {
+                    // if the roots are cache aligned in iteration i, then in iteration i+1,
+                    // cache alignment requires selecting every other root.
+                    // (The even powers relative to the current iterations generator)
+                    for i in 1..(root_len / 2) {
+                        roots[i] = roots[i * 2];
+                    }
+                    root_len /= 2;
+                    first_iteration = false;
                 }
-                root_len /= 2;
-                first_iteration = false;
             }
 
             let butterfly_fn = |(chunk_index, (lo, hi)): (usize, (&mut T, &mut T))| {
@@ -178,6 +180,7 @@ impl<F: FftField> Radix2EvaluationDomain<F> {
 
                 #[cfg(feature = "parallel")]
                 let index = nchunks * chunk_index;
+                // Due to cache aligning, the index into roots is the chunk index
                 #[cfg(not(feature = "parallel"))]
                 let index = chunk_index;
 
--- a/poly/src/domain/radix2/fft.rs
+++ b/poly/src/domain/radix2/fft.rs
@@ -145,9 +145,7 @@ impl<F: FftField> Radix2EvaluationDomain<F> {
         let mut roots = self.roots_of_unity(root);
 
         #[cfg(not(feature = "parallel"))]
-        let mut root_len = roots.len();
-        #[cfg(not(feature = "parallel"))]
-        let mut first_iteration = true;
+        let (mut root_len, first_iteration) = (roots.len(), true);
 
         let mut gap = xi.len() / 2;
         while gap > 0 {
--- a/poly/src/domain/radix2/fft.rs
+++ b/poly/src/domain/radix2/fft.rs
@@ -137,21 +137,53 @@ impl<F: FftField> Radix2EvaluationDomain<F> {
             });
     }
 
+    #[allow(unused)]
     fn io_helper<T: DomainCoeff<F>>(&self, xi: &mut [T], root: F) {
-        let roots = self.roots_of_unity(root);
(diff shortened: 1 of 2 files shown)


ValarDragon:
Note, the PR doesn't include the update for the IFFT, since its not clear to me how one should go about cache aligning it. 

You start with only needing a small subset, and continue needing more every iteration. The main solution I can think of is either building the new roots as you go through iterations, or using 50% extra RAM and storing essentially two copies of the root list.

poly/src/domain/radix2/fft.rs
```diff
@@ -137,21 +137,53 @@ impl<F: FftField> Radix2EvaluationDomain<F> {
             });
     }
 
+    #[allow(unused)]
```
Pratyush, line 140: This is not unused, right?

poly/src/domain/radix2/fft.rs
```diff
@@ -137,21 +137,53 @@ impl<F: FftField> Radix2EvaluationDomain<F> {
...
+            {
+                // Roots are already cache aligned in the first iteration, so we don't need to do anything.
+                if !first_iteration {
+                    // if the roots are cache aligned in iteration i, then in iteration i+1,
+                    // cache alignment requires selecting every other root.
+                    // (The even powers relative to the current iterations generator)
+                    for i in 1..(root_len / 2) {
+                        roots[i] = roots[i * 2];
+                    }
+                    root_len /= 2;
```
Pratyush, line 168: Should we also shrink `roots` via `truncate` + `shrink_to_fit`? This would cause re-allocations but can also reduce memory consumption. We don't have to do this at every iteration, maybe every other iteration?

poly/src/domain/radix2/fft.rs
ValarDragon: root_len is unused in the parallel cae

poly/src/domain/radix2/fft.rs
ValarDragon: I don't see any benefit to doing this. We're in the sequential case, so any re-allocation hurts us here, and the advantage would be that we optimistically freed up some memory for processes running in parallel, of which there would be none

poly/src/domain/radix2/fft.rs
ValarDragon: and roots needs `allow(unused_mut)` in the parallel case as well

Pratyush: Maybe let's wrap it in a `#[cfg(not(feature = "parallel"))]`?

poly/src/domain/radix2/fft.rs
```diff
@@ -137,21 +137,53 @@ impl<F: FftField> Radix2EvaluationDomain<F> {
...
     }
 
+    #[allow(unused)]
     fn io_helper<T: DomainCoeff<F>>(&self, xi: &mut [T], root: F) {
-        let roots = self.roots_of_unity(root);
+        // In the sequential case, we will keep on making the roots cache-aligned,
+        // according to the access pattern that the FFT uses.
+        // It is left as a TODO to implement this for the parallel case
+        let mut roots = self.roots_of_unity(root);
+        let mut root_len = roots.len();
```
Pratyush, line 146:
```suggestion
		#[cfg(not(feature = "parallel"))]
        let mut root_len = roots.len();
```

poly/src/domain/radix2/fft.rs
```diff
@@ -137,36 +137,37 @@ impl<F: FftField> Radix2EvaluationDomain<F> {
...
             let chunk_size = 2 * gap;
+            #[cfg(feature = "parallel")]
             let nchunks = xi.len() / chunk_size;
 
             // Cache align the FFT roots in the sequential case.
             // In this case, we are aiming to make every root that is accessed one after another,
             // appear one after another in the list of roots.
+            // Roots are already cache aligned in the first iteration, so we don't need to do anything.
             #[cfg(not(feature = "parallel"))]
-            {
```
ValarDragon: Is it fine if I switch this back to having the braces, for scoping the 
`             // Roots are already cache aligned in the first iteration, so we don't need to do anything.`
comment?

poly/src/domain/radix2/fft.rs
```diff
@@ -179,9 +180,10 @@ impl<F: FftField> Radix2EvaluationDomain<F> {
                 {
                     *hi *= roots[nchunks * chunk_index];
                 }
+
+                // Cache-aligned, so the index is just the chunk_index
```
ValarDragon, line 184: Mind if I move this back into the inner scope?

ValarDragon:
I confirmed that our stylistic changes above didn't affect performance

Pratyush (approved):

jon-chuang:
Another way to improve cache locality might be to unroll the butterfly loops by constant factor (2 or 4 etc.) which can help by encouraging prefetching of cache lines. Not sure if this applies since group elements generally already take up a single cache line. But it could help prefetching, perhaps.

jon-chuang:
I think that for IFFT, since we are using the roots multiple times, it still does make sense to spend the initial data copy cost.

So one just allocates the array/vector with capacity and then writes to it just like is being done for `io_helper`. Btw, what's with the funny name? Actually, what's the out of order ordering...? How is it possible to perform an FFT if it is out of order?

weikengchen:
Just a quick question, why in the case of non-parallel, it is using:
```
let index = chunk_index;
```

But in the original code, it seems that it would use `nchunks * chunk_index`? Note that in both cases, it is done over different chunks:
`ark_std::cfg_chunks_mut!(xi, chunk_size).for_each(|cxi| {`

Pratyush:
@weikengchen It's because in the serial case we re-arrange the roots so that the relevant roots are at `chunk_index`, instead of `nchunks * chunk_index`.

jon-chuang:
@weikengchen I think it's cos in the parallel case we don't do the cache alignment step.

Btw @ValarDragon don't you think we should cache align for the parallel case too, since it helps when the gap is small? Further, I think I could also post a PR where either the chunk parallelisation itself is chunked (instead of par_iter) when the gap is large.

In addition, we also do loop unrolling

weikengchen:
Ok. Got it now!

Issue #219, opened 2021-02-13
compilation error


brucechin:
Hello guys,

I use `rustc 1.47.0` stable version, and when I tried to compile the algebra library I met the following error:

```
error[E0308]: mismatched types
   --> ff/src/fields/models/quadratic_extension.rs:272:24
    |
272 |             self.c0 *= v0;
    |                        ^^
    |                        |
    |                        expected reference, found associated type
    |                        help: consider borrowing here: `&v0`
    |
    = note:    expected reference `&<P as fields::models::quadratic_extension::QuadExtParameters>::BaseField`
            found associated type `<P as fields::models::quadratic_extension::QuadExtParameters>::BaseField`
    = help: consider constraining the associated type `<P as fields::models::quadratic_extension::QuadExtParameters>::BaseField` to `&<P as fields::models::quadratic_extension::QuadExtParameters>::BaseField`
    = note: for more information, visit https://doc.rust-lang.org/book/ch19-03-advanced-traits.html

error[E0308]: mismatched types
   --> ff/src/fields/models/quadratic_extension.rs:275:24
    |
275 |             self.c1 *= c0_copy;
    |                        ^^^^^^^
    |                        |
    |                        expected reference, found associated type
    |                        help: consider borrowing here: `&c0_copy`
    |
    = note:    expected reference `&<P as fields::models::quadratic_extension::QuadExtParameters>::BaseField`
            found associated type `<P as fields::models::quadratic_extension::QuadExtParameters>::BaseField`
    = help: consider constraining the associated type `<P as fields::models::quadratic_extension::QuadExtParameters>::BaseField` to `&<P as fields::models::quadratic_extension::QuadExtParameters>::BaseField`
    = note: for more information, visit https://doc.rust-lang.org/book/ch19-03-advanced-traits.html

error: aborting due to 2 previous errors
```

Has anyone met this error before? Thanks!!

weikengchen:
I compiled locally with rust 1.49.0, and it seems to pass.

Two things to try:
1. first, do `cargo update` in case there are some libraries out of sync, and then see if `cargo build` works.
2. upgrade to rust 1.49.0, the library has used several features that are recently stable in newer versions.

weikengchen:
And I tried compiling in rust 1.47.0 and encountered the same problem---you might have to switch to 1.49.0

brucechin:
Thanks for your reply! i will update my rust version then

brucechin:
[...]

PR #215, opened 2021-02-08
cargo check failed when the rust version less than 1.49.0, add rust-t…

mengsuenyan:
cargo check failed when the rust version less than 1.49.0. may be can add rust-toolchain to specify the rust version;

when using the rust 1.48.0, the following error occurs:

```txt
    Checking zeroize v1.2.0
error[E0308]: mismatched types
   --> /home/lzj/.cargo/git/checkouts/algebra-7e23afa68841b66e/a3a1ab2/ff/src/fields/models/quadratic_extension.rs:272:24
    |
272 |             self.c0 *= v0;
    |                        ^^
    |                        |
    |                        expected reference, found associated type
    |                        help: consider borrowing here: `&v0`
    |
    = note:    expected reference `&<P as quadratic_extension::QuadExtParameters>::BaseField`
            found associated type `<P as quadratic_extension::QuadExtParameters>::BaseField`
    = help: consider constraining the associated type `<P as quadratic_extension::QuadExtParameters>::BaseField` to `&<P as quadratic_extension::QuadExtParameters>::BaseField`
    = note: for more information, visit https://doc.rust-lang.org/book/ch19-03-advanced-traits.html

error[E0308]: mismatched types
   --> /home/lzj/.cargo/git/checkouts/algebra-7e23afa68841b66e/a3a1ab2/ff/src/fields/models/quadratic_extension.rs:275:24
    |
275 |             self.c1 *= c0_copy;
    |                        ^^^^^^^
    |                        |
    |                        expected reference, found associated type
    |                        help: consider borrowing here: `&c0_copy`
    |
    = note:    expected reference `&<P as quadratic_extension::QuadExtParameters>::BaseField`
            found associated type `<P as quadratic_extension::QuadExtParameters>::BaseField`
    = help: consider constraining the associated type `<P as quadratic_extension::QuadExtParameters>::BaseField` to `&<P as quadratic_extension::QuadExtParameters>::BaseField`
    = note: for more information, visit https://doc.rust-lang.org/book/ch19-03-advanced-traits.html

error: aborting due to 2 previous errors

For more information about this error, try `rustc --explain E0308`.
error: could not compile `ark-ff`

```

PR #255, opened 2021-03-28
Tiny-const-generics

huitseeker:
This is an easy initial fragment of #186 aimed at breaking off one easy part out of that big PR, and get us started on the path to `min_const_generics`. Props to @jon-chuang for the actual work.

(#186 needs a bit of bespoke specialization to achieve performance parity, so this just aims for code simplification) 

bench on master:
https://gist.github.com/413013073c2c81e8999e77c1cb37c5b8
bench on this PR:
https://gist.github.com/1f1cc7a70d02709889b9a0dfa180f2cd

- [ ] Wrote unit tests

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@ In addition, the [`curves`](https://github.com/arkworks-rs/curves) repository co
 
 ## Build guide
 
-The library compiles on the `stable` toolchain of the Rust compiler. To install the latest version of Rust, first install `rustup` by following the instructions [here](https://rustup.rs/), or via your platform's package manager. Once `rustup` is installed, install the Rust toolchain by invoking:
+The library compiles on the `stable` toolchain of the Rust compiler (v 1.51+). To install the latest version of Rust, first install `rustup` by following the instructions [here](https://rustup.rs/), or via your platform's package manager. Once `rustup` is installed, install the Rust toolchain by invoking:
 ```bash
 rustup install stable
 ```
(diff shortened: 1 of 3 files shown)


daira (approved):
Untested ACK

ChenxingLi:
Have you planned to replace the macro `bigint_impl!` with the const-generics trait?

Pratyush:
@ValarDragon are we okay with bumping the next release to require a minimum supported rust version of 1.51?

ValarDragon:
Isn't rust 1.51 only a week old? imo, its overly prohibitive for a library to be backwards incompatible with versions that aren't even a month old. cref #215 #219.

Perhaps we should only depend on 1.51 after 6 weeks or so, when 1.52 comes out? And until then maintain a 1.51 branch?

ChenxingLi:
> @ChenxingLi The work in that direction is ongoing in #186, I believe.

Thanks! That's cool.

ValarDragon:
I made a compiler_1.51 branch, lets switch this to point at that branch, and merge into there? (And then plan to merge that branch into master once the rust compiler hits 1.53)?

weikengchen:
Dev, you mentioned 1.53 in the last reply, is it 1.51? https://blog.rust-lang.org/2021/03/25/Rust-1.51.0.html Or we just want to wait until 1.53 so it is more commonly used?

ValarDragon:
The 1.51 branch was meant to indicate targeting that MSRV, and then we'd want to hold off on merging that into master until 1.53.

I think this is good to merge now

weikengchen:
Sure, we can do it soon. Do you think the mentioning of 1.51+ in README.md and the version check necessary then?

ValarDragon:
I think so, it will make the MSRV clearly documented, and if a compilation error occurs from someone using an insufficient version, it will provide a clear error message.

PR #258, opened 2021-04-03
Add scalar multiplication impl for DensePolynomial

ryanleh:
## Description

Removes the need for performing FFTs when multiplying polynomials by a field element.

---

Before we can merge this PR, please make sure that all the following items have been
checked off. If any of the checklist items are not applicable, please leave them but
write a little note why.

- [x] Targeted PR against correct branch (master)
- [x] Linked to Github issue with discussion and accepted design OR have an explanation in the PR that describes this work.
- [x] Wrote unit tests
- [ ] Updated relevant documentation in the code
- [x] Added a relevant changelog entry to the `Pending` section in `CHANGELOG.md`
- [x] Re-reviewed `Files changed` in the Github PR explorer

PR #259, opened 2021-04-04
Add sparse scalar mul and sub

ryanleh:
Same thing as https://github.com/arkworks-rs/algebra/pull/258 for `SparsePolynomial<F>`. Also adds impls for adding and subtracting `SparsePolynomial`s from `DensePolynomial`s. 

---

write a little note why.

- [x] Wrote unit tests

### DIFF
--- a/poly/src/polynomial/univariate/dense.rs
+++ b/poly/src/polynomial/univariate/dense.rs
@@ -346,6 +346,39 @@ impl<'a, 'b, F: Field> AddAssign<(F, &'a DensePolynomial<F>)> for DensePolynomia
     }
 }
 
+
+impl<'a, F: Field> AddAssign<&'a SparsePolynomial<F>> for DensePolynomial<F> {
+    #[inline]
+    fn add_assign(&mut self, other: &'a SparsePolynomial<F>) {
+        if self.is_zero() {
+            self.coeffs.truncate(0);
+            self.coeffs.resize(other.degree() + 1, F::zero());
+
+            for (i, coeff) in other.iter() {
+                self.coeffs[*i] = *coeff;
+            }
+            return;
+        } else if other.is_zero() {
+            return;
+        } else {
+            // If `other` has higher degree than `self`, create a dense vector
+            // storing the upper coefficients of the addition
+            let mut upper_coeffs = match other.degree() > self.degree() {
+                true => vec![F::zero(); other.degree() - self.degree()],
+                false => Vec::new(),
+            };
+            for (pow, coeff) in other.iter() {
+                if *pow <= self.degree() {
+                    self.coeffs[*pow] += coeff;
+                } else {
+                    upper_coeffs[*pow - self.degree() - 1] = *coeff;
+                }
+            }
+            self.coeffs.extend(upper_coeffs);
+        }
+    }
+}
+
 impl<F: Field> Neg for DensePolynomial<F> {
     type Output = DensePolynomial<F>;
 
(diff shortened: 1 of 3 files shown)


weikengchen (approved):
Looks good.

poly/src/polynomial/univariate/dense.rs
```diff
@@ -262,6 +261,37 @@ impl<'a, 'b, F: Field> Add<&'a DensePolynomial<F>> for &'b DensePolynomial<F> {
...
+                } else {
+                    upper_coeffs[*pow - result.degree() - 1] = *coeff;
+                }
+            }
+            result.coeffs.extend(upper_coeffs);
+            result
+        };
+        result
+    }
+}
```
Pratyush, line 293: Probably in this case we should add `AddAssign`/`SubAssign` impls too.

poly/src/polynomial/univariate/sparse.rs
```diff
@@ -44,6 +47,12 @@ impl<F: Field> core::ops::Deref for SparsePolynomial<F> {
     }
 }
 
+impl<F: Field> core::ops::DerefMut for SparsePolynomial<F> {
+    fn deref_mut(&mut self) -> &mut [(usize, F)] {
+        &mut self.coeffs
+    }
+}
```
Pratyush, line 54: Why do we need this? Isn't this an implementation detail?

weikengchen: I looked at the code. I think this is to be consistent with DensePolynomial, which already has the two handy implementations like this one:

```
impl<F: Field> Deref for DensePolynomial<F> {
    type Target = [F];

    fn deref(&self) -> &[F] {
        &self.coeffs
    }
}

impl<F: Field> DerefMut for DensePolynomial<F> {
    fn deref_mut(&mut self) -> &mut [F] {
        &mut self.coeffs
    }
}
```

It does simplify some coding.

poly/src/polynomial/univariate/dense.rs
weikengchen: added, also with tests.

poly/src/polynomial/univariate/sparse.rs
weikengchen: This might be a useful thing to keep.

weikengchen:
I do a pass and it is ready for a follow-up review.

poly/src/polynomial/univariate/sparse.rs
Pratyush: The impl for `DensePolynomial` makes sense, as the natural implementation of a polynomial would be as a list of coefficients. IMO for sparse polynomials it's less clear, because you can have different "natural" implementations: a list of (index, coeff) pairs like right now, or a Hashmap<usize, F>, or maybe another implementation.

weikengchen: Got it. Then I think the best way is to get rid of `DerefMut` for `SparsePolynomial` for now.

weikengchen: Actually, this is a problem---we have to expose an interface for the data inside `SparsePolynomial`, otherwise `DensePolynomial` would be unable to add a sparse polynomial. 

A possibility is `Iterator` instead of `Deref` or `DeferMut`, where  `SparsePolynomial` should be able to allow one to iterate over all its key-value pairs.

Pratyush: Maybe we can just add an `Index` impl which returns the `i`-th coefficient? And a corresponding `fn get(&self, index: usize) -> Option<&F>` method which returns the coefficient? This is analogous to the methods on slices.

weikengchen: Would that be slightly slow? Since one now has to iterate all the possible indices (up to the degree).

Pratyush: I see what you mean. Hm let's just keep this as is.

Pratyush (approved):

PR #260, opened 2021-04-11
Only check CI against some curves

Pratyush:
I picked the following curves as they probably most of the field and curve instantiations we care about
- `ark-bls12-377`
- `ark-bls12-381`
- `ark-bn254`
- `ark-pallas`
- `ark-bw6-761`
- `ark-mnt4-298`
- `ark-mnt6-298`
- `ark-ed-on-bls12-377`

---

write a little note why.

- [ ] Wrote unit tests

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -215,4 +215,12 @@ jobs:
             command: test
             args: "--workspace \
                    --all-features \
+                   -p ark-bls12-377 \
+                   -p ark-bls12-381 \
+                   -p ark-bn254 \
+                   -p ark-pallas \
+                   -p ark-bw6-761 \
+                   -p ark-mnt4-298 \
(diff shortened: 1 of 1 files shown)


weikengchen (approved):
I looked up the other curves available in `curves` and I think this subsetting makes sense.

ValarDragon (approved):
Seems good to me

Pratyush:
Hm as is this didn't save much, as most of the time in the curve tests is spent in compiling, and that seems to not be affected much by the fact that we're only compiling a few crates. Tomorrow I'll attempt to use the approach we use in the curves crate here.

weikengchen:
One possibility is that we add some of these curves into the `ark-test-curves` repo here. This may reduce the compiling time as it avoids `curves` repo.

Pratyush:
Ok this approach works now!

PR #261, opened 2021-04-16
Add support for 448-bit integers and fields

daira:
## Description

Add support for 448-bit integers and fields. These will be needed for the Pluto and Eris curves
(https://github.com/daira/pluto-eris).

The changes are trivial and amount to applying the existing macros to a length of 448 bits (7 64-bit words).
This interacts with #186, but it should be easy to resolve any conflict.

---

write a little note why.

- [x] ~~Linked to Github issue with discussion and accepted design~~ OR have an explanation in the PR that describes this work.
- [x] Wrote unit tests
- [ ] Updated relevant documentation in the code [not needed]

### DIFF
--- a/ff/src/biginteger/mod.rs
+++ b/ff/src/biginteger/mod.rs
@@ -25,6 +25,7 @@ bigint_impl!(BigInteger128, 2);
 bigint_impl!(BigInteger256, 4);
 bigint_impl!(BigInteger320, 5);
 bigint_impl!(BigInteger384, 6);
+bigint_impl!(BigInteger448, 7);
 bigint_impl!(BigInteger768, 12);
 bigint_impl!(BigInteger832, 13);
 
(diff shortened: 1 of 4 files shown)


Pratyush (approved):
Thanks for the PR!

daira:
Changing this to draft while I check that it actually works for Pluto/Eris (there seems to be some minor problem with `SquareRootField`, so perhaps I missed something).

daira:
I didn't miss anything. When testing with the Pluto/Eris implementation I just needed to patch `ark-ff` and `ark-serialize` to use the 0.2.0 releases plus this patch only. I.e. in `Cargo.toml`:
```
[patch.crates-io]
ark-ff = { path = '../algebra/ff' }
ark-serialize = { path = '../algebra/serialize' }
```
And then it works. So, I think this PR is fine. (I rebased it relative to the 0.2.0 release for more straightforward testing.)

ValarDragon (approved):
LGTM!

Issue #226, opened 2021-03-09
Remove `Ord`, `PartialOrd` bounds from `Field` and `ProjectiveCurve`/`AffineCurve` traits


Pratyush:
There is no universal ordering on finite field or group elements. This is exemplified right now by the fact that our `Fp` elements impose a lexicographic ordering by enforcing representatives to be in the range `0..(p-1)`. However, an equally natural representation would be `-(p-1)/2..(p_1)/2`, and the ordering on this representation is different from the first one. This is especially problematic for extension fields: is (-1, 1) > (0, 0)?

What should we do here?

burdges:
There exist many handy trait impls that do not exactly make sense or do not provide exactly the assurances one desires.  In this case, I'd just give doc comments on the `impl`s saying ``/// Arbitrary ordering implemented for `BTreeMap` users`` and then continuing with exactly what you wrote here. 

I think `HashMap: Clone` provides about the best example:  We clearly want `HashMap: Clone` sometimes, but it avoids rehashing by cloning the randomness, which risks DoS attacks.  There are two possible solutions, either add a `HashMap::clone_fast` method and make `HashMap: Clone` create new randomness, or else add a clippy lint that suggests `iter().collect::<..>()` whenever using `HashMap: Clone`.  I'd vote for the clippy lint against  `HashMap: Clone` since so many users do not require DoS protection.

Pratyush:
That makes sense; I'll add that documentation now.

PR #263, opened 2021-04-16
Adds `From<iXXX>`  implementations to fields

huitseeker:
Those delegate to the respective, existing `From<uXXX>` implementations for` XXX` in `[8, 16, 32, 64, 128]`

### DIFF
--- a/ff/Cargo.toml
+++ b/ff/Cargo.toml
@@ -20,6 +20,7 @@ ark-std = { version = "0.2.0", default-features = false }
 ark-serialize = { version = "^0.2.0", path = "../serialize", default-features = false }
 derivative = { version = "2", features = ["use_core"] }
 num-traits = { version = "0.2", default-features = false }
+paste = "1.0"
 rayon = { version = "1", optional = true }
 zeroize = { version = "1", default-features = false, features = ["zeroize_derive"] }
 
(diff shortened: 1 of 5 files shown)


ValarDragon:
Should we add these `From<ixxx>` trait bounds to the field trait?

huitseeker:
@ValarDragon Good question. It breaks compatibility of the libraries out there which may implement an `ark-ff::Field` on their own custom structure. At any rate, and so that this trait bound is easier to do in the future, I've added a commit doing the harder part of deriving the `impl`as part of the prime field macros.

ValarDragon (approved):
This looks good to me!
I've never seen / used paste before, but from looking at the docs this seems right to me.

cref #226, but I think its fine to assume that the ordering `(-p/2)...(p/2 - 1)` would be implied for signed ints here

huitseeker:
@ValarDragon FYI, you can inspect the output with [cargo expand](https://github.com/dtolnay/cargo-expand)
Here's the output of :
`cargo expand 2>/dev/null | grep -A7 -Ee 'impl<P:.*> From<[iu][0-9]+> for' |head -n 50`
https://gist.github.com/8f66c293bc5b231932dca3544460feec

Pratyush (changes requested):

Pratyush (approved):

Issue #86, opened 2020-02-28
Hashing Structures helpers


gakonst:
Depends on arkworks-rs/snark#104 

It'd be nice to have a helper method as an extension trait over `CanonicalSerialize` which calls a cryptographically secure hash function (maybe make it take a [`Digest`](https://docs.rs/digest/0.8.1/digest/trait.Digest.html)?) on the serialized input, and returns the fixed size hash of the provided data. 

It'd be even better if it would also be available on structs composing such elements.

Example:
```rust
use blake2::Blake2b;
use generic_array::GenericArray;
use zexe::algebra::*;
use digest::Digest;

trait Hasher: CanonicalSerialize {
   fn hash<H: Digest>(&self) -> GenericArray<u8, H::OutputSize> {
       let buffer = self.serialize();
       let mut hasher = H::new();
       hasher.input(buffer);
       hasher.result()
   }
}

struct X<E: PairingEngine> {
    a: E::G1Affine,
    b: E::G2Affine,
    c: Vec<E::G1Affine>,
}

x = X::<E> {
  a: E::G1Affine::zero(),
  b: E::G2Affine::zero(),
  c: vec![],
}
let hash = x.hash::<Blake2b>();
```

This would serialize all elements (compressed or uncompressed? i don't think it matters) to a buffer, and then take the hash of the buffer.

It becomes a bit trickier I believe if the struct ends up having non-CanonicalSerialize elements.

e.g.
```rust
struct X<E: PairingEngine> {
    a: E::G1Affine,
    b: E::G2Affine,
    c: Vec<E::G1Affine>,
    transcript_hash: [u8; 64],
} 
```

PR #265, opened 2021-04-21
Implements hashing as a blanket trait for instances of `CanonicalSerialize`

huitseeker:
- this saves an alllocation w.r.t the issue-suggested approach by implementing `io::Write` on the input instance of `digest::Digest`,
- note that most instances of `digest::Digest` [already](https://gist.github.com/huitseeker/e827161413063e347ce5a496b66ff287) have an [`io::Write` instance](https://github.com/rustcrypto/hashes#hashing-readable-objects), but `CanonicalSerialize` consuming its `io::Write` argument prevents its usage,
- this hence implements `io::Write` on a [cheap newtype wrapper](https://rust-unofficial.github.io/patterns/patterns/behavioural/newtype.html)

closes: #86 

---

- [x] Wrote unit tests

### DIFF
--- a/serialize/src/lib.rs
+++ b/serialize/src/lib.rs
@@ -115,17 +115,25 @@ impl<'a, H: Digest> ark_std::io::Write for HashMarshaller<'a, H> {
 
 /// The CanonicalSerialize induces a natural way to hash the
 /// corresponding value, of which this is the convenience trait.
-pub trait Hash: CanonicalSerialize {
+pub trait CanonicalSerializeHashExt: CanonicalSerialize {
     fn hash<H: Digest>(&self) -> GenericArray<u8, <H as Digest>::OutputSize> {
         let mut hasher = H::new();
         self.serialize(HashMarshaller(&mut hasher))
             .expect("HashMarshaller::flush should be infaillible!");
         hasher.finalize()
     }
+
+    fn hash_uncompressed<H: Digest>(&self) -> GenericArray<u8, <H as Digest>::OutputSize> {
+        let mut hasher = H::new();
+        self.serialize_uncompressed(HashMarshaller(&mut hasher))
+            .expect("HashMarshaller::flush should be infaillible!");
+        hasher.finalize()
+    }
 }
 
-/// Hash is a (blanket) extension trait of CanonicalSerialize
-impl<T: CanonicalSerialize> Hash for T {}
+/// CanonicalSerializeHashExt is a (blanket) extension trait of
+/// CanonicalSerialize
+impl<T: CanonicalSerialize> CanonicalSerializeHashExt for T {}
 
 /// Deserializer in little endian format allowing flags to be encoded.
 pub trait CanonicalDeserializeWithFlags: Sized {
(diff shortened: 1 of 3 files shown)


serialize/src/lib.rs
```diff
@@ -94,6 +96,37 @@ pub trait CanonicalSerialize {
...
+    fn hash<H: Digest>(&self) -> GenericArray<u8, <H as Digest>::OutputSize> {
+        let mut hasher = H::new();
+        self.serialize(HashMarshaller(&mut hasher))
+            .expect("HashMarshaller::flush should be infaillible!");
+        hasher.finalize()
+    }
+}
+
+/// Hash is a (blanket) extension trait of CanonicalSerialize
+impl<T: CanonicalSerialize> Hash for T {}
```
Pratyush, line 128: Should we just make this a standalone method(s)?
```rust
fn hash_with_digest_compressed<D: Digest, T: CanonicalSerialize>() -> GenericArray<u8, <H as Digest>::OutputSize> { ... }

fn hash_with_digest_uncompressed<D: Digest, T: CanonicalSerialize>() -> GenericArray<u8, <H as Digest>::OutputSize> { ... }
```

Not sure of the trade-off between importing a trait (`Hash`) vs. two methods. What do you think?

Regardless, I think we should rename it from `Hash` to something else, as it collides with `core::hash::Hash`.

huitseeker: > Not sure of the trade-off between importing a trait (Hash) vs. two methods.

The [extension trait pattern](https://rust-lang.github.io/rfcs/0445-extension-trait-conventions.html) is clear enough that most will prefer it to manually importing several methods, I think.

I renamed the trait according to best practices.

huitseeker: I could also simply add those two methods to the `CanonicalSerialize` trait? Wdyt?

serialize/src/lib.rs
```diff
@@ -94,6 +96,44 @@ pub trait CanonicalSerialize {
...
+
+    #[inline]
+    fn flush(&mut self) -> ark_std::io::Result<()> {
+        Ok(())
+    }
+}
+
+/// The CanonicalSerialize induces a natural way to hash the
+/// corresponding value, of which this is the convenience trait.
+pub trait CanonicalSerializeExt: CanonicalSerialize {
```
Pratyush, line 118: Maybe we call this something like `CanonicalSerializeThenHash`?

serialize/src/lib.rs
Pratyush: I prefer the separation of concerns, and maybe we should put this behind a separate feature?

Pratyush:
> but CanonicalSerialize consuming its io::Write argument prevents its usage,

Is it not possible to work around this by doing `&mut writer` (`&mut W` is `Write` whenever `W` is)

serialize/src/lib.rs
huitseeker: - The serialization is never materialized anywhere, whereas `CanonicalSerializeThenHash` implies that it is, in the way the peudo-code in #86 did.
- The best practice for an extension trait is really to finish in `Ext`. `CanonicalSerializeHashExt`?

huitseeker:
@Pratyush Sure, though this signature change would change the API of `CanonicalSerialize` in a backwards-incompatible way. 

Moreover, I'm not sure we want to depend on the `std::io::Write` implementations in RustCrypto/hashes: 
- the `Digest` trait does not guarantee its existence on its own, 
- it's implemented from unrelated (MAC) trait impls through a macro [which future in the code base isn't assured](https://github.com/RustCrypto/traits/pull/217/files)
- the no-std context adds an additional wrinkle.

serialize/src/lib.rs
Pratyush: `CanonicalSerializeHashExt` sounds good.

Pratyush:
Ok that makes sense, let's proceed as is.

Pratyush (approved):

Pratyush:
Oh @huitseeker if you could add a CHANGELOG entry that would be great (also for #263)

ValarDragon (commented):
Thanks for adding this!

Issue #274, opened 2021-05-11
Support pairing with negative `u` for BN curves


weikengchen:
Recently in one of my projects, I need to implement a BN254 curve with very efficient pairing, but two-arity in Fr is not needed. The best known BN254 curve for this situation is:
 
> a BN curve over the 254-bit prime p = 36u^4 + 36u^3 + 24u^2 + 6u + 1 where u = -(2^62 + 2^55 + 1) < 0

with optimal ate pairing, as documented in https://github.com/herumi/ate-pairing, where the `u` is negative.

This curve differs from the BN254 curve in the arkworks/curves repo mainly because it has a lower NAF weight of |6u+2| and therefore permits a faster pairing. Yet, it does not have a good two-arity in Fr, so not very suitable for FFT.

### Problems

Although our BN curve framework leaves space for negative `u` (here, `X`), as follows. The support is not full.

```
pub trait BnParameters: 'static {
    const X: &'static [u64];
    const X_IS_NEGATIVE: bool;
}
```

Basically, the preparing algorithm for G2 (https://github.com/arkworks-rs/algebra/blob/master/ec/src/models/bn/g2.rs#L68) as well as the Miller loop implementation is a version for `u > 0` (https://eprint.iacr.org/2010/186.pdf) rather than a generalized version that works with `u < 0` (as in https://eprint.iacr.org/2010/526.pdf). This issue is found by a comparison between the prepared G2 generator between `herumi/ate-pairing` and arkworks's. 

There are two solutions:
(1) prepare the G2 elements differently when u < 0
(2) or, do the Miller loop differently when u < 0

A PR may be coming soon to add this support. Though BN254 curves are less likely to be used in cryptocurrency systems since it does not offer 128-bit security, larger BN curves may still be useful---BN curves have G_1 cofactor to be one, a property that many other curves do not have that some applications may want. As time goes, though, we may all go to BLS-24 or BLS-48 curves.

daira:
Pluto (https://github.com/daira/pluto-eris and https://github.com/arkworks-rs/curves/pull/54) is also a BN curve with negative u.

weikengchen:
When fixing, I found that all the places to change co-occur what `ATE_LOOP_COUNT_IS_NEGATIVE` is flagging.

So, turns out that the generators can be matched if one simply sets:

```
const ATE_LOOP_COUNT_IS_NEGATIVE: bool = true;
```

[...]

PR #275, opened 2021-05-12
Improve documentation for BN curve parameter `ATE_LOOP_COUNT`

weikengchen:
## Description

This PR is to clarify that `ATE_LOOP_COUNT` is `6 X + 2` instead of `|6 X + 2|`, a possible confusion.

This is related to #274 and arkworks-rs/curves#54.

---

write a little note why.

N/A:
- [ ] Wrote unit tests

### DIFF
--- a/ec/src/models/bn/mod.rs
+++ b/ec/src/models/bn/mod.rs
@@ -23,7 +23,7 @@ pub trait BnParameters: 'static {
     // Whether or not `X` is negative.
     const X_IS_NEGATIVE: bool;
 
-    // The absolute value of `6X + 2`
+    // The absolute value of `6X + 2`.
     const ATE_LOOP_COUNT: &'static [i8];
     // Whether or not `6X + 2` is negative. `6X + 2` is negative often when `X` is negative.
     const ATE_LOOP_COUNT_IS_NEGATIVE: bool;
--- a/ec/src/models/bn/mod.rs
+++ b/ec/src/models/bn/mod.rs
@@ -18,10 +18,16 @@ pub enum TwistType {
 }
 
 pub trait BnParameters: 'static {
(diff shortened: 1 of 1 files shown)


ValarDragon (approved):

ec/src/models/bn/mod.rs
```diff
@@ -18,10 +18,16 @@ pub enum TwistType {
...
 
 pub trait BnParameters: 'static {
+    // The absolute value of the BN curve parameter `X` (as in `q = 36 X^4 + 36 X^3 + 24 X^2 + 6 X + 1`).
     const X: &'static [u64];
+    // Whether or not `X` is negative.
     const X_IS_NEGATIVE: bool;
+
+    // The absolute value of `6X + 2`.
     const ATE_LOOP_COUNT: &'static [i8];
+    // Whether or not `6X + 2` is negative. `6X + 2` is negative often when `X` is negative.
```
daira, line 28: Isn't 6X + 2 negative *exactly* when X is negative? (Because X is an integer and 2 < 6.) That implies that this field should probably just be removed and `X_IS_NEGATIVE` used instead.

daira: Also I'm confused because this doc appears to contradict the following sentence in the PR description:
> This PR is to clarify that ATE_LOOP_COUNT is 6 X + 2 instead of |6 X + 2|, a possible confusion.

weikengchen: You are right. 

Let me clarify this PR's documentation as well as the documentation in the code (you are right, 2 < 6 so X must be negative). 

I think there are two possible future plans:

1. Not too breaking, which can be done now:
- Use `X_IS_NEGATIVE ` to decide whatever `ATE_LOOP_COUNT_IS_NEGATIVE` is deciding, since they should be equivalent for BN curves.

2. Breaking:
- Remove `ATE_LOOP_COUNT_IS_NEGATIVE` and update the BN254 implementation accordingly.

Issue #271, opened 2021-04-29
Easier conversion for scalars from a smaller sized field to a larger scalar field


alxiong:
It should be a common task to add smaller scalar (e.g. Jubjub scalar) to a circuit over a larger scalar field (e.g. BLS), and right now there's no clear API from the library that support such conversion.

The only hacky way I find is through their commonly shared `BigInteger` trait, thus one may do the following as of `v0.2.0`:

```rust
use ark_bls12_381::Fr as BlsScalar;
use ark_ed_on_bls12_381::Fr as JubjubScalar;
use ark_ff::{BigInteger, PrimeField};

let s = JubjubScalar::rand(&mut rng);
let s_in_bls = BlsScalar::from_le_bytes_mod_order(&s.into_repr().to_bytes_le()); // THIS LOOK UGLY
```

Hopefully there's some `impl<P: FrParameters> TryFrom<FpXXX<P>> for FpXXX<P>` which compares the `MODULUS` and if dest scalar is larger, then does the conversion?
Or maybe there're some better ways to do it?

burdges:
Why would this be common?  Just ECDSA?  See also https://github.com/arkworks-rs/nonnative

weikengchen:
I think Alex is thinking about a version of nonnative field, but not in the R1CS, rather, in the Rust ecosystem.

The hacky way that you pointed out is also the only way I know: convert it into bytes and take it back.

alxiong:
> I think Alex is thinking about a version of nonnative field, but not in the R1CS,

yes, that's right.

> Why would this be common?

curve operation inside circuit is an example. I think it's a common practice to use a curve (Jubjub) over a base field that's the same size as another scalar fields (Bls) to reduce the total number of constraints required for operations like scalar multiplication, point addition etc. 

> The hacky way that you pointed out is also the only way I know

I'm not 100% sure that hacky way is correct tho, it definitely compiles, but I don't know whether the `xxx_mod_order()` part would skew the value.

weikengchen:
Let me summary the issue, and let me suggest one change to the repo...

I worked with Alex offline. I think we find one thing:

If one does this to a prime field element s,

`println!("the value of a field element: {}", s)`, this would output the correct value (via the Display trait implementation).

`println!("the value of a field element: {:?}", s)`, this would output the Montgomery representation (via the Debug trait implementation).

It might be confusing. Should we make them the same? 

[...]

PR #280, opened 2021-05-14
Implement `Into<BigUint>` and `From<BigUint>` for `BigInteger` and `PrimeField`

weikengchen:
## Description

This PR partially handles the challenges of converting values between different fields, by allowing BigInteger of different sizes to convert between each other, through an intermediate representation `BigUint`. This can also make it easy to print a field element.

More support for field conversions is possible: We could implement the same thing for `PrimeField`, which likely is a simple wrapper of this. 

Note that it is also possible to enable `From<T: To<Uint>> for U: From<Uint>` so this can make the conversion simplest. I feel it is too risky since the enforcement of which field can be applied to which object can prevent certain misuse, and this more aggressive form may sacrifice such enforcement, especially conversions between different fields are, though useful, not occurring in the majority of the code.

closes: #271 

---

write a little note why.

- [x] Wrote unit tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,8 +15,8 @@
 ### Improvements
 
 ### Bug fixes
-- [\#252](https://github.com/arkworks-rs/algebra/pull/252) (ark-ff) Fix prime field sampling when `REPR_SHIFT_BITS` is 64.
 
+- [\#252](https://github.com/arkworks-rs/algebra/pull/252) (ark-ff) Fix prime field sampling when `REPR_SHIFT_BITS` is 64.
 
 ## v0.2.0
 
@@ -33,6 +33,7 @@ The main features of this release are:
 - Adding new traits for basic curve cycles and pairing based curve cycles
 
 ### Breaking changes
+
 - [\#20](https://github.com/arkworks-rs/algebra/pull/20) (ark-poly) Move univariate DensePolynomial and SparsePolynomial into a
     univariate sub-crate. Make this change by:
     find w/ regex `ark_poly::(Dense|Sparse)Polynomial`, and replace with `ark_poly::univariate::$1Polynomial`.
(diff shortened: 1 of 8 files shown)


weikengchen:
@ValarDragon Hi Dev, there is a markdown linter error that seems strange---I feel superlinter's current docker may have a possible conflict with Rust at this moment.

ff/src/biginteger/macros.rs
```diff
@@ -352,5 +352,48 @@ macro_rules! bigint_impl {
...
+                let mut limbs = [0u64; $num_limbs];
+
+                for i in 0..$num_limbs {
+                    tmp >>= 64;
+                    tmp <<= 64;
+
+                    limbs[i] = (tmp2 - tmp.clone()).to_u64().unwrap();
+
+                    tmp >>= 64;
+                    tmp2 = tmp.clone();
```
daira, line 374: I see how this works, but it seems to have quadratic cost in the number of limbs. Is there no better way?

ff/src/biginteger/macros.rs
```diff
@@ -352,5 +352,48 @@ macro_rules! bigint_impl {
...
+            #[inline]
+            fn into(self) -> BigUint {
+                use num_traits::Zero;
+
+                let mut sum = BigUint::zero();
+
+                for i in (0..$num_limbs).rev() {
+                    sum <<= 64;
+                    sum += self.0[i];
+                }
```
daira, line 393: This also seems to have quadratic cost in the number of limbs.

ff/src/biginteger/macros.rs
weikengchen: You are right. Maybe I should use a better way: BigUint has `from_bytes_le` and `to_bytes_le`, from which I may have a better construction.

daira (approved):
On the topic of how explicit conversions should be, they absolutely need to be explicit between different fields, and between integers and fields, and from an integer to a potentially shorter integer type (i.e. where the value can change).

weikengchen:
Two updates: (1) I changed the algorithm, which tries to use the to/from_bytes_le as much as possible, though sometimes I need to make do with the second-best (2) The two traits are implemented for `PrimeField` as well.

ff/src/biginteger/macros.rs
```diff
@@ -352,5 +352,36 @@ macro_rules! bigint_impl {
...
+                        let mut chunk_padded = [0u8; 8];
+                        for j in 0..chunk.len() {
+                            chunk_padded[j] = chunk[j];
+                        }
+                        limbs[i] = u64::from_le_bytes(chunk_padded)
+                    });
+
+                Self(limbs)
+            }
+        }
```
Pratyush, line 378: Maybe we change this to `TryFrom`? In case the `BigUInt is bigger?

ff/src/fields/macros.rs
```diff
@@ -717,5 +717,19 @@ macro_rules! impl_Fp {
                 self.0.zeroize();
             }
         }
+
+        impl<P: $FpParameters> From<num_bigint::BigUint> for $Fp<P> {
+            #[inline]
+            fn from(val: num_bigint::BigUint) -> $Fp<P> {
+                $Fp::<P>::from_repr(<$BigIntegerType>::from(val)).unwrap()
```
Pratyush, line 724: For this, let's use `from_bytes_mod_order`, or convert `P::MODULUS` to `BigUint` and then use the modulo operator there.

ff/src/biginteger/macros.rs
weikengchen: Done.

ff/src/fields/macros.rs
weikengchen: Done.

ff/src/biginteger/macros.rs
weikengchen: (more fixings needed. on the way)

weikengchen:
Done the changes.

ff/src/biginteger/macros.rs
```diff
@@ -352,5 +352,45 @@ macro_rules! bigint_impl {
...
+
+                    bytes
+                        .chunks(8)
+                        .into_iter()
+                        .enumerate()
+                        .for_each(|(i, chunk)| {
+                            let mut chunk_padded = [0u8; 8];
+                            for j in 0..chunk.len() {
+                                chunk_padded[j] = chunk[j];
+                            }
```
daira, line 380:
```suggestion
                            chunk_padded[..chunk.len()].copy_from_slice(chunk);
```

daira (approved):
Untested ACK

weikengchen:
Done.

weikengchen:
All checks have passed, and the code is ready for review.

CHANGELOG.md
```diff
@@ -10,12 +12,13 @@
 - [\#261](https://github.com/arkworks-rs/algebra/pull/261) (ark-ff) Add support for 448-bit integers and fields.
 - [\#263](https://github.com/arkworks-rs/algebra/pull/263) (ark-ff) Add `From<iXXX>` implementations to fields.
 - [\#265](https://github.com/arkworks-rs/algebra/pull/265) (ark-serialize) Add hashing as an extension trait of `CanonicalSerialize`.
+- [\#280](https://github.com/arkworks-rs/algebra/pull/280) (ark-ff) Add `Into<BigUint>` and `From<BigUint>` implementations to `BigInteger` and `PrimeField`.
```
daira, line 15: This is outdated.

weikengchen: Good catch. Let me rename the PR as well.

daira (approved):
Untested ACK

weikengchen:
Ready for review. 

The linter test failure is related to a separate reason: #283

Pratyush (approved):

PR #286, opened 2021-05-28
test mdlinter change

ValarDragon:
Looking into why mdlinter isn't working, making a PR just to get it to run

---

write a little note why.

- [ ] Wrote unit tests

### DIFF
--- a/.github/workflows/mdlinter.yml
+++ b/.github/workflows/mdlinter.yml
@@ -7,7 +7,6 @@ on:
     paths:
       - "**.md"
   pull_request:
-    branches: [master]
     paths:
       - "**.md"
 
@@ -17,7 +16,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout
-        uses: actions/checkout@v1
+        uses: actions/checkout@v2
       - name: Lint Code Base
         uses: docker://github/super-linter:latest
         env:
--- a/.github/workflows/mdlinter.yml
+++ b/.github/workflows/mdlinter.yml
@@ -16,8 +16,8 @@ jobs:
     name: Markdown linter
     runs-on: ubuntu-latest
     steps:
-      - name: Checkout Code
-        uses: actions/checkout@v2
+      - name: Checkout
+        uses: actions/checkout@v1
       - name: Lint Code Base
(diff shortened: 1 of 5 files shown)


ValarDragon:
hrmm, looks like its working fine on master, maybe it only fails on PR's from outside the repo?

weikengchen:
Yes. I previously fixed the linter failure by upgrading the linter docker image version (due to a bug related to GCC, #283). However, it is unable to handle #259 since it is in a different repo, and it seems that currently linter is unable to get that (linter pulls the master and then tries to switch the branch).

ValarDragon:
Thanks for fixing that in 283!

I'm a bit confused why the correct branch doesn't get checked out, since the github actions checkout should handle going to the fork

weikengchen:
A possibility, which I am totally unsure about, is whether the pull-request parameters are omittable? Since `ci.yml` does not have them.

https://github.com/arkworks-rs/algebra/blob/master/.github/workflows/mdlinter.yml#L10

(I am totally using heuristics here.)

ValarDragon:
Lets try removing that, and see if it fixes anything

weikengchen (approved):

weikengchen:
I have the feeling that we need to merge this in order to test. Merge?

ValarDragon:
Yeah, lets merge this, and then check it on the next PR from outside the repo

weikengchen:
Based on my rerunning of https://github.com/arkworks-rs/algebra/runs/2697545270, nope. 

I will we will revisit this issue when we have more ideas.

weikengchen:
The current idea seems to be this: (1) before docker super-linter, the branch has been switched into an intermediate branch created by GitHub, like `pull/279/merge` (2) super-linter wants to build a list of files in master, and then sees which files have been changed.

https://github.com/github/super-linter/blob/3dc85fc3bc1f47c6cf14831c666f569b153e9fb9/lib/functions/buildFileList.sh#L67

And the failure seems to be right there.

One possibility: We can fix this by setting `VALIDATE_ALL_CODEBASE = true`.

ValarDragon:
Thanks for tracing that, lets set that to true then

PR #287, opened 2021-06-01
Resolve linter failure on PRs from external repos

weikengchen:
## Description

As discussed in #286, we discovered that in a number of PRs where the base repo is external, the linter fails.

This seems to be a bug of the super-linter, but fixing it is beyond our capacity. We noticed that the problem is related to `VALIDATE_ALL_CODEBASE = false`, in which case the super-linter will switch branches and find files that have been changed, and the super-linter will only perform linting on those files. The failure happens during the branch switch.

This PR, therefore, changes `VALIDATE_ALL_CODEBASE` to `true`. We expect that this change can get rid of such failures.

---

write a little note why.

N/A:
- [ ] Wrote unit tests

### DIFF
--- a/.github/workflows/mdlinter.yml
+++ b/.github/workflows/mdlinter.yml
@@ -21,7 +21,7 @@ jobs:
         uses: docker://github/super-linter:latest
         env:
           LINTER_RULES_PATH: .github
-          VALIDATE_ALL_CODEBASE: false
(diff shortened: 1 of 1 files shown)


ValarDragon (approved):

ValarDragon:
Lets hope this fixes it 🙏

Issue #288, opened 2021-06-01
Implement Sum trait for Group elements


ggkitsas:

weikengchen:
Ideally, it should have been already there: https://github.com/arkworks-rs/algebra/blob/master/ec/src/group.rs#L39

(untested)

weikengchen:
Oh wait I know: GroupAffine by default does not implement the `Group` trait and does not have "sum". 

Because GroupAffine cannot nicely represent "zero", which can happen when:
- summing an empty vector
- summing aG and -aG (e.g., part of the natural computation)

But G1Projective should have it.

weikengchen:
Thinking about that further, it seems also possible to add a Sum implementation to GroupAffine.

But I want to caution that the computation might be faster if all are done on projective curves (avoid some back-and-forth conversions).

weikengchen:
Maybe the performance is a strong motivation toward adding a Sum trait.

The Sum trait can make sure that it adds a bunch of points together (over projective curves) and then does only one conversion back to the affine curve, which can avoid many unnecessary inversions over the field.

PR #289, opened 2021-06-01
Implement the `Sum` trait for all `AffineCurve`

weikengchen:
## Description

George Gkitsas mentions that it would be convenient to have a `Sum` trait for `AffineCurve`, specifically SW curve's `GroupAffine`.

This PR adds the requirement that all `AffineCurve` must implement the `Sum` trait. 

It is useful to have this trait: a user who uses `+` or `+=` in SW's AffineCurve indeed would pay performance overhead --- there are unnecessary conversions between ProjectiveCurve and AffineCurve during each addition. A `Sum` trait implementation that avoids such unnecessary conversions can significantly boost the performance.

closes: #288 

---

write a little note why.

Feels skippable:
- [ ] Wrote unit tests

### DIFF
--- a/ec/src/models/short_weierstrass_jacobian.rs
+++ b/ec/src/models/short_weierstrass_jacobian.rs
@@ -278,6 +278,22 @@ impl<P: Parameters> Default for GroupAffine<P> {
     }
 }
 
+impl<P: Parameters> core::iter::Sum<Self> for GroupAffine<P> {
+    fn sum<I: Iterator<Item = Self>>(iter: I) -> Self {
+        iter.map(|x| x.into_projective())
+            .sum::<GroupProjective<P>>()
+            .into()
+    }
+}
+
+impl<'a, P: Parameters> core::iter::Sum<&'a Self> for GroupAffine<P> {
+    fn sum<I: Iterator<Item = &'a Self>>(iter: I) -> Self {
+        iter.map(|x| x.into_projective())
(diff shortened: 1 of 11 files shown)


ValarDragon (approved):
utACK

ec/src/models/short_weierstrass_jacobian.rs
```diff
@@ -278,6 +278,22 @@ impl<P: Parameters> Default for GroupAffine<P> {
...
+        iter.map(|x| x.into_projective())
+            .sum::<GroupProjective<P>>()
+            .into()
+    }
+}
+
+impl<'a, P: Parameters> core::iter::Sum<&'a Self> for GroupAffine<P> {
+    fn sum<I: Iterator<Item = &'a Self>>(iter: I) -> Self {
+        iter.map(|x| x.into_projective())
+            .sum::<GroupProjective<P>>()
```
Pratyush, line 292: It would be faster to use mixed addition, no?

weikengchen: Good point. Let me double-check. Yes, it is very likely that the conversions from affine to projective may be unnecessary as well.

weikengchen: (though the latter is cheap)

weikengchen: Done in #291

Issue #41, opened 2020-11-15
Simplify `Curve` API


Pratyush:
Currently we have a proliferation of 2-3 different APIs related to groups and curves: `Group`, `ProjectiveCurve` and `AffineCurve`. This causes unnecessary imports for users, and increases the API surface of the library. @ValarDragon and I developed the following simpler interface for group-related operations in `ark-ec`.

```rust
/// Represents (elements of) a group having prime order `r`.
pub trait Group:
	'static
    + Sized
	+ Eq
    + Copy
    + Default
    + Send
    + Sync
    + Hash
    + Debug
    + Display
    + Zero
    + Neg<Output = Self>
    + Add<Self, Output = Self>
    + Sub<Self, Output = Self>
    + AddAssign<Self>
    + SubAssign<Self>
    + MulAssign<<Self as ProjectiveCurve>::ScalarField>
    + for<'a> Add<&'a Self, Output = Self>
    + for<'a> Sub<&'a Self, Output = Self>
    + for<'a> AddAssign<&'a Self>
    + for<'a> SubAssign<&'a Self>
    + core::iter::Sum<Self>
    + for<'a> core::iter::Sum<&'a Self>
    + From<<Self as ProjectiveCurve>::CanonicalRepr>
{
	/// The scalar field `F_r`, where `r` is the order of this group.
    type ScalarField: PrimeField + SquareRootField;

	/// The canonical representation of elements of this group.
	/// For example, in elliptic curve groups, this can be defined to be the affine representation of curve points.
    type CanonicalRepr:
        Copy + Eq + Debug + Display + From<Self> + Into<Self> + UniformRand + Zero + Neg 
        + Mul<<<Self::ScalarField> as PrimeField>::BigInt, Output = Self>;

    /// Returns a fixed generator of this group.
    #[must_use]
    fn generator() -> Self::CanonicalRepr;


    /// Converts `self` into the canonical representation.
    fn to_canonical(&self) -> Self::CanonicalRepr {
        (*self).into()
    }

    /// Normalizes a batch of `self` elements into canonical form.
    fn batch_to_canonical(v: &[Self]) -> Vec<Self::CanonicalRepr>;

    /// Doubles `self`.
    #[must_use]
    fn double(&self) -> Self {
        let mut copy = *self;
        copy.double_in_place();
        copy
    }

    /// Double `self` in place.
    fn double_in_place(&mut self) -> &mut Self;

[...]

Issue #268, opened 2021-04-28
Change `GroupAffine`/`GroupProjective` to maintain invariants


hdevalence:
Currently the `GroupAffine` and `GroupProjective` structs do not maintain either of these invariants:

* the point is on the curve;
* the point is in the prime-order subgroup;

The `GroupAffine` struct provides methods to check these, but `GroupProjective` does not.  The API could be improved by:

- [ ] removing the invariant check methods;
- [ ] adding the invariant checks to the `new()` method and making it fallible;
- [ ] adding immutable field accessors `fn x(&self) -> &F`, `fn y(&self) -> &F`, etc;
- [ ] making the coordinate fields private.

This ensures that any users of the structures outside of the module subtree can't violate these invariants.

daira:
Please consider using the `zkcrypto/group` and `zkcrypto/ff` crates (the latter probably without the "derive" feature), if Arkworks' group APIs are going to change. These crates are pretty stable now (despite the 0.9 version number) and used in production for Zcash and the `bls12_381` crate, so interoperability with them would be very useful, and they're quite minimal dependencies.

hdevalence:
As a user rather than a maintainer of arkworks, I don't really have a stake in this, but I'm not sure it's fair to describe those crates as being minimal dependencies -- regardless of code size or number of transitive dependencies, a fundamental abstraction like a group or field trait is a significant dependency, affecting nearly every part of the codebase, and switching to it isn't trivial.

burdges:
If this happens, it's worth retaining some `new_unchecked` interface for points not in the prime order subgroup:  You can avoid doublings like `((..((c + a_1) + a_2)..) + a_n) - c` with `c` independent.  I've always seen this trick with DDH yielding independence of `c`, but I've heard people argue for also placing`c` in another coset, and maybe that's useful sometimes for reducing constraint count at the cost of fragility elsewhere.  This seems orthogonal to everything else discussed here.

daira:
@hdevalence That's why I only suggested switching to it in the case that Arkworks' APIs were going to change in order to make essentially the same `PrimeGroup`/`CofactorGroup` distinction that the `group` crate makes — which had been proposed.

Pratyush:
[...]

Issue #27, opened 2020-11-09
Clarify guarantees in traits around point validity


ValarDragon:
The elliptic curve point API's ([AffineCurve](https://github.com/arkworks-rs/algebra/blob/b4c3b438f63dcbcf854a6dbcaab5e983f19789fb/ec/src/lib.rs#L221), [ProjectiveCurve](https://github.com/arkworks-rs/algebra/blob/b4c3b438f63dcbcf854a6dbcaab5e983f19789fb/ec/src/lib.rs#L119)) currently claim that the interface should be providing quite strong guarantees, all points should be on the curve, and moreover in the prime order subgroup.

However this guarantee isn't achieved at the moment. 

I've only looked at short weierstrass jacobian curves so far (which encompasses the BLS12 curves), and they don't guarantee that points are on the curve, or that they are in the prime order subgroup for AffineCurve.

The [`new()`](https://github.com/arkworks-rs/algebra/blob/master/ec/src/models/short_weierstrass_jacobian.rs#L72) method for AffineGroup (which implements AffineCurve) does not ensure that points are on the curve.

Also the sw curve's Affine Curve implementation of [`from_random_bytes()`](https://github.com/arkworks-rs/algebra/blob/b4c3b438f63dcbcf854a6dbcaab5e983f19789fb/ec/src/models/short_weierstrass_jacobian.rs#L179) uses `get_point_from_x()`, which as stated in its comments provides no guarantees around being in the prime order subgroup.

I believe that the trait comments should drop the description that they are in the prime order subgroup, and we should have new traits that enforce that this is the case. (Or alternatively, make the two above methods enforce the claimed properties)

Pratyush:
@kobigurk @paberr do you have any thoughts around this? We could have a separate trait that handles the boundary conversions, and leave `AffineCurve` to be only prime order.

Pratyush:
One idea would be to put these methods in `SWModelParameters` and in `TEModelParameters` (I don't see how to put them in `ModelParameters`, because the trait doesn't understand the notion of `AffineCurve`.

kobigurk:
Good point! In my opinion, the current methods are useful and needed so I'd leave them as is.
New traits/methods sound like a good idea to me. Ideally they'd be wrappers that check validity and maybe one that multiplies by the cofactor, but that should be marked clearly. That bit us on the gadget side when we migrated our pre-refactor gadgets to the new constrain system.

Pratyush:
[...]

PR #294, opened 2021-06-05
Improve `ec` traits.

Pratyush:
## Description

closes: #41.

The aim is to also address #27 and #268.

---

write a little note why.

- [ ] Wrote unit tests


ff/src/fields/mod.rs
```diff
@@ -209,6 +209,94 @@ pub trait Field:
     }
 }
 
+/// Fields that have a cyclotomic subgroup, and which can leverage efficient inversion
+/// and squaring operations for elements in this subgroup.
+pub trait CyclotomicField: Field {
```
Pratyush, line 214: New field that should be implemented for fields that have more efficient operations for their cyclotomic subgroup, such as faster inverses or squaring.

ff/src/fields/mod.rs
```diff
@@ -209,6 +209,94 @@ pub trait Field:
...
+            let naf = crate::biginteger::arithmetic::find_wnaf(e.as_ref());
+            exp_loop(self, naf.into_iter().rev())
+        } else {
+            exp_loop(self, BitIteratorBE::without_leading_zeros(e.as_ref()).map(|e| e as i8))
+        };
+    }
+}
+
+/// Helper function to calculate the double-and-add loop for exponentiation.
+fn exp_loop<F: CyclotomicField, I: Iterator<Item = i8>>(f: &mut F, e: I) {
```
Pratyush, line 274: Helper method to abstract over the NAF and normal double-and-add loops. Can't be a closure due to weird Rust type inference issues.

ff/src/fields/models/fp12_2over3over2.rs
```diff
@@ -50,88 +50,19 @@ impl<P: Fp12Parameters> QuadExtParameters for Fp12ParamsWrapper<P> {
     fn mul_base_field_by_frob_coeff(fe: &mut Self::BaseField, power: usize) {
         fe.mul_assign_by_fp2(Self::FROBENIUS_COEFF_C1[power % Self::DEGREE_OVER_BASE_PRIME_FIELD]);
     }
-
-    fn cyclotomic_exp(fe: &Fp12<P>, exponent: impl AsRef<[u64]>) -> Fp12<P> {
```
Pratyush, line 54: Moved to the trait.

ff/src/fields/models/fp12_2over3over2.rs
```diff
@@ -203,30 +134,79 @@ impl<P: Fp12Parameters> Fp12<P> {
...
+        self.c1.sub_assign(&bb);
+        self.c0 = bb;
+        self.c0 = P::mul_fp6_by_nonresidue(&self.c0);
+        self.c0.add_assign(&aa);
     }
 }
 
-// TODO: make `const fn` in 1.46.
-pub fn characteristic_square_mod_6_is_one(characteristic: &[u64]) -> bool {
+pub const fn characteristic_square_mod_6_is_one(characteristic: &[u64]) -> bool {
```
Pratyush, line 196: This can now be a `const fn`.

ff/src/fields/models/fp2.rs
```diff
@@ -99,3 +100,12 @@ impl<P: Fp2Parameters> Fp2<P> {
         self.c1 *= other;
     }
 }
+
+impl<P: Fp2Parameters> CyclotomicField for Fp2<P> {
+    fn cyclotomic_inverse_in_place(&mut self) -> Option<&mut Self> {
```
Pratyush, line 105: We only have faster inverses here.

ff/src/fields/models/fp4.rs
```diff
@@ -49,6 +50,15 @@ impl<P: Fp4Parameters> QuadExtParameters for Fp4ParamsWrapper<P> {
 
 pub type Fp4<P> = QuadExtField<Fp4ParamsWrapper<P>>;
 
+impl<P: Fp4Parameters> CyclotomicField for Fp4<P> {
+    fn cyclotomic_inverse_in_place(&mut self) -> Option<&mut Self> {
```
Pratyush, line 54: Same for Fp4.

ff/src/fields/models/fp6_2over3.rs
```diff
@@ -23,6 +23,15 @@ pub trait Fp6Parameters: 'static + Send + Sync {
     }
 }
 
+impl<P: Fp6Parameters> CyclotomicField for Fp6<P> {
+    fn cyclotomic_inverse_in_place(&mut self) -> Option<&mut Self> {
```
Pratyush, line 27: I think technically we can also have cyclotomic squaring ala Fp12, but I didn't want to implement that in this PR.

ff/src/fields/models/quadratic_extension.rs
```diff
@@ -89,34 +89,6 @@ pub trait QuadExtParameters: 'static + Send + Sync + Sized {
     /// A specializable method for multiplying an element of the base field by
     /// the appropriate Frobenius coefficient.
     fn mul_base_field_by_frob_coeff(fe: &mut Self::BaseField, power: usize);
-
-    /// A specializable method for exponentiating that is to be used
-    /// *only* when `fe` is known to be in the cyclotommic subgroup.
-    fn cyclotomic_exp(fe: &QuadExtField<Self>, exponent: impl AsRef<[u64]>) -> QuadExtField<Self> {
```
Pratyush, line 95: There's no need for this to be in the `QuadExtParameters` trait anymore, which simplifies things slightly.

ff/src/fields/models/quadratic_extension.rs
```diff
@@ -153,11 +125,6 @@ impl<P: QuadExtParameters> QuadExtField<P> {
         self.c1 = -self.c1;
     }
 
-    /// This is only to be used when the element is *known* to be in the cyclotomic subgroup.
-    pub fn cyclotomic_exp(&self, exponent: impl AsRef<[u64]>) -> Self {
-        P::cyclotomic_exp(self, exponent)
-    }
```
Pratyush, line 159: Moved to the `CyclotomicField` trait.

ff/src/fields/mod.rs
```diff
@@ -209,6 +209,94 @@ pub trait Field:
...
+        let mut result = *self;
+        result.cyclotomic_inverse_in_place()?;
+        Some(result)
+    }
+
+    /// Compute the inverse of [`self`]. See [`Self::INVERSE_IS_FAST`] for details.
+    /// Returns [`None`] if `self.is_zero()`, and [`Some`] otherwise.
+    fn cyclotomic_inverse_in_place(&mut self) -> Option<&mut Self> {
+        self.inverse_in_place()
+    }
```
Pratyush, line 248: The default implementation just uses slow inversion.

ff/src/fields/mod.rs
```diff
@@ -209,6 +209,94 @@ pub trait Field:
...
+    fn cyclotomic_square(&self) -> Self {
+        let mut result = *self;
+        result.cyclotomic_square_in_place();
+        result
+    }
+
+    /// Square [`self`] in place. By default this is computed using [`Self::square_in_place`], but for
+    /// degree 12 extensions, this can be computed faster than normal squaring.
+    fn cyclotomic_square_in_place(&mut self) -> &mut Self {
+        self.square_in_place()
```
Pratyush, line 233: The default implementation just uses normal squaring.

ff/src/fields/mod.rs
```diff
@@ -209,6 +209,94 @@ pub trait Field:
...
 
+/// Fields that have a cyclotomic subgroup, and which can leverage efficient inversion
+/// and squaring operations for elements in this subgroup.
+pub trait CyclotomicField: Field {
+
+    /// Is the inverse fast to compute? For example, in quadratic extensions, the inverse
+    /// can be computed at the cost of negating one coordinate, which is much faster than
+    /// standard inversion.
+    /// By default this is `false`, but should be set to `true` for quadratic extensions.
+    const INVERSE_IS_FAST: bool = false;
```
Pratyush, line 220: By default inversion is slow, unless we are in a quadratic extension, in which case it can be implemented as just conjugation.

ff/src/biginteger/arithmetic.rs
```diff
@@ -52,7 +52,7 @@ pub(crate) fn mac_discard(a: u64, b: u64, c: u64, carry: &mut u64) {
     *carry = (tmp >> 64) as u64;
 }
 
-pub fn find_wnaf(num: &[u64]) -> Vec<i64> {
+pub fn find_wnaf(num: &[u64]) -> Vec<i8> {
```
Pratyush, line 55: Drive-by improvement to memory consumption of `find_wnaf`.

ec/src/models/mod.rs
```diff
@@ -1,12 +1,12 @@
...
-pub mod bls12;
-pub mod bn;
-pub mod bw6;
-pub mod mnt4;
-pub mod mnt6;
+// pub mod bls12;
+// pub mod bn;
+// pub mod bw6;
+// pub mod mnt4;
+// pub mod mnt6;
```
Pratyush, line 7: TODO: revert this once things compile.

ec/src/models/short_weierstrass_jacobian.rs
```diff
@@ -4,28 +4,26 @@ use ark_serialize::{
...
+    ops::{Add, AddAssign, MulAssign, Mul, Neg, Sub, SubAssign},
     vec::Vec,
+    rand::{
+        distributions::{Distribution, Standard},
+        Rng,
+    },
 };
 
 use ark_ff::{
-    bytes::{FromBytes, ToBytes},
```
Pratyush, line 14: We no longer implement `FromBytes` and `ToBytes`, as part of a move towards #5.

ec/src/models/short_weierstrass_jacobian.rs
```diff
@@ -41,37 +39,37 @@ use rayon::prelude::*;
...
     #[derivative(Debug = "ignore")]
     _params: PhantomData<P>,
 }
 
-impl<P: Parameters> PartialEq<GroupProjective<P>> for GroupAffine<P> {
-    fn eq(&self, other: &GroupProjective<P>) -> bool {
-        self.into_projective() == *other
+impl<P: Parameters> PartialEq<Projective<P>> for Affine<P> {
+    fn eq(&self, other: &Projective<P>) -> bool {
+        Projective::from(*self) == *other
```
Pratyush, line 52: TODO: specialize this to take advantage of the fact that self.z == 1.

ec/src/models/short_weierstrass_jacobian.rs
```diff
@@ -344,47 +338,29 @@ impl<P: Parameters> PartialEq for GroupProjective<P> {
...
+impl<P: Parameters> Default for Projective<P> {
     #[inline]
     fn default() -> Self {
         Self::zero()
     }
 }
 
-impl<P: Parameters> GroupProjective<P> {
+impl<P: Parameters> Projective<P> {
+    /// Construct a [`Self`] *without* validating that the coordinates form a valid point.
```
Pratyush, line 363: TODO: address #27 and #268

ec/src/msm/mod.rs
```diff
@@ -3,10 +3,7 @@ mod variable_base;
 pub use fixed_base::*;
 pub use variable_base::*;
 
-/// The result of this function is only approximately `ln(a)`
-/// [`Explanation of usage`]
-///
-/// [`Explanation of usage`]: https://github.com/scipr-lab/zexe/issues/79#issue-556220473
+/// The result of this function is only approximately `ln(a)`. See [here](https://github.com/arkworks-rs/snark/79#issue-556220473) for details.
```
Pratyush, line 6: drive-by fix

ec/src/pairing.rs
```diff
@@ -0,0 +1,201 @@
+use ark_ff::{One, Zero, PrimeField, SquareRootField, CyclotomicField};
+use ark_std::{io::{Read, Write}, fmt::{Display, Formatter, Result as FmtResult}};
+use ark_serialize::SerializationError;
+use ark_std::rand::{Rng, distributions::{Distribution, Standard}};
+use crate::*;
+
+pub trait Pairing: Sized + 'static + Copy + Debug + Sync + Send + Eq {
```
Pratyush, line 7: Revampled `Pairing` trait.

ec/src/pairing.rs
```diff
@@ -0,0 +1,201 @@
...
+    /// This is the scalar field of `Self::G1` and `Self::G2`.
+    type ScalarField: PrimeField + SquareRootField;
+
+    /// An element of G1.
+    type G1: CurveGroup<ScalarField = Self::ScalarField>
+        + MulAssign<Self::ScalarField>; // needed due to https://github.com/rust-lang/rust/issues/69640
```
Pratyush, line 13: Note that there is no longer a separate `G1Affine` and `G1Projective`.

ec/src/pairing.rs
```diff
@@ -0,0 +1,201 @@
...
+}
+
+impl<P: Pairing> Distribution<PairingOutput<P>> for Standard {
+    #[inline]
+    fn sample<R: Rng + ?Sized>(&self, rng: &mut R) -> PairingOutput<P> {
+        // Sample a random G1 element
+        let g1 = P::G1::rand(rng);
+        // Sample a random G2 element
+        let g2 = P::G2::rand(rng);
+        P::pairing(g1, g2)
```
Pratyush, line 152: Sampling a random GT element requires sampling random g1 and g2 elements, and then pairing them.

ec/src/pairing.rs
```diff
@@ -0,0 +1,201 @@
...
+    type ScalarField = P::ScalarField;
+    type UniqueRepr = Self;
+
+    fn generator() -> Self::UniqueRepr {
+        // TODO: hardcode these values.
+        // Sample a random G1 element
+        let g1 = P::G1::generator();
+        // Sample a random G2 element
+        let g2 = P::G2::generator();
+        P::pairing(g1.into(), g2.into())
```
Pratyush, line 166: To obtain the GT generator, we just pair the G1 and G2 generators. Since this value is fixed, we can technically hardcode it, but that requires work, and is also unlikely to really be used anywhere.

ec/src/lib.rs
```diff
@@ -32,328 +30,159 @@ use zeroize::Zeroize;
...
 
-pub mod group;
+pub mod pairing;
+pub use self::pairing::*;
 
 pub mod msm;
 
 pub mod wnaf;
 
-pub trait PairingEngine: Sized + 'static + Copy + Debug + Sync + Send + Eq + PartialEq {
```
Pratyush, line 41: The `PairingEngine` trait is renamed to `Pairing` and the implementation now lives in the `pairing` module.

ec/src/pairing.rs
burdges: https://indiv0.github.io/lazycell/lazycell/
https://doc.rust-lang.org/std/lazy/index.html

ec/src/pairing.rs
burdges: I wonder if this impacts anything..

ff/src/fields/mod.rs
```diff
@@ -209,6 +209,97 @@ pub trait Field:
...
+    }
+}
+
+/// Helper function to calculate the double-and-add loop for exponentiation.
+fn exp_loop<F: CyclotomicField, I: Iterator<Item = i8>>(f: &mut F, e: I) {
+    let self_inverse = if F::INVERSE_IS_FAST {
+        f.cyclotomic_inverse().unwrap() // must exist because self is not zero.
+    } else {
+        F::one()
+    };
```
ValarDragon, line 281:
```suggestion
    // If the inverse is fast and we're using wnaf, we compute the base points inverse.
    // Otherwise we do nothing with the variablel, so we default it to one.
    let self_inverse = if F::INVERSE_IS_FAST {
        f.cyclotomic_inverse().unwrap() // The inverse must exist because self is not zero.
    } else {
        F::one()
    };
```

ec/src/pairing.rs
ValarDragon: Does it not suffice to sample a random GT element, and then take the final exponentiation on it?

ff/src/fields/mod.rs
```diff
@@ -209,6 +209,97 @@ pub trait Field:
     }
 }
 
+/// Fields that have a cyclotomic subgroup, and which can leverage efficient inversion
+/// and squaring operations for elements in this subgroup.
```
ValarDragon, line 213:
```suggestion
/// Fields that have a cyclotomic subgroup, and which can leverage efficient inversion
/// and squaring operations for elements in this subgroup.
/// If a field has multiplicative order p^d - 1, the cyclotomic subgroups refer to subgroups of order φ_n(p),
/// for any n < d, where φ_n is the [nth cyclotomic polynomial](https://en.wikipedia.org/wiki/Cyclotomic_polynomial)
```

ff/src/fields/models/fp2.rs
```diff
@@ -99,3 +100,13 @@ impl<P: Fp2Parameters> Fp2<P> {
         self.c1 *= other;
     }
 }
+
+impl<P: Fp2Parameters> CyclotomicField for Fp2<P> {
+    const INVERSE_IS_FAST: bool = true;
+    fn cyclotomic_inverse_in_place(&mut self) -> Option<&mut Self> {
```
ValarDragon, line 106:
```suggestion
    fn cyclotomic_inverse_in_place(&mut self) -> Option<&mut Self> {
    /// As the multiplicative subgroup is of order p^2 - 1, the only non-trivial cyclotomic subgroup is of order p+1
    /// Therefore, for any element in the cyclotomic subgroup, we have that `x^(p+1) = 1`.
    /// Recall that `x^(p+1)` in a quadratic extension field is equal to the norm in the base field, so we have that
    /// `x * x.conjugate() = 1`. By uniqueness of inverses, for this subgroup, x.inverse() = x.conjugate()

```

ValarDragon:
Cyclotomic field work LGTM, I suggest we move that to a standalone PR to avoid a massive PR at the end of the refactor that will be super hard to review

ec/src/pairing.rs
Pratyush: That's a good idea, it should work. That can also provide us with a generator, right?

ec/src/pairing.rs
Pratyush: ooh I wonder if we should use that for other APIs too now...

ec/src/pairing.rs
burdges: It'd only provide a generators independent of pairing outputs, but most protocols need a generator e(G_1,G_2) if they need a G_t generator at all.

ec/src/pairing.rs
burdges: Atomic laziness incurs a performance hit of like 2ns I hear, but afaik most protocols access their generator only rarely, so whatever.  It's more likely recomputation once per thread or once per exec causes slowness, especiall in WASM, so then the fastest would be pre-computed or if const generics work then 
```
const GENERATOR: Self::UniqueRepr = {
    let g1 = P::G1::generator();
    let g2 = P::G2::generator();
    P::pairing(g1.into(), g2.into())
};
```
but you could just leave a performance TODO for whoever cares about those cases.

As an aside, we've layers of tables that depend upon tables in https://github.com/w3f/rs-ec-perf/ so I used https://gitlab.com/okannen/static_init to avoid the 2ns performance hit, but it brings portability issues so likely unsuitable here.  As a further aside, rand had a long discussion around mutable thread locals in https://github.com/rust-random/rand/issues/968

ec/README.md
```diff
@@ -6,18 +6,18 @@
...
-* [*Twisted Edwards*](https://github.com/arkworks-rs/algebra/blob/master/ec/src/models/twisted_edwards_extended.rs) curves. The `AffineCurve` in this case is in standard Twisted Edwards curve representation, whereas the `ProjectiveCurve` uses points in [Extended Twisted Edwards Coordinates](https://eprint.iacr.org/2008/522.pdf).
+* [*Short Weierstrass*](https://github.com/arkworks-rs/algebra/blob/master/ec/src/models/short_weierstrass.rs) curves. The underlying canonical representation uses standard short Weierstrass coordinates, while arithmetic relies on formulae over [Jacobian Coordinates](https://en.wikibooks.org/wiki/Cryptography/Prime_Curve/Jacobian_Coordinates).
+* [*Twisted Edwards*](https://github.com/arkworks-rs/algebra/blob/master/ec/src/models/twisted_edwards.rs) curves. The underlying canonical representation uses standard twisted Edwards coordinates, while arithmetic relies on formulae over [Extended Twisted Edwards Coordinates](https://eprint.iacr.org/2008/522.pdf).
```
ValarDragon, line 23:
```suggestion
* [*Twisted Edwards*](https://github.com/arkworks-rs/algebra/blob/master/ec/src/models/twisted_edwards.rs) curves. The underlying canonical representation uses standard Twisted Edwards coordinates, while arithmetic relies on formulae over [Extended Twisted Edwards Coordinates](https://eprint.iacr.org/2008/522.pdf).
```

ec/src/pairing.rs
```diff
@@ -0,0 +1,281 @@
...
+    #[must_use]
+    fn final_exponentiation(mlo: MillerLoopOutput<Self>) -> Option<PairingOutput<Self>>;
+
+    /// Computes a product of pairings.
+    #[must_use]
+    fn product_of_pairings<'a>(
+        i: impl IntoIterator<Item = &'a (Self::G1Prepared, Self::G2Prepared)>,
+    ) -> PairingOutput<Self> {
+        Self::final_exponentiation(Self::miller_loop(i)).unwrap()
+    }
```
bhgomes, line 60: Can we please use a pair of references instead of a reference to a pair?

```rust
impl IntoIterator<Item = (&'a Self::G1Prepared, &'a Self::G2Prepared)>
```

It's almost impossible to compose this API with others that involve calling the pairing algorithm more than once.

Pratyush: Yes, that's a good suggestion.

Issue #299, opened 2021-07-01
Inconsistent hash for GroupProjective struct


alxiong:
currently the `trait Hash` derived for `GroupProjective` is inconsistent, here's a quick way to reproduce the bug:

```rust
use ark_ec::AffineCurve;
use ark_ec::ProjectiveCurve;
use ark_ed_on_bls12_381::{EdwardsAffine as JubjubPointAffine, EdwardsProjective as JubjubPoint};
use std::collections::HashMap;

#[test]
fn hash_trait_on_projective() {
    let mut map: HashMap<JubjubPoint, u64> = HashMap::new();
    let mut map2: HashMap<JubjubPointAffine, u64> = HashMap::new();

    let g = JubjubPoint::prime_subgroup_generator();
    let g_affine = JubjubPointAffine::prime_subgroup_generator();
    map.insert(g.double(), 2);
    map2.insert(g_affine, 2);
    assert!(map2.contains_key(&g_affine));
    assert!(map.contains_key(&g));  // WILL FAIL !!
}
```

The solution is easy, just manually implement the Hash trait by converting to affine repr first then hash.

PR #300, opened 2021-07-01
manual impl Hash on GroupProjective for consistency

alxiong:
## Description

closes: #299 

---

write a little note why.

- [ ] Wrote unit tests
- [ ] ~Updated relevant documentation in the code~


alxiong:
@weikengchen Do you want me to add tests in `test-templates/src/curves.rs` for this?

weikengchen (approved):

weikengchen:
I think additional tests, though can discover this issue, may not be needed, given that the change is quite predictable.

weikengchen:
Thanks for the PR!

burdges:
Rust need some facility for warnings when using specific trait impl.  I think even `HashMap: Clone` in std runs into "if you're using this then you maybe fucked up".  This is not as serious as that one, but it indicates a possible mistake.

alxiong:
@burdges May I ask what's the context for this feedback? Are you suggesting we should do something different, revert back the changes or? (If so, would you help take a look at the original motivating case in #299 ?)

burdges:
Nope.  No change.  :)  It's likely someone needs this, but also likely someone would use this when they should use `batch_normalize` first.  I've just noticed many cases like this where maybe clippy should advise against using the impl, but the impl itself is necessary.

PR #302, opened 2021-07-04
Add the relaxed NAF

weikengchen:
## Description

This PR adds the relaxed NAF, which differs from the original `find_wnaf` (should be `find_naf`?) by allowing the top 2 bits to be both nonzeros. This can, in some cases, reduce the overhead.

The algorithm is indeed quite simple. The only case when the standard NAF may be suboptimal (note: NAF is optimal for hamming weight, but not the length + the hamming weight) is when the NAF ends with "-1 0 1", which can be systematically changed into "1 1". There are no other cases.

---

write a little note why.

- [x] Wrote unit tests


Pratyush:
Is this used anywhere? (Eg: for faster scalar muls?)

ff/src/biginteger/arithmetic.rs
```diff
@@ -104,3 +104,62 @@ pub fn find_wnaf(num: &[u64]) -> Vec<i64> {
 
     res
 }
+
+pub fn find_relaxed_naf(num: &[u64]) -> Vec<i64> {
```
ValarDragon, line 108: Can the relaxed naf definition be commented here

weikengchen:
> Is this used anywhere? (Eg: for faster scalar muls?)

I originally planned to use it for Poseidon. Unfortunately, for essentially all the prime numbers being considered, NAF is not better than the original form, because when NAF is better, it needs to use "inverse", and "inverse" here is costly, which requires 3 constraints to compute (it is needed to be compatible with the case that `d` is zero).

Maybe it would be more useful for EC scalar mul, in which "inverse" (aka negative) is cheap.

ff/src/biginteger/arithmetic.rs
weikengchen: Yes, let me add some description to it.

weikengchen:
On a separate note, should we rename `find_wnaf` to `find_naf`? there is no window.

Pratyush:
Yeah, `find_naf` makes sense.

Pratyush:
@weikengchen what's the status of this PR?

weikengchen:
I should be able to revisit this PR today or tomorrow.

ff/src/biginteger/arithmetic.rs
weikengchen: Added.

weikengchen:
The PR is ready to go. @Pratyush please take a look

ff/src/biginteger/arithmetic.rs
```diff
@@ -119,3 +119,75 @@ pub fn find_wnaf(num: &[u64]) -> Vec<i64> {
...
+        res.resize(len - 1, 0);
+    }
+
+    res
+}
+
+#[test]
+fn test_find_relaxed_naf_usefulness() {
+    let vec = find_relaxed_naf(&[12u64]);
+    assert_eq!(vec.len(), 4);
```
Pratyush, line 152: Maybe we can add here `find_naf(&12u64])` which has a shorter length?

Pratyush (approved):

ff/src/biginteger/arithmetic.rs
weikengchen: Yes. Added. It has a slightly longer len---5.

PR #303, opened 2021-07-04
Discuss: should we panic when powering 0 by 0?

weikengchen:
## Description

This is half PR half discussion.

For `pow` function, there is a corner case when the base is zero and the pow is also zero. 

There are two ways to handle it:
1. panic, as in this PR
2. let it be "1", which is indeed an accepted practice in some fields. 

Note that this treatment is sort of for "soundness". It is hard to imagine a concrete application where the developer wants to power a value by "0".

Any opinions? Note that letting it be "1" is simpler. 

---

write a little note why.

- [x] Wrote unit tests

not ready:

### DIFF
--- a/ff/src/fields/mod.rs
+++ b/ff/src/fields/mod.rs
@@ -178,6 +178,12 @@ pub trait Field:
     /// least significant limb first.
     #[must_use]
     fn pow<S: AsRef<[u64]>>(&self, exp: S) -> Self {
+        // handle the special case when self = 0 and exp = 0
+        if self.is_zero() {
+            assert_ne!(exp.as_ref().iter().filter(|x| **x != 0u64).count(), 0);
+            return Self::zero();
+        }
+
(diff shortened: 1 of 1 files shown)


weikengchen:
Note: Rust itself considers `0^0 = 1`. I think we just need to consider whether it will ever cause soundness problems in any application. Note that if we panic when 0^0, it may cause additional issues that the program may panic unnecessarily.
```
if exp == 0 {
     return 1;
}
```

ValarDragon:
I think setting it to 1 makes sense personally. I think risk of panic for a core algebraic routine will cause more error handling / cognitive overhead code bloat, for something that seems like a very low risk error to get introduced.

weikengchen:
I think let us settle that 0^0 = 1. I will update https://github.com/arkworks-rs/r1cs-std/pull/72 accordingly.

Issue #305, opened 2021-08-11
Migrate away from `llvm_asm` to `asm`


Pratyush:
`llvm_asm` is being deprecated: https://github.com/rust-lang/rust/pull/87590

weikengchen:
We need to boost the urgency of this. As it appears that the latest version of Rust stable has deprecated it.

weikengchen:
> @weikengchen This is the issue I mentioned during our meeting.
> 
> Using the latest release of this repository 0.3.0 and using the nightly channel of Rust.
> 
> I tried building the release locally as well as letting cargo take care of the package.
> 
> Here is the dump output: [dump.txt](https://github.com/arkworks-rs/algebra/files/8006072/dump.txt)
> 
> edit: Note this dump was generated from attempting to run Gemini's docs.

Got it. I think this is expected, as the nightly officially bans the llvm_asm. The stable version of Rust should be fine.

There is still one remaining fix necessary for the #306. I will watch closely.

s7117:
> Got it. I think this is expected, as the nightly officially bans the llvm_asm. The stable version of Rust should be fine.
> 
> There is still one remaining fix necessary for the #306. I will watch closely.

Thanks. I was using nightly as rust complains about using in Gemini `#![feature]` on the stable release channel.
[stable_dump.txt](https://github.com/arkworks-rs/algebra/files/8006555/stable_dump.txt)

weikengchen:
> > Got it. I think this is expected, as the nightly officially bans the llvm_asm. The stable version of Rust should be fine.
> > There is still one remaining fix necessary for the #306. I will watch closely.
> 
> Thanks. I was using nightly as rust complains about using in Gemini `#![feature]` on the stable release channel. [stable_dump.txt](https://github.com/arkworks-rs/algebra/files/8006555/stable_dump.txt)

This seems to be a Gemini issue now, so we will discuss somewhere else. 

The fix is probably this: use nightly, and change the Cargo.toml of Gemini to remove "asm" from its default feature list.

https://github.com/arkworks-rs/gemini/blob/main/Cargo.toml#L91

PR #306, opened 2021-08-12
Upgrade to new `asm` syntax

Pratyush:
## Description

closes: #305

There is one change in the assembly here. Earlier, we used to declare `P::INV` as an immediate input for the assembly block, but due to limitations of `const` in Rust, we can no longer do that, and must declare it as an input. We can probably work around this manually, but it would make things slightly uglier. I would benchmark the performance difference, but unfortunately I don't have access to an x86 machine atm. Could one of @jon-chuang or @ValarDragon benchmark this PR?

---

write a little note why.

- [ ] Wrote unit tests


Pratyush:
This is going to be stabilized soon: https://github.com/rust-lang/rust/issues/72016#issuecomment-964186287

I'd like to check whether the strategy in [this playground](https://play.rust-lang.org/?version=nightly&mode=release&edition=2021&gist=2dd56e74f835ad95ee9460e5c7898a2d) would allow for a cleaner impl (eg: getting rid of the funky macro for constructing the `asm` block), but otherwise we should move quickly to update to the new syntax.

<details>
<summary>Copy of playground code for posterity</summary>
```rust
#![feature(asm)]

mod intrinsics {
    #[inline(always)]
    pub unsafe fn adcx(a: &mut u64, b: u64) {
        asm! {
            "adcx {0}, {1}",
            inout(reg) *a,
            in(reg) b
        };
    }

    #[inline(always)]
    pub unsafe fn adox(a: &mut u64, b: u64) {
        asm! {
            "adox {0}, {1}",
            inout(reg) *a,
            in(reg) b
        };
    }

    #[inline(always)]
    pub unsafe fn clear_flags() {
        asm! {
            "xor {0}, {0}",
            out(reg) _
        }
    }

    // doesn't inline properly
    // pub use std::arch::x86_64::_mulx_u64 as mulx;

    #[inline(always)]
    pub unsafe fn mulx(a: u64, b: u64, lo: &mut u64, hi: &mut u64) {
        asm!(
            "mulx {2}, {1}, {0}",
            in(reg) b,
            lateout(reg) *lo,
            lateout(reg) *hi,
            in("rdx") a
        );
    }
}

/// doesn't do actual work, just demonstration
pub fn carry_chain(a: [u64; 2], b: [u64; 2]) -> u64 {
    use intrinsics::*;
    unsafe {
        // clear flags CF and OF
        intrinsics::clear_flags();
        let mut hi = 0;
        let mut lo = 0;

        mulx(a[0], b[0], &mut lo, &mut hi);
        adox(&mut lo, hi);
        adcx(&mut hi, b[0]);

        mulx(a[1], b[1], &mut lo, &mut hi);
        adox(&mut lo, hi);
        adcx(&mut hi, b[1]);

        lo
    }
}


#[test]
fn test_adcx() {
    let mut a = 5;
    let b = 3;
    unsafe {
        intrinsics::adcx(&mut a, b);
    }
    assert_eq!(a, 8);
}


#[test]
fn test_mulx() {
    let a = 5;
    let b = 3;
    let mut hi = 0;
    let mut lo = 0;
    unsafe {
        intrinsics::mulx(a, b, &mut lo, &mut hi);
    }
    assert_eq!(lo, 5 * 3);
    assert_eq!(hi, 0);
}
```
</details>

weikengchen:
There is a decision to make:

- Since 1.59.0, `asm!` is now stable.
- However, `asm_const` is not stable, but we do use `const` in asm in certain places.

Ideas? We can still have the assembly in nightly.

weikengchen:
Based on the CI, there might be a hard-to-find problem:

```
Error: test failed, to rerun pass '-p ark-ff --lib'
Caused by:
  process didn't exit successfully: `/home/runner/work/algebra/algebra/target/debug/deps/ark_ff-a63347d5ec55c228` (signal: 11, SIGSEGV: invalid memory reference)
```

ff/src/lib.rs
```diff
@@ -8,8 +8,10 @@
 )]
 #![allow(clippy::op_ref, clippy::suspicious_op_assign_impl)]
 #![cfg_attr(not(feature = "asm"), forbid(unsafe_code))]
-#![cfg_attr(use_asm, feature(llvm_asm))]
 #![cfg_attr(feature = "asm", deny(unsafe_code))]
+#![allow(stable_features)]
+#![cfg_attr(use_asm, feature(asm))]
```
Pratyush, line 13: I think we can just get rid of these? It's already stable in nightly, right?

weikengchen (approved):

ff/Cargo.toml
```diff
@@ -36,4 +36,3 @@ rustc_version = "0.4"
 default = []
 std = [ "ark-std/std", "ark-serialize/std" ]
 parallel = [ "std", "rayon", "ark-std/parallel" ]
-asm = []
```
Pratyush, line 39: Hm I don't think we should enable it by default, no? It's still unsafe that people might not want.

weikengchen: then let me add it back.

weikengchen: Resolved.

weikengchen:
One thing to check is whether the code can compile in non-x86-64 machines (aka arm). In the build.rs, I use `is_x86_feature_detected!`. I worried that it would be undefined in arm.

weikengchen:
For the previous comment---I think check_std does that job.

ff/build.rs
```diff
@@ -1,17 +1,24 @@
...
+
+    let is_x86_64 = target_arch == "x86_64";
+    let mut bmi2_enabled = false;
+    let mut adx_enabled = false;
+    if is_x86_64 {
+        bmi2_enabled = is_x86_feature_detected!("bmi2");
+        adx_enabled = is_x86_feature_detected!("adx");
+    }
+
+    let should_use_asm = is_asm_stable && asm_enabled && bmi2_enabled && adx_enabled && is_x86_64;
```
Pratyush, line 21: Hm should we just get rid of this, and instead change our minimum supported rust version to 1.59? And in the places that we use the assembly backend, we can directly gate it with a 
```rust
#[cfg(
    feature = "asm",
    target_feature = "bmi2",
    target_feature = "adx",
    target_arch = "x86_64"
)]
```

weikengchen: I think there is only one question: do we want to bump up the minimal supported rust version to 1.59?

That would clearly be ideal, and we can cut the `build.rs`.

Pratyush: I think it's fine personally

weikengchen: I would recommend doing that in a separate PR---the current stable is 1.58.1, and it would need 20+ days to be 1.59.0.

So, if we drop it, for the next three weeks, we may have some trouble.

weikengchen: I can add an issue and a separate PR.

Pratyush: Hm then I would still move most of the `cfg` logic out of the `build.rs`, but emit the `asm` cfg in `build.rs`, so that the `cfg` gate becomes
```rust
#[cfg(all(
    asm,
    target_feature = "bmi2",
    target_feature = "adx",
    target_arch = "x86_64"
))]
```

weikengchen: and the `build.rs` becomes the following? So it would in a shape that is ready to drop.

```
use rustc_version::Version;

fn main() {
    println!("cargo:rerun-if-changed=build.rs");
    
    let rust_version = rustc_version::version().unwrap();
    let is_asm_stable = rust_version >= Version::new(1, 59, 0);
    let asm_enabled = cfg!(feature = "asm");

    let should_use_asm = is_asm_stable && asm_enabled;
    if should_use_asm {
        println!("cargo:rustc-cfg=use_asm");
    }
}
```

Pratyush: Yup, exactly!

weikengchen: There is a complication: what do you think would be the best solution here?

https://github.com/arkworks-rs/algebra/blob/migrate-to-asm/ff/src/fields/models/fp/montgomery_backend.rs#L86

weikengchen: ah, it seems that I can simply do:

```
if N <= 6
                && N > 1
                && cfg!(all(
                    asm,
                    target_feature = "bmi2",
                    target_feature = "adx",
                    target_arch = "x86_64"
                ))
            {
```

Pratyush: We write `cfg!(all(
    use_asm,
    target_feature = "bmi2",
    target_feature = "adx",
    target_arch = "x86_64"
))`

weikengchen: another complication---actually this is a mistake:
https://github.com/arkworks-rs/algebra/blob/migrate-to-asm/ff/src/biginteger/mod.rs#L242

It just detects the `asm` feature flag but not other things. 

Seems to be complicated. 

Should we do `cfg_attr!` in the `lib.rs`?

Pratyush: That one is fine, as it doesn't require 1.59, nor `bmi2` or `adx`

weikengchen: Changed the logic. It appears that rust version has to be queried in build.rs. The rest can be moved out. 

I use `inline_asm_stable` since not all asm requires 1.59.0, as you pointed out.

weikengchen: `cfg(version(1.59))` is not yet stable, though, we can drop it when we decided that the minimal supported rust version is >=1.59.

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 ### Breaking changes
 
 - [\#300](https://github.com/arkworks-rs/algebra/pull/300) (`ark-ec`) Change the implementation of `Hash` trait of `GroupProjective` to use the affine co-ordinates.
+- [\#306](https://github.com/arkworks-rs/algebra/pull/306) (`ark-ff`) Remove `asm` feature flag.
```
Pratyush, line 8:
```suggestion
```

weikengchen: The other CHANGELOG entry also no longer applies. Any suggestion?

CHANGELOG.md
```diff
@@ -28,18 +28,18 @@
...
 - [\#301](https://github.com/arkworks-rs/algebra/pull/301) (`ark-ec`) Add `GLVParameters` trait definition.
 - [\#312](https://github.com/arkworks-rs/algebra/pull/312) (`ark-ec`) Add `is_in_correct_subgroup_assuming_on_curve` for all `SWModelParameters`.
+- [\#321](https://github.com/arkworks-rs/algebra/pull/321) (`ark-ff`) Change bigint conversions to impl `From` instead of `Into`.
 - [\#348](https://github.com/arkworks-rs/algebra/pull/348) (`ark-ec`) Add `msm:{Fixed,Variable}Base:msm_checked_len`.
 - [\#364](https://github.com/arkworks-rs/algebra/pull/364) (`ark-ec`) Add `ChunkedPippenger` to variable-base MSM.
 - [\#371](https://github.com/arkworks-rs/algebra/pull/371) (`ark-serialize`) Add serialization impls for arrays
 
 ### Improvements
 
+- [\#306](https://github.com/arkworks-rs/algebra/pull/306) (`ark-ff`, `ark-ff-asm`) Automatically use inline assembly from `asm!` when available.
```
Pratyush, line 42:
```suggestion
- [\#306](https://github.com/arkworks-rs/algebra/pull/306) (`ark-ff`, `ark-ff-asm`) Make the assembly backend available on `stable`.
```

Pratyush:
Thanks for getting this over the line, @weikengchen!

There's one remaining TODO that we should file an issue for: in the previous version of the backend, `Self::INV` was a constant in the generated assembly, but in the current version it's passed in a register. This decreases the number of available registers, increasing register and hence memory pressure.

Unfortunately making `Self::INV` into a constant requires stabilization of the `asm_const` and (part of) the `generic_const_exprs` features, so it might take a while.

weikengchen:
Talking about which, since we have already used the ugly `movq_zero!`, would it make sense to just hardcode the constants?

There is an issue---if we hardcode that as a constant, does it mean that Rust would generate different assembly codes for each different field? Or, is that already happening?

weikengchen:
My feeling is that `asm_const` is mostly a Rust side problem because it needs to be implemented in a general manner, but we are doing a very specific and simple case (`movq`).

Pratyush:
We can't quite hardcode the constants, because we don't know what they are until instantiation time in some downstream crate. We *could* provide a macro that's the equivalent of https://crates.io/crates/ff_derive, which people invoke on the `MontConfig` in the appropriate downstream crate, and that macro could generate the assembly with full knowledge of the various involved constants.

weikengchen:
...and this approach should also allow us to hardcode `modulus`, which at this moment is a memory access?

Pratyush:
Yeah, it would allow us to hardcode modulus also.

weikengchen:
With that, I think we are reaching the destination of single-field-element asm optimization. 

To improve further, we will need to do batching and SIMD.

Pratyush:
@weikengchen is this ready to merge? I can do it now if so.

weikengchen:
It is really to merge. I am waiting for you (technically the person who created the PR did the final decision right?)

PR #317, opened 2021-09-22
Let ark-ff specify the same `num-bigint` version as the rest of the repos

### LINKED ISSUE #316: CI Breakage on num-bigint 0.4.0
### **Problem**

 int_roundings has now been feature gated as an unstable feature, and is now a breaking change on recent nightly builds. 

Tracking issue: https://github.com/rust-lang/rust/issues/88581

In the issue, they note that num-bigint version 0.4.0 has this problem, which is the version that arkworks is using in ark-ff ( and possibly in other crates).

### **Quick fix**:

The simplest fix is to locally run `cargo update -p num-bigint` to update the num-bigint dependency in your lockfile or `cargo update` to update all dependencies including num-bigint.

### **Solution**:

Since num-bigint 0.3.3 or 0.4.2 has this fix applied, a solution would be to update to 0.4.2.

### **Reasons to not apply solution**:

- This is not breaking for stable, so may not be a priority since there is a quick fix.
- It is not clear to me that discussions around function naming have been resolved, so eagerly updating may require you to update again if the function names do change.

For the second point, it is possible to fix the version to 0.4.2 to stop the CI breaking on the now unstable feature and then update again at a later date, if the names do change.

weikengchen:
## Description

This is related to #316 but is not expected to be a solution.

There is a pending "unstable feature" in Rust 1.56 beta that requires `num-bigint` to be bumped up to `0.4.2` or higher to be able to compile.

My local test seems to show that the compiler would always retrieve the `0.4.2` version (since arkwork-rs never uses Cargo.lock). So this seems to be an issue for downstream applications to handle.

Nevertheless, we can improve the current code: only the `algebra` and `r1cs-std` repos use `num-bigint`, and `ark-ff` is the only one that is specifying `0.4.0`, while the rest are all using `0.4`.

This PR changes what `ark-ff` specifies in the Cargo.toml to 0.4, thereby keeping it consistent with the rest.

This does not closes #316. We will see whether more changes will be needed due to potential Rust changes.

---

write a little note why.

N/A:
- [ ] Wrote unit tests

### DIFF
--- a/ff/Cargo.toml
+++ b/ff/Cargo.toml
@@ -23,7 +23,7 @@ num-traits = { version = "0.2", default-features = false }
 paste = "1.0"
 rayon = { version = "1", optional = true }
 zeroize = { version = "1", default-features = false, features = ["zeroize_derive"] }
-num-bigint = { version = "0.4.0", default-features = false }
+num-bigint = { version = "0.4", default-features = false }
 
 [build-dependencies]
 rustc_version = "0.4"


vlopes11 (approved):
LGTM. Generally speaking its a good thing to use only major/minor, but we commit that `num-bigint` follows semver good practices.

I'll in parallel check their history and will come back to you in case historically they make breaking changes as patches

mmagician (approved):
Looks good to me 👍🏼

weikengchen:
Some additional notes before I merged:

1. About semver (major/minor/patch), I agree that keeping major/minor should be sufficient for our cases.

2. There seem to be no breaking changes between 0.4.0 and 0.4.2 (https://docs.rs/crate/num-bigint/0.4.2/source/RELEASES.md). The main commit that is related to the issue is https://github.com/rust-num/num-bigint/pull/219.

PR #321, opened 2021-10-04
Change bigint conversions to impl From instead of Into.

hdevalence:
## Description

This is a strictly more general bound, and means that it's possible to write
the conversion without relying on type inference.
---

write a little note why.

- ~~[ ] Wrote unit tests~~
- ~~[ ] Updated relevant documentation in the code~~

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 
 ### Features
 
+- [\#321](https://github.com/arkworks-rs/algebra/pull/321) (ark-ff) Change bigint conversions to impl `From` instead of `Into`.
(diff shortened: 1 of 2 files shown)


Pratyush (approved):
Thanks!

weikengchen:
Let me try to fix the patching script. So I can get the "test against curves" passing.

Pratyush:
@weikengchen a better way would be to temporarily add both `git` and `version` specifiers, like here: https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#multiple-locations

Pratyush:
@hdevalence I pushed a fix for the CI issue in #324, and added the corresponding `Fp` impl in #325. If you could rebase then we can get this merged (I can't rebase because of the issue with "allow edits from maintainers")

Issue #320, opened 2021-10-04
Fp256 (and other) elements don't `Display` correctly


hdevalence:
The `Display` impl for `Fp256` (and other sizes, since they're all macro-generated) prints the output in hex, so it's not possible to feed it back into the `from_str` function.

PR #322, opened 2021-10-04
Fix the Display impl for Fp* instances

hdevalence:
## Description

This change allows the output of the derived `ToString` impl to be used with
the `from_str` functions.  Previously, the `Display` impl output hex.

closes: #320 

---

write a little note why.

- ~~[ ] Wrote unit tests~~
- ~~[ ] Added a relevant changelog entry to the `Pending` section in `CHANGELOG.md`~~ (I did not do this, to avoid merge conflicts while rebasing)
- ~~[ ] Re-reviewed `Files changed` in the Github PR explorer~~

### DIFF
--- a/ff/src/fields/macros.rs
+++ b/ff/src/fields/macros.rs
@@ -607,12 +607,11 @@ macro_rules! impl_Fp {
             }
         }
 
-        /// Outputs a string containing the value of `self`, chunked up into
-        /// 64-bit limbs.
         impl<P: $FpParameters> Display for $Fp<P> {
             #[inline]
             fn fmt(&self, f: &mut Formatter<'_>) -> FmtResult {
-                write!(f, stringify!($Fp"({})"), self.into_repr())
+                let value: num_bigint::BigUint = self.into_repr().into();
+                f.write_fmt(format_args!("{}", value))
             }
         }
 
--- a/test-templates/src/fields.rs
+++ b/test-templates/src/fields.rs
@@ -230,6 +230,10 @@ pub fn from_str_test<F: PrimeField>() {
             let b = F::from(n);
 
             assert_eq!(a, b);
+            let c = F::from_str(&ark_std::format!("{}", a))
+                .map_err(|_| ())
+                .unwrap();
+            assert_eq!(a, c);
         }
     }
 


mmagician:
@hdevalence That makes sense to me. I've added a test which verifies your change - could you please allow edits on this PR and I'll just add my commit right here?
In addition we could implement `UpperHex` to still preserve the option to output hex.

hdevalence:
> @hdevalence That makes sense to me. I've added a test which verifies your change - could you please allow edits on this PR and I'll just add my commit right here?

@mmagician Oops! sorry about that .... I actually don't see the "allow edits from maintainers" box, mysteriously? [This page](https://docs.github.com/en/github/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork) says it should be a checkbox in the sidebar, but it seems to be missing.


Screen Shot 2021-10-04 at 12 24 59 PM

I'll try to figure this out but in the meantime, if you point me at the commit you made, I can cherry-pick it into the branch.

> In addition we could implement `UpperHex` to still preserve the option to output hex.

Sounds reasonable. Would you like to do that as part of this PR, or as a follow-on issue?

This is a bit bikesheddy, so I don't want to waste time on it, but would it make sense to align the `UpperHex` /  `LowerHex` impls to the canonical serialization formats (e.g., make them print the hex encoding of the canonical serialization?)

weikengchen:
I think Henry would not see the box to "allow edits from maintainers"---organization repos have more restricted permission settings.

(I do recommend submitting a PR from personal repos.)

ff/src/fields/macros.rs
```diff
@@ -607,12 +607,11 @@ macro_rules! impl_Fp {
             }
         }
 
-        /// Outputs a string containing the value of `self`, chunked up into
-        /// 64-bit limbs.
         impl<P: $FpParameters> Display for $Fp<P> {
             #[inline]
             fn fmt(&self, f: &mut Formatter<'_>) -> FmtResult {
-                write!(f, stringify!($Fp"({})"), self.into_repr())
+                let value: num_bigint::BigUint = self.into_repr().into();
```
Pratyush, line 613: I believe it should be safe to directly call `BigUint::from(self)`, as the `From`/`Into` impl already calls `into_repr()` https://github.com/arkworks-rs/algebra/blob/d7fa00e708d76465cc49ed5e33ea42ff17f599b1/ff/src/fields/macros.rs#L728

Pratyush:
> This is a bit bikesheddy, so I don't want to waste time on it, but would it make sense to align the UpperHex / LowerHex impls to the canonical serialization formats (e.g., make them print the hex encoding of the canonical serialization?)

I guess this depends on the purported use case for `UpperHex`/`LowerHex`: are they intended to be machine-friendly serializations? Or are they just human-friendly representations?

hdevalence:
> I guess this depends on the purported use case for `UpperHex`/`LowerHex`: are they intended to be machine-friendly serializations? Or are they just human-friendly representations?

IMO there's not a clear distinction here -- as a human using the library, it's convenient if the hex display is something I can decode as bytes and parse using the serializer, since it means I can use it to generate test vectors, etc.  I'm not sure what a human-friendly hex representation would be, a decimal integer is pretty human-friendly (can be pasted into Sage, is a number, etc), but usually hex data is already intended for machine interactions.

hdevalence:
> I think Henry would not see the box to "allow edits from maintainers"---organization repos have more restricted permission settings.
> 
> (I do recommend submitting a PR from personal repos.)

Ah, apparently there is no way to enable this setting for organization repos, which is deeply unfortunate -- we'd like to use our fork to accumulate changes, so that we're not pulling upstream maintainers into a development cycle for a product they're not part of.

ff/src/fields/macros.rs
hdevalence: Ah, yeah, it looks like the `into_repr()` could be skipped. Unfortunately it's not possible to call `BigUint::from`, since `Into` doesn't imply `From` (see #321, although this doesn't change the `Fp` conversions, only the `BigInteger` ones)

Pratyush:
> Ah, apparently there is no way to enable this setting for organization repos, which is deeply unfortunate -- we'd like to use our fork to accumulate changes, so that we're not asking upstream maintainers to work around our development cycle.

Can you give arkworks-rs/algebra access to the penumbra fork? Then maintainers of this repo can push to your fork

ff/src/fields/macros.rs
Pratyush: If you'd like to add the `impl From<Fp*> for BigUint` in #321 that would be great. It probably makes sense to base this PR on top of #321 anyway due to changelog conflicts

hdevalence:
> > Ah, apparently there is no way to enable this setting for organization repos, which is deeply unfortunate -- we'd like to use our fork to accumulate changes, so that we're not asking upstream maintainers to work around our development cycle.
> 
> Can you give arkworks-rs/algebra access to the penumbra fork? Then maintainers of this repo can push to your fork

I tried setting this as a team, but I wasn't able to. I can try to add people individually but the easiest thing might be to just point at commits that I can cherry-pick.

niooss-ledger:
Hello, I am interested in the feature implemented in this PR (which fixes https://github.com/arkworks-rs/algebra/issues/320) and I am wondering whether this would eventually be merged.

I understand that the function which is modified by this PR was modified very much a few days ago in `master`. It is now https://github.com/arkworks-rs/algebra/blob/08c4cdb5137591e27171f7aea2543b0602ad27d9/ff/src/fields/models/fp/mod.rs#L720-L727

What is preventing this function to be replaced to output something which can be used with `from_str`?

PR #337, opened 2021-10-21
Add `PrimeField::mul_add_assign`.

mmaker:
Introduce a new method `mul_add_assign` and `mul_add` in `PrimeField` for computing `a * b + c` using a single modular reduction. Benchmarks are attached can be run with:
```
 cargo bench ark_ff::mul_add
```
By default, this is realized calling `mul` and `add`, but for some specific fields (i.e. the ones implemented using macros) this avoids the modular reduction.
I am opening this pull request to check if it's worth addition in arkworks. Here's some benchmark results on my machine:

```
ark_ff::mul_add/mul+add_assign  time:   [34.824 ns 34.854 ns 34.886 ns]
ark_ff::mul_add/raw time:   [34.434 ns 34.458 ns 34.489 ns]      
ark_ff::mul_add/mul_add time:   [27.579 ns 27.611 ns 27.650 ns]                                                                
```

If we remove the black_box() calls from the benchmark `ark_ff::mul_add/raw`,
we get roughly the same performance as `mul_add`. It's not clear to me this is due to the compiler being extremely smart and skipping modular reduction, or if it's just avoiding the computation of the result.

---

- [x] Wrote unit tests
> I'd like to have a discussion over this first. For instance, even if this will be deemed worth adding, we might want to lift it for the trait `Field` before merging.

### DIFF
--- a/ec/src/lib.rs
+++ b/ec/src/lib.rs
@@ -247,6 +247,8 @@ pub trait AffineCurve:
     + Neg<Output = Self>
     + Zeroize
     + core::iter::Sum<Self>
+    + Mul<Self::ScalarField, Output = Self::Projective>
+    + for<'a> Mul<&'a Self::ScalarField, Output = Self::Projective>
     + for<'a> core::iter::Sum<&'a Self>
     + From<<Self as AffineCurve>::Projective>
 {
--- a/ec/src/models/short_weierstrass.rs
+++ b/ec/src/models/short_weierstrass.rs
@@ -376,6 +376,15 @@ impl<'a, P: SWCurveConfig> core::iter::Sum<&'a Self> for Affine<P> {
     }
 }
 
+impl<'a, P: SWCurveConfig, T: Borrow<P::ScalarField>> Mul<T> for Affine<P> {
+    type Output = Projective<P>;
+
+    #[inline]
+    fn mul(self, other: T) -> Self::Output {
+        self.mul_bigint(other.borrow().into_bigint())
+    }
+}
+
 /// Jacobian coordinates for a point on an elliptic curve in short Weierstrass
 /// form, over the base field `P::BaseField`. This struct implements arithmetic
 /// via the Jacobian formulae
--- a/ec/src/models/twisted_edwards.rs
+++ b/ec/src/models/twisted_edwards.rs
@@ -338,6 +338,15 @@ impl<P: TECurveConfig> Distribution<Affine<P>> for Standard {
     }
 }
 
+impl<'a, P: TECurveConfig, T: Borrow<P::ScalarField>> Mul<T> for Affine<P> {
+    type Output = Projective<P>;
+
+    #[inline]
+    fn mul(self, other: T) -> Self::Output {
+        self.mul_bigint(other.borrow().into_bigint())
+    }
+}
+
 //////////////////////////////////////////////////////////////////////////////
 
 /// `Projective` implements Extended Twisted Edwards Coordinates


ff/src/fields/arithmetic.rs
```diff
@@ -56,6 +56,70 @@ macro_rules! impl_field_mul_assign {
...
+                        r[j] = a as u64;
+                        r[j-1] = b as u64;
+                        carry1 = (a >> 64) as u64;
+                        carry2 = (b >> 64) as u64;
+
+                    }
+                    r[$limbs - 1] = carry1 + carry2;
+                }
+
+                (self.0).0 = r;
```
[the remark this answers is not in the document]
mmaker: done!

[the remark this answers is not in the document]
mmaker: done!

ff/src/fields/arithmetic.rs
```diff
@@ -56,6 +56,70 @@ macro_rules! impl_field_mul_assign {
...
+            // No-carry optimisation applied to CIOS
+            if _no_carry {
+                #[cfg(use_asm)]
+                #[allow(unsafe_code, unused_mut)]
+                {
+                    // Tentatively avoid using assembly for `$limbs == 1`.
+                    if $limbs <= 6 && $limbs > 1 {
+                        ark_ff_asm::x86_64_asm_mul!($limbs, (self.0).0, (other.0).0);
+                        self.reduce();
+                        return;
```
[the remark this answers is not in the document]
mmaker: done!

ff/src/fields/arithmetic.rs
```diff
@@ -1,14 +1,9 @@
-/// This modular multiplication algorithm uses Montgomery
-/// reduction for efficient implementation. It also additionally
-/// uses the "no-carry optimization" outlined
-/// [here](https://hackmd.io/@zkteam/modular_multiplication) if
-/// `P::MODULUS` has (a) a non-zero MSB, and (b) at least one
-/// zero bit in the rest of the modulus.
-macro_rules! impl_field_mul_assign {
+macro_rules! impl_no_carry {
     ($limbs:expr) => {
+        // XXX. This function should be made "const" in a glorious future.
```
Pratyush, line 3: Is the reason it can't be const right now because of indexing? Or because it access P::MODULUS?

ff/src/fields/arithmetic.rs
```diff
@@ -31,25 +64,42 @@ macro_rules! impl_field_mul_assign {
...
-                let mut carry2 = 0u64;
-
-                for i in 0..$limbs {
-                    r[0] = fa::mac(r[0], (self.0).0[0], (other.0).0[i], &mut carry1);
-                    let k = r[0].wrapping_mul(P::INV);
-                    fa::mac_discard(r[0], k, P::MODULUS.0[0], &mut carry2);
-                    for j in 1..$limbs {
-                        r[j] = mac_with_carry!(r[j], (self.0).0[j], (other.0).0[i], &mut carry1);
-                        r[j - 1] = mac_with_carry!(r[j], k, P::MODULUS.0[j], &mut carry2);
+                self.mul_assign_nocarry(other);
```
Pratyush, line 67: I believe the `asm` and the `mul_assign_nocarry` algorithms are the same, so we can just move the `asm` into `mul_assign_nocarry` directly.

Pratyush (changes requested):

ff/src/fields/arithmetic.rs
mmaker: ```error[E0658]: trait bounds other than `Sized` on const fn parameters are unstable```
cf. https://github.com/rust-lang/rust/issues/57563 and the feature gate `const_fn_trait_bound`

Pratyush: Ah yeah, so it's because we're inside `impl<P: FpParameters> FpXYZ<P>`. Hopefully we'll see that get stabilized, because then basically half this library becomes `const` immediately...

Pratyush: This is the relevant feature: https://github.com/rust-lang/rust/issues/67792

mmaker: Should I remove this line, keep it as a placeholder elsewhere?

Pratyush: Nah, it's fine.

PR #343, opened 2021-10-29
Add Wahby-Boneh hash-to-curve

drskalman:
## Description
This PR uses  #147 PR to implement WB hash to curve to it by essentially by translating  https://github.com/zcash/pasta_curves/blob/main/src/hashtocurve.rs to arkworks language

We have also implemented the relevant parameters in our fork of arkworks-curves for bls12-381 and bls12-377 and now it works with our bls library.

write a little note why.

- [x] Wrote unit tests


ec/src/hashing/curve_maps/wb/mod.rs
```diff
@@ -0,0 +1,90 @@
...
+{
+    //The isogenouscurve should be defined over the same basefield but it can have different scalar field
+    //type IsogenousCurveScalarField :
+    type IsogenousCurve : SWUParams::<BaseField = <Self as ModelParameters>::BaseField>;
+
+    const PHI_X_NOM: &'static [<Self::IsogenousCurve as ModelParameters>::BaseField];
+    const PHI_X_DEN: &'static [<Self::IsogenousCurve as ModelParameters>::BaseField];
+    
+    const PHI_Y_NOM: &'static [<Self::IsogenousCurve as ModelParameters>::BaseField];
+    const PHI_Y_DEN: &'static [<Self::IsogenousCurve as ModelParameters>::BaseField];
```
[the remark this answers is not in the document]
drskalman: I'm not sure what order "small" refers to. Currently the largest isogeny we have encountered is of degree 23. It is for the case of  BLS12-377 G2. This means PHI_Y could have of up to degree 46 numerator and denominator with each coefficient of 377bit size: https://github.com/w3f/arkworks-curves/blob/a22aa14c2306ee23ff9f36ed639a9a969daa157b/bls12_377/src/curves/g2.rs#L242.  However, one might consider even larger degree isogenies.

However because this evaluation carried out twice each time we hash something to the curve, any slight optimization makes a difference for protocols which need hashing lots of values to the curve. In that case, the benefit of using  `ark_poly` outweights the extra dependency concern, if it provides even a small optimization over naïve evaluation.

ec/src/hashing/curve_maps/wb/mod.rs
```diff
@@ -0,0 +1,90 @@
...
+                                           
+	let x_den : DensePolynomial<<Self::IsogenousCurve as ModelParameters>::BaseField> = <DensePolynomial<<Self::IsogenousCurve as ModelParameters>::BaseField>>::from_coefficients_slice(&Self::PHI_X_DEN[..]);
+
+	let y_num : DensePolynomial<<Self::IsogenousCurve as ModelParameters>::BaseField> = <DensePolynomial<<Self::IsogenousCurve as ModelParameters>::BaseField>>::from_coefficients_slice(&Self::PHI_Y_NOM[..]);
+
+	let y_den : DensePolynomial<<Self::IsogenousCurve as ModelParameters>::BaseField> = <DensePolynomial<<Self::IsogenousCurve as ModelParameters>::BaseField>>::from_coefficients_slice(&Self::PHI_Y_DEN[..]);
+
+	
+    let img_x = x_num.evaluate(&domain_point.x) / x_den.evaluate(&domain_point.x);
+    let img_y = (y_num.evaluate(&domain_point.x) * domain_point.y) / y_den.evaluate(&domain_point.x);
```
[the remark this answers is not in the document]
drskalman: this is done in the new commit I think.

ec/src/hashing/curve_maps/wb/mod.rs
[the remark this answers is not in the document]
Pratyush: I see, let's just use `ark-poly` then. I just wanted to avoid any circular dependency issues, but it seems that there aren't any right now.

drskalman:
I think I fixed all errors I saw in the CI. The fmt fails for `ec/src/models/` but I haven't touched those file, so I left them as they are. There was a warning for num-traits = { version = "0.2", default-features = false} about default feature not being relavant so I dropped default-features = false to get rid of the warning.

drskalman:
> There was a warning for num-traits = { version = "0.2", default-features = false} about default feature not being relavant so I dropped default-features = false to get rid of the warning.

I had to put `default-features` back otherwise `num-traits` fails the no-std test.

CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## Pending
 
+  - [\#???] Add WB and SWU hash to curve maps to hashing PR.
```
weikengchen, line 5:
```suggestion
  - [\#???] Add WB and SWU hash to curve maps to hashing PR.
  + [\#343](https://github.com/arkworks-rs/algebra/pull/343) Add WB and SWU hash-to-curve maps.
```

Pratyush:
I will try to take a look at this over the coming weekend.

daira:
I'll try to find time to review the algorithm code (I'm really busy so no guarantees).

ec/src/hashing/curve_maps/swu/mod.rs
```diff
@@ -161,10 +163,14 @@ impl<P: SWUParams> MapToCurve<GroupAffine<P>> for SWUMap<P> {
...
-        // 9. If sgn0(u) != sgn0(y), set y = -y
-        // let y = if y % 2 {-y} or {y};
+        
         let x_affine = num_x / div;
-        let y_affine = y;
+        // 9. If sgn0(u) != sgn0(y), set y = -y
+        let mut a = [0u8; 128];
+        let mut b =  [0u8; 128];
+        point.write(&mut a[..]).unwrap();
+        y.write(&mut b[..]).unwrap();
```
drskalman, line 172: @weikengchen @Pratyush I know this is inefficient, ugly and not suitable for production,  but we just left it there as a place holder of what needs to be done here, so we can ask you if  there is a better way of doing it. 

should we add a function like parity for ark_ff:field_trait and implement it for all models (as `self.c0.into_repr().is_odd()`  for example)?

drskalman: It is meant to support all curves. In particular, we need to support schemes where messages are hashed to G2 which is defined over Fp2.

Because the building of all extensions is precisely specified in both Arkworks and the Standard (the specific irreducible polynomial that should be used) one can agree on a specific basis {1,x,x^2,..} and define parity to be the parity of the coefficient of {1} basis element. 

Alternatively one can also use the parity of the norm, (I think trace is biased and is inconsistent across extensions) but this might be more expensive.

@daira has already solved this problem so we can dig into Zcash code to see their choice and implement it here. I was wondering if you think it is a good idea and if it hasn't been implemented before.

ec/src/hashing/curve_maps/swu/mod.rs
```diff
@@ -0,0 +1,181 @@
...
+
+use crate::{
+    hashing::{map_to_curve_hasher::MapToCurve, HashToCurveError},
+    models::short_weierstrass_jacobian::GroupAffine,
+    AffineCurve,
+};
+
+/// Trait defining the necessary parameters for the SWU hash-to-curve method
+/// for the curves of Weierstrass form of:
+/// y^2 = x^3 + a*x + b where ab != 0. From [WB2019]
```
Pratyush, line 17:
```suggestion
/// y^2 = x^3 + a*x + b where ab != 0. From [\[WB2019]\]
```

ec/src/hashing/curve_maps/swu/mod.rs
```diff
@@ -0,0 +1,181 @@
...
+///
+/// - [WB19] Wahby, R. S., & Boneh, D. (2019). Fast and simple constant-time
```
Pratyush, line 19:
```suggestion
/// - [\[WB2019\]] Wahby, R. S., & Boneh, D. (2019). Fast and simple constant-time
```

ec/src/hashing/curve_maps/swu/mod.rs
```diff
@@ -0,0 +1,181 @@
...
+///   hashing to the bls12-381 elliptic curve. IACR Transactions on
+///   Cryptographic Hardware and Embedded Systems, nil(nil), 154–179.
+///   http://dx.doi.org/10.46586/tches.v2019.i4.154-179
```
Pratyush, line 22:
```suggestion
/// - [\[WB19\]]: https://eprint.iacr.org/2019/403
```

ec/src/hashing/curve_maps/swu/mod.rs
```diff
@@ -0,0 +1,181 @@
...
+pub trait SWUParams: SWModelParameters {
+    // we need an element of the base field which is not a square root see [1] Sect.
+    // 4. it is also convenient to have $g(b/xi * a)$ to be square. In general
+    // we use a xi with low absolute value coefficients when they are
+    // represented as element of ZZ.
```
Pratyush, line 27:
```suggestion
    /// An element of the base field that is not a square root see \[WB2019, Section 4\]. 
    /// It is also convenient to have $g(b/xi * a)$ to be square. In general
    /// we use a `XI` with low absolute value coefficients when they are
    /// represented as integers.
```

ec/src/hashing/curve_maps/swu/mod.rs
```diff
@@ -0,0 +1,181 @@
...
+    const XI: Self::BaseField; // a nonsquare in Fq
+    const ZETA: Self::BaseField; // arbitatry root of unity
+    const XI_ON_ZETA_SQRT: Self::BaseField; // square root of THETA
```
Pratyush, line 30: Q: Is the comment about `THETA` correct?

```suggestion
    const XI: Self::BaseField;
    /// An arbitrary root of unity in the base field.
    const ZETA: Self::BaseField;
    /// Square root of `THETA = Self::XI/Self::ZETA`.
    const XI_ON_ZETA_SQRT: Self::BaseField;
```

ec/src/hashing/curve_maps/swu/mod.rs
```diff
@@ -0,0 +1,181 @@
...
+}
+
+pub struct SWUMap<P: SWUParams> {
```
Pratyush, line 33:
```suggestion
/// Represents the SWU hash-to-curve map defined by `P`.
pub struct SWUMap<P: SWUParams> {
```

ec/src/hashing/curve_maps/swu/mod.rs
```diff
@@ -0,0 +1,181 @@
...
+    pub domain: Vec<u8>,
+    curve_params: PhantomData<fn() -> P>,
+}
+
+impl<P: SWUParams> MapToCurve<GroupAffine<P>> for SWUMap<P> {
+    /// This is to verify if the provided SWUparams makes sense, doesn't do much
+    /// for now
```
Pratyush, line 40:
```suggestion
    /// Constructs a new map if `P` represents a valid map.
```

ec/src/hashing/curve_maps/swu/mod.rs
```diff
@@ -0,0 +1,181 @@
...
+
+        Ok(SWUMap {
+            domain: domain.to_vec(),
+            curve_params: PhantomData,
+        })
+    }
+
+    /// Map random field point to a random curve point
+    /// inspired from
+    /// https://github.com/zcash/pasta_curves/blob/main/src/hashtocurve.rs
```
Pratyush, line 79:
```suggestion
    /// Map an arbitrary base field element to a curve point.
    /// Based on
    /// <https://github.com/zcash/pasta_curves/blob/main/src/hashtocurve.rs>.
```

ec/src/hashing/curve_maps/swu/mod.rs
```diff
@@ -0,0 +1,181 @@
...
+    fn map_to_curve(
+        &self,
+        point: <GroupAffine<P> as AffineCurve>::BaseField,
```
Pratyush, line 82:
```suggestion
        point: P::BaseField,
```

ec/src/hashing/curve_maps/swu/mod.rs
Pratyush: I see. Let's add a method to the `Field` trait that returns an array of the underlying prime field elements then? So something like

```rust
pub trait Field {
	type BasePrimeFieldArr: AsRef<[Self::BasePrimeField]>
	fn to_base_prime_field_elements(&self) -> Self::BasePrimeFieldArr;
}
```

And then you can add a `parity` helper function here:

```rust
fn parity<F: Field>(element: &F) -> bool {
	element.to_base_prime_field_elements().as_ref()[0].into_repr().is_odd()
}
```

ec/src/hashing/curve_maps/wb/mod.rs
```diff
@@ -0,0 +1,102 @@
...
+
+/// Trait defining the necessary parameters for the WB hash-to-curve method
+/// for the curves of Weierstrass form of:
+/// of y^2 = x^3 + a*x + b where b != 0 but `a` can be zero like BLS-381 curve.
+/// From [WB2019]
+///
+/// - [WB19] Wahby, R. S., & Boneh, D. (2019). Fast and simple constant-time
+///   hashing to the bls12-381 elliptic curve. IACR Transactions on
+///   Cryptographic Hardware and Embedded Systems, nil(nil), 154–179.
+///   http://dx.doi.org/10.46586/tches.v2019.i4.154-179
```
Pratyush, line 23: Same change as above.

ec/src/hashing/curve_maps/wb/mod.rs
```diff
@@ -0,0 +1,102 @@
...
+
+pub struct WBMap<P: WBParams> {
+    pub domain: Vec<u8>,
+    swu_field_curve_hasher: SWUMap<P::IsogenousCurve>,
+    curve_params: PhantomData<fn() -> P>,
+}
+
+impl<P: WBParams> MapToCurve<GroupAffine<P>> for WBMap<P> {
+    /// This is to verify if the provided WBparams makes sense, doesn't do much
+    /// for now
```
Pratyush, line 66:
```suggestion
    /// Constructs a new map if `P` represents a valid map.
```

ec/src/hashing/field_hashers/mod.rs
```diff
@@ -0,0 +1,103 @@
...
+    len_in_bytes as usize
+}
+
+fn map_bytes_to_field_elem<F: Field>(bz: &[u8]) -> Option<F> {
+    let d: usize = F::extension_degree() as usize;
+    let len_per_elem = bz.len() / d;
+    let mut base_field_elems = Vec::new();
+    for i in 0..d {
+        let val = F::BasePrimeField::from_be_bytes_mod_order(
+            &bz[i * len_per_elem..(i + 1) * len_per_elem],
```
Pratyush, line 31:
```suggestion
            &bz[(i * len_per_elem)..((i + 1) * len_per_elem)],
```

ec/src/hashing/field_hashers/mod.rs
```diff
@@ -0,0 +1,103 @@
...
+
+use crate::hashing::{map_to_curve_hasher::*, *};
+use ark_std::string::ToString;
+use digest::{Update, VariableOutput};
+
+// This function computes the length in bytes that a hash function should output
+// for hashing `count` field elements.
+// See section 5.1 and 5.3 of the
+// [IETF hash standardization draft](https://tools.ietf.org/html/draft-irtf-cfrg-hash-to-curve-10)
+fn hash_len_in_bytes<F: Field>(security_parameter: usize, count: usize) -> usize {
```
Pratyush, line 12:
```suggestion
fn hash_len_in_bytes<F: Field, const SEC_PARAM: usize>(count: usize) -> usize {
```

ec/src/hashing/field_hashers/mod.rs
```diff
@@ -0,0 +1,103 @@
...
+    // ceil(log(p))
+    let base_field_size_in_bits = F::BasePrimeField::size_in_bits();
+    // ceil(log(p)) + security_parameter
+    let base_field_size_with_security_padding_in_bits =
+        base_field_size_in_bits + security_parameter;
```
Pratyush, line 17:
```suggestion
        base_field_size_in_bits + SEC_PARAM;
```

ec/src/hashing/field_hashers/mod.rs
```diff
@@ -0,0 +1,103 @@
...
+}
+
+// Implement HashToField from F and a variable output hash
+impl<F: Field, H: VariableOutput + Update + Sized + Clone> HashToField<F>
+    for DefaultFieldHasher<H>
+{
+    fn new_hash_to_field(domain: &[u8], count: usize) -> Result<Self, HashToCurveError> {
+        // Hardcode security parameter
+        let security_parameter = 128;
+        let bytes_per_base_field_elem = hash_len_in_bytes::<F>(security_parameter, count);
```
Pratyush, line 55:
```suggestion
        let bytes_per_base_field_elem = hash_len_in_bytes::<F, 128>(security_parameter, count);
```

ec/src/hashing/field_hashers/mod.rs
```diff
@@ -0,0 +1,103 @@
...
+        // Clone the hasher, and hash our message
+        let mut cur_hasher = self.domain_seperated_hasher.clone();
+        cur_hasher.update(msg);
+        let mut hashed_bytes = Vec::new();
+        cur_hasher.finalize_variable(|res| hashed_bytes.extend_from_slice(res));
+        // Now separate the hashed bytes according to each field element.
+        let mut result = Vec::with_capacity(self.count);
+        let len_per_elem = hashed_bytes.len() / self.count;
+        for i in 0..self.count {
+            let bz_per_elem = &hashed_bytes[i * len_per_elem..(i + 1) * len_per_elem];
```
Pratyush, line 96:
```suggestion
            let bz_per_elem = &hashed_bytes[i * len_per_elem..][..len_per_elem];
```

ec/src/hashing/map_to_curve_hasher.rs
```diff
@@ -0,0 +1,76 @@
...
+use ark_std::{marker::PhantomData, vec::Vec};
+
+/// Trait for mapping a random field element to a random curve point.
+pub trait MapToCurve<T: AffineCurve> {
+    fn new_map_to_curve(domain: &[u8]) -> Result<Self, HashToCurveError>
+    where
+        Self: Sized;
+    /// Map random field point to a random curve point
+    fn map_to_curve(&self, point: T::BaseField) -> Result<T, HashToCurveError>;
+}
```
Pratyush, line 12:
```suggestion
pub trait MapToCurve<T: AffineCurve> {
	/// Constructs a new mapping.
    fn new_map_to_curve(domain: &[u8]) -> Result<Self, HashToCurveError>
    where
        Self: Sized;
    /// Maps an arbitrary field element to a corresponding curve point.
    fn map_to_curve(&self, point: T::BaseField) -> Result<T, HashToCurveError>;
}
```

ec/src/hashing/map_to_curve_hasher.rs
```diff
@@ -0,0 +1,76 @@
...
+
+/// Trait for hashing messages to field elements
+pub trait HashToField<F: Field>: Sized {
+    fn new_hash_to_field(domain: &[u8], count: usize) -> Result<Self, HashToCurveError>;
+
+    fn hash_to_field(&self, msg: &[u8]) -> Result<Vec<F>, HashToCurveError>;
```
Pratyush, line 18: Let's add similar comments for these.

ec/src/hashing/map_to_curve_hasher.rs
```diff
@@ -0,0 +1,76 @@
...
+    _params_t: PhantomData<T>,
+}
+
+impl<T, H2F, M2C> HashToCurve<T> for MapToCurveBasedHasher<T, H2F, M2C>
+where
+    T: AffineCurve,
+    H2F: HashToField<T::BaseField>,
+    M2C: MapToCurve<T>,
+{
+    fn new(domain: &[u8]) -> Result<Self, HashToCurveError> {
```
Pratyush, line 41: Can the domain be dynamic, or is it generally statically known? I.e. do we sometimes need to set it at run-time? I can imagine the answer being yes sometimes, e.g. `for ii in 0..100, HashToCurve("domain {i}")`.

However if not, then maybe we can rely on constants instead of needing to construct a new `Vec` for the struct.

Pratyush:
Thanks for this PR, it's nicely self-contained and pretty clean overall =). Left an initial round of comments.

ec/src/hashing/curve_maps/swu/mod.rs
daira, line 19:
```suggestion
/// - [WB2019] Wahby, R. S., & Boneh, D. (2019). Fast and simple constant-time
```

ec/src/hashing/curve_maps/swu/mod.rs
daira:
```suggestion
/// - [\[WB2019\]] <https://eprint.iacr.org/2019/403>
```

ec/src/hashing/curve_maps/swu/mod.rs
```diff
@@ -0,0 +1,181 @@
...
+}
+
+impl<P: SWUParams> MapToCurve<GroupAffine<P>> for SWUMap<P> {
+    /// This is to verify if the provided SWUparams makes sense, doesn't do much
+    /// for now
+    fn new_map_to_curve(domain: &[u8]) -> Result<Self, HashToCurveError> {
+        // Verifying that both XI and ZETA are non-squares
+        if P::XI.legendre().is_qr() || P::ZETA.legendre().is_qr() {
+            return Err(HashToCurveError::MapToCurveError(
+                "both Xi and Zeta should be quadratic non-residues for the SWU map".to_string(),
```
daira, line 45:
```suggestion
                "both xi and zeta should be quadratic non-residues for the SWU map".to_string(),
```
(lowercase ξ and ζ are different to uppercase Ξ and Ζ).

ec/src/hashing/curve_maps/swu/mod.rs
```diff
@@ -0,0 +1,181 @@
...
+            ));
+        }
+
+        // Verifying precomupted values
+        let xi_on_zeta = P::XI / P::ZETA;
+        match xi_on_zeta.sqrt() {
+            Some(xi_on_zeta_sqrt) => {
+                if xi_on_zeta_sqrt != P::XI_ON_ZETA_SQRT && xi_on_zeta_sqrt != -P::XI_ON_ZETA_SQRT {
+                    return Err(HashToCurveError::MapToCurveError(
+                        "precomupted P::XI_ON_ZETA_SQRT is not what it suppose to be".to_string(),
```
daira, line 55:
```suggestion
                        "precomupted P::XI_ON_ZETA_SQRT is not what it is supposed to be".to_string(),
```

ec/src/hashing/curve_maps/swu/mod.rs
```diff
@@ -0,0 +1,181 @@
...
+        //   Z * u^2    ξ * t^2 (called u, confusingly)
+        //   x1         X_0(t)
+        //   x2         X_1(t)
+        //   gx1        g(X_0(t))
+        //   gx2        g(X_1(t))
+        //
+        // Using the "here" names:
+        //    x1 = num_x1/div      = [B*(Z^2 * u^4 + Z * u^2 + 1)] / [-A*(Z^2 * u^4 + Z
+        // * u^2]   gx1 = num_gx1/div_gx1 = [num_x1^3 + A * num_x1 * div^2 + B *
+        // div^3] / div^3
```
daira, line 105:
```suggestion
        //    x1 = num_x1/div      = [B*(Z^2 * u^4 + Z * u^2 + 1)] / [-A*(Z^2 * u^4 + Z * u^2]
        //   gx1 = num_gx1/div_gx1 = [num_x1^3 + A * num_x1 * div^2 + B * div^3] / div^3
```

ec/src/hashing/curve_maps/swu/mod.rs
daira: In `pasta_curves` we use an `is_odd` method on field elements which is implemented (e.g. for the Pallas base field) [here](https://github.com/zcash/pasta_curves/blob/c052756831fd4721fcf6ebeb3f2bf83ee80a5cda/src/fields/fp.rs#L592-L608), which is equivalent to what @Pratyush suggests.

ec/src/hashing/curve_maps/wb/mod.rs
```diff
@@ -0,0 +1,102 @@
...
+/// for the curves of Weierstrass form of:
+/// of y^2 = x^3 + a*x + b where b != 0 but `a` can be zero like BLS-381 curve.
+/// From [WB2019]
+///
+/// - [WB19] Wahby, R. S., & Boneh, D. (2019). Fast and simple constant-time
+///   hashing to the bls12-381 elliptic curve. IACR Transactions on
+///   Cryptographic Hardware and Embedded Systems, nil(nil), 154–179.
+///   http://dx.doi.org/10.46586/tches.v2019.i4.154-179
+pub trait WBParams: SWModelParameters + Sized {
+    // The isogenouscurve should be defined over the same basefield but it can have
```
daira, line 25:
```suggestion
    // The isogenous curve should be defined over the same base field but it can have
```

ec/src/hashing/README.md
```diff
@@ -0,0 +1,3 @@
+# Hash to Curve
+
+This folder provides a trait for `HashToCurve` and a number of implementations of hashing to the curve.
```
Pratyush, line 3: We should move this description to the module documentation.

ec/src/lib.rs
```diff
@@ -44,6 +44,7 @@ pub mod group;
 
 pub mod msm;
 
+pub mod hashing;
```
Pratyush, line 47:
```suggestion
/// Provides a `HashToCurve` trait and implementations of this trait via
/// different hashing strategies.
pub mod hashing;
```

ec/src/hashing/curve_maps/swu/mod.rs
simonmasson: It seems to be `t^3` in the notations of [WB19, Section 4.2]. I don't understand what means "An arbitrary root of unity in the base field".

simonmasson: From the file `ec/src/hasing/curve_maps/swu/mod.rs` line 149, I guess `ZETA` is a fixed fixed non-square ?

drskalman: a primitive root of unity I think is any generator of the multiplicative group and is a non-sequare. I think any non-square will do here. But this is coming from @daira 's code and I think it is chosen because there was already a constant time function which was using a pre defined root of unity to carry out this operation: https://github.com/arkworks-rs/algebra/pull/343/files#diff-b13fb8649e8438bfe74b4b6f00cf6e78debf4df9bc5fae8d56775e3ee15966adR128

We could use any other non-square including the one suggested in the paper  but I stuck to hir implementation so I'm able to run it on also pasta and make sure I get the same point as the one from @daira so I know the implementation is correct.

ec/src/hashing/curve_maps/wb/mod.rs
```diff
@@ -0,0 +1,101 @@
...
+        let x_den : DensePolynomial<<Self::IsogenousCurve as ModelParameters>::BaseField> = <DensePolynomial<<Self::IsogenousCurve as ModelParameters>::BaseField>>::from_coefficients_slice(&Self::PHI_X_DEN[..]);
+
+        let y_num : DensePolynomial<<Self::IsogenousCurve as ModelParameters>::BaseField> = <DensePolynomial<<Self::IsogenousCurve as ModelParameters>::BaseField>>::from_coefficients_slice(&Self::PHI_Y_NOM[..]);
+
+        let y_den : DensePolynomial<<Self::IsogenousCurve as ModelParameters>::BaseField> = <DensePolynomial<<Self::IsogenousCurve as ModelParameters>::BaseField>>::from_coefficients_slice(&Self::PHI_Y_DEN[..]);
+
+        let mut v: [<Self as ModelParameters>::BaseField; 2] = [
+            x_den.evaluate(&domain_point.x),
+            y_den.evaluate(&domain_point.x),
+        ];
```
Pratyush, line 49:
```suggestion
type BaseField<MP> = <MP as ModelParameters>::BaseField;

/// Trait defining the necessary parameters for the WB hash-to-curve method
/// for the curves of Weierstrass form of:
/// of y^2 = x^3 + a*x + b where b != 0 but `a` can be zero like BLS-381 curve.
/// From [WB2019]
///
/// - [WB19] Wahby, R. S., & Boneh, D. (2019). Fast and simple constant-time
///   hashing to the bls12-381 elliptic curve. IACR Transactions on
///   Cryptographic Hardware and Embedded Systems, nil(nil), 154–179.
///   http://dx.doi.org/10.46586/tches.v2019.i4.154-179
pub trait WBParams: SWModelParameters + Sized {
    // The isogenous curve should be defined over the same base field but it can have
    // different scalar field type IsogenousCurveScalarField :
    type IsogenousCurve: SWUParams<BaseField = BaseField<Self>>;

    const PHI_X_NOM: &'static [BaseField<Self>];
    const PHI_X_DEN: &'static [BaseField<Self>];

    const PHI_Y_NOM: &'static [BaseField<Self>];
    const PHI_Y_DEN: &'static [BaseField<Self>];

    fn isogeny_map(
        domain_point: GroupAffine<Self::IsogenousCurve>,
    ) -> Result<GroupAffine<Self>, HashToCurveError> {
        let x_num = DensePolynomial::from_coefficients_slice(&Self::PHI_X_NOM[..]);
        let x_den = DensePolynomial::from_coefficients_slice(&Self::PHI_X_DEN[..]);

        let y_num = DensePolynomial::from_coefficients_slice(&Self::PHI_Y_NOM[..]);
        let y_den = DensePolynomial::from_coefficients_slice(&Self::PHI_Y_DEN[..]);

        let mut v: [BaseField<Self>; 2] = [
            x_den.evaluate(&domain_point.x),
            y_den.evaluate(&domain_point.x),
        ];
```

ec/src/hashing/curve_maps/wb/mod.rs
```diff
@@ -0,0 +1,101 @@
...
+use ark_poly::{univariate::DensePolynomial, Polynomial, UVPolynomial};
+
+use crate::{
+    hashing::{map_to_curve_hasher::MapToCurve, HashToCurveError},
+    models::short_weierstrass_jacobian::GroupAffine,
+    AffineCurve,
+};
+
+use super::swu::{SWUMap, SWUParams};
+
```
Pratyush, line 14:
```suggestion
type BaseField<MP> = <MP as ModelParameters>::BaseField;

```

ec/src/hashing/curve_maps/wb/mod.rs
```diff
@@ -0,0 +1,101 @@
...
+        let mut v: [<Self as ModelParameters>::BaseField; 2] = [
+            x_den.evaluate(&domain_point.x),
+            y_den.evaluate(&domain_point.x),
+        ];
+        batch_inversion(&mut v);
+        let img_x = x_num.evaluate(&domain_point.x) * v[0];
+        let img_y = (y_num.evaluate(&domain_point.x) * domain_point.y) * v[1];
+
+        Ok(GroupAffine::<Self>::new(img_x, img_y, false))
```
Pratyush, line 54:
```suggestion
/// Trait defining the necessary parameters for the WB hash-to-curve method
/// for the curves of Weierstrass form of:
/// of y^2 = x^3 + a*x + b where b != 0 but `a` can be zero like BLS-381 curve.
/// From [WB2019]
///
/// - [WB19] Wahby, R. S., & Boneh, D. (2019). Fast and simple constant-time
///   hashing to the bls12-381 elliptic curve. IACR Transactions on
///   Cryptographic Hardware and Embedded Systems, nil(nil), 154–179.
///   http://dx.doi.org/10.46586/tches.v2019.i4.154-179
pub trait WBParams: SWModelParameters + Sized {
    // The isogenous curve should be defined over the same base field but it can have
    // different scalar field type IsogenousCurveScalarField :
    type IsogenousCurve: SWUParams<BaseField = BaseField<Self>>;

    const PHI_X_NOM: &'static [BaseField<Self>];
    const PHI_X_DEN: &'static [BaseField<Self>];

    const PHI_Y_NOM: &'static [BaseField<Self>];
    const PHI_Y_DEN: &'static [BaseField<Self>];

    fn isogeny_map(
        domain_point: GroupAffine<Self::IsogenousCurve>,
    ) -> Result<GroupAffine<Self>, HashToCurveError> {
        let x_num = DensePolynomial::from_coefficients_slice(&Self::PHI_X_NOM[..]);
        let x_den = DensePolynomial::from_coefficients_slice(&Self::PHI_X_DEN[..]);

        let y_num = DensePolynomial::from_coefficients_slice(&Self::PHI_Y_NOM[..]);
        let y_den = DensePolynomial::from_coefficients_slice(&Self::PHI_Y_DEN[..]);

        let mut v: [BaseField<Self>; 2] = [
            x_den.evaluate(&domain_point.x),
            y_den.evaluate(&domain_point.x),
        ];
        batch_inversion(&mut v);
        let img_x = x_num.evaluate(&domain_point.x) * v[0];
        let img_y = (y_num.evaluate(&domain_point.x) * domain_point.y) * v[1];

        Ok(GroupAffine::new(img_x, img_y, false))
```

ec/src/hashing/map_to_curve_hasher.rs
drskalman: We can't think of an application that the domain needs to be dynamic. It has been the structure that @ValarDragon had put in place. We can go ahead and replace it by a constant and hand it to the field hasher as const generic, if that is what you are suggesting.

Another somehow related change to @ValarDragon structure, is that MapToCurve does not use the domain to decide which element should be mapped to which point on the curve, so @mmagician removed domian completely from MapToCurve in this [pull request](https://github.com/w3f/arkworks-algebra/pull/23). Could you verify that this make sense and that you have no objection to this?

ec/src/hashing/curve_maps/swu/mod.rs
drskalman: @Pratyush could you tell me why do you prefer `fn parity<F: Field>(element: &F) -> bool` over simply adding a member method `fn parity(&self) -> bool` to Field trait?

Pratyush: The reason is that `parity` isn't really a general operation that's defined for arbitrary fields, and is only really used in this hash-to-curve algorithm. So IMO it's not worth it to complicate the `Field` API with an additional method. On the other hand, adding an API of `BasePrimeFieldIter` makes sense, because you can always view a field element as a list of base prime field elements.

ec/src/hashing/map_to_curve_hasher.rs
Pratyush: I approved the changes in that PR; the design looks good to me.

ec/src/hashing/field_hashers/mod.rs
```diff
@@ -51,8 +49,7 @@ impl<F: Field, H: VariableOutput + Update + Sized + Clone> HashToField<F>
 {
     fn new_hash_to_field(domain: &[u8], count: usize) -> Result<Self, HashToCurveError> {
         // Hardcode security parameter
-        let security_parameter = 128;
-        let bytes_per_base_field_elem = hash_len_in_bytes::<F>(security_parameter, count);
+        let bytes_per_base_field_elem = hash_len_in_bytes::<F, 128>(count);
```
daira, line 52: Note that this won't be compatible with the `pasta_curves` implementation; we use 256.

Here's the argument for why:
* There is no performance loss from doing so, at least when using XMD:BLAKE2b.
* If you're aiming for a 128-bit security level, and being pedantic about provable concrete security, then all of the ~2<sup>-128</sub> success probabilities for the attacker add up and you have to take them into account for the final security bound. Whereas any 2<sup>-256</sub> success probabilities are still completely negligible and don't matter.

drskalman: We are going to add different FieldHasher(s) and work on compatibility with different network on our next iteration. I'd leave it as it was from the original pull request for now to facilitate merging. Unless @Pratyush would prefer to change it for the DefaultFieldHasher.

ec/src/hashing/curve_maps/swu/mod.rs
```diff
@@ -0,0 +1,221 @@
...
+trait ElementParity<F: Field> {
+    fn parity(element: &F) -> bool {
+        element
+            .to_base_prime_field_elements()
+            .next()
+            .unwrap()
+            .into_repr()
+            .is_odd()
+    }
+}
```
Pratyush, line 43: We can add a blanket implementation for all fields here, so that we don't need to add manual impls downstream.

Pratyush: Also, is this the same parity as that suggested in the IETF draft?

ec/src/hashing/curve_maps/swu/mod.rs
```diff
@@ -0,0 +1,221 @@
...
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use ark_test_curves::bls12_381::{Fq, Fq2, Fq6};
+
+    impl ElementParity<Fq> for Fq {}
+    impl ElementParity<Fq2> for Fq2 {}
+    impl ElementParity<Fq6> for Fq6 {}
```
Pratyush, line 188: Hmm these impls seem unnecessary; does `SWUMap::parity` not suffice for this?

ec/src/hashing/curve_maps/wb/mod.rs
```diff
@@ -0,0 +1,95 @@
...
+
+        Ok(WBMap {
+            swu_field_curve_hasher: SWUMap::<P::IsogenousCurve>::new_map_to_curve().unwrap(),
+            curve_params: PhantomData,
+        })
+    }
+
+    /// Map random field point to a random curve point
+    /// inspired from
+    /// https://github.com/zcash/pasta_curves/blob/main/src/hashtocurve.rs
```
Pratyush, line 86:
```suggestion
    /// <https://github.com/zcash/pasta_curves/blob/main/src/hashtocurve.rs>
```

ec/src/hashing/field_hashers/mod.rs
Pratyush: One middle ground would be to expose the security parameter as a const parameter on `DefaultFieldHasher`?

ff/src/fields/mod.rs
```diff
@@ -242,7 +245,7 @@ pub trait FpParameters: FftParameters {
     /// the representation when randomly sampling.
     const REPR_SHAVE_BITS: u32;
 
-    /// Let `M` be the power of 2^64 nearest to `Self::MODULUS_BITS`. Then
+    /// Let `M` be the nearest greater power of 2^64 to `Self::MODULUS_BITS`. Then
```
Pratyush, line 248:
```suggestion
    /// Let `M` be the nearest power of 2^64 greater than `Self::MODULUS_BITS`. Then
```

Pratyush (commented):
This looks almost ready, modulo the small nits.

ec/Cargo.toml
```diff
@@ -17,11 +17,18 @@ rust-version = "1.56"
 ark-std = { version = "^0.3.0", default-features = false }
 ark-serialize = { version = "^0.3.0", path = "../serialize", default-features = false }
 ark-ff = { version = "^0.3.0", path = "../ff", default-features = false }
+ark-poly = { version = "^0.3.0", path = "../poly", default-features = false }
 derivative = { version = "2", features = ["use_core"] }
+digest = { version = "0.9", default-features = false }
```
Pratyush, line 22:
```suggestion
digest = { version = "0.10", default-features = false }
```

ec/src/hashing/curve_maps/swu/mod.rs
mmagician: I am not sure how this would work - `SWUMap` is defined over `SWUParams`, and I think we would need to implement the parameters for each base field we want to test, like it is done [here](https://github.com/w3f/arkworks-algebra/blob/hash_to_curve_wb/ec/src/hashing/tests.rs#L115). That would bloat the test code in the swu module.

I agree that the above 3 lines look a little unusual, but I think it's a neat hack that leverages the fact that the `ElementParity` trait is not tied to SWUMap, and that luckily in this scope the trait is local, so we don't have to worry about orphan traits.

Correct me if I'm wrong, I'd happily integrate another solution.

ec/src/hashing/curve_maps/swu/mod.rs
drskalman: I changed it to match the IETF draft definition here: https://github.com/w3f/arkworks-algebra/pull/33

Pratyush:
Hi all, is there anything I can do to help push this along?

drskalman:
> Hi all, is there anything I can do to help push this along?

I think it requires your judgement on @mmagician comment here:

https://github.com/arkworks-rs/algebra/pull/343/files/e18fd44106008cd3c885cf955556f805f24db9bc#diff-b13fb8649e8438bfe74b4b6f00cf6e78debf4df9bc5fae8d56775e3ee15966adR188

ec/src/hashing/curve_maps/swu/mod.rs
Pratyush: I guess I don't fully see the necessity of defining a separate trait for this, especially if we might run into orphan impl conflicts somehow.

ec/src/hashing/field_hashers/mod.rs
mmagician: @Pratyush Done in https://github.com/w3f/arkworks-algebra/pull/35/commits/15677da9468bba5fe689b017c189197f893280bf

ec/src/hashing/curve_maps/swu/mod.rs
mmagician: Done in @https://github.com/w3f/arkworks-algebra/pull/35/commits/80082c447e8d183711972b8bbc01db0c53fb3b63

mmagician:
With the latest commits I have addressed the outstanding comments. Let us know if there's anything else @Pratyush.

ec/src/hashing/curve_maps/swu/mod.rs
```diff
@@ -14,12 +14,12 @@ use crate::{
...
 /// - [\[WB2019\]] <https://eprint.iacr.org/2019/403>
 pub trait SWUParams: SWModelParameters {
-    /// An element of the base field that is not a square root see \[WB2019, Section 4\]. 
+    /// An element of the base field that is not a square root see \[WB2019, Section 4\].
     /// It is also convenient to have $g(b/xi * a)$ to be square. In general
     /// we use a `XI` with low absolute value coefficients when they are
     /// represented as integers.
     const XI: Self::BaseField;
-    /// An arbitrary root of unity in the base field.
+    /// An arbitrary nonsquare conviniently chosen to be a primitve element of the base field
```
daira, line 22:
```suggestion
    /// An arbitrary nonsquare conveniently chosen to be a primitive element of the base field
```

ec/src/hashing/curve_maps/wb/mod.rs
```diff
@@ -16,12 +16,9 @@ type BaseField<MP> = <MP as ModelParameters>::BaseField;
...
 /// for the curves of Weierstrass form of:
 /// of y^2 = x^3 + a*x + b where b != 0 but `a` can be zero like BLS-381 curve.
-/// From [WB2019]
+/// From [\[WB2019\]]
 ///
-/// - [WB19] Wahby, R. S., & Boneh, D. (2019). Fast and simple constant-time
-///   hashing to the bls12-381 elliptic curve. IACR Transactions on
-///   Cryptographic Hardware and Embedded Systems, nil(nil), 154–179.
-///   http://dx.doi.org/10.46586/tches.v2019.i4.154-179
+/// - [\[WB19\]] <http://dx.doi.org/10.46586/tches.v2019.i4.154-179>
```
daira, line 21:
```suggestion
/// - [\[WB2019\]] <http://dx.doi.org/10.46586/tches.v2019.i4.154-179>
```

daira (approved):
Do you have any test vectors comparing against the Pasta implementation of hash-to-curve? I didn't see any (but this is a large PR and I might have missed them).

Pratyush (approved):

Issue #330, opened 2021-10-16
What is the rationale for Edwards curve serialisation?


kevaundray:
In the arkworks library, the default encoding format for a point on a Twisted Edwards curve is to take the `x` co-ordinate and the sign of the `y` co-ordinate: https://github.com/arkworks-rs/algebra/blob/master/ec/src/models/twisted_edwards_extended.rs#L708

This is opposed to the format specified here: https://datatracker.ietf.org/doc/html/rfc8032#section-3.1 , where we take the `y` co-ordinate and the sign of the `x` co-ordinate. Can you explain the rationale for this not being the default? I'd assume that if no special encoding is specified in the specification, that other curve developers would use the one specified in the RFC linked. If this is the case, then it may make interoperability a problem in the future.

Two crates which use this encoding are:

- https://github.com/dalek-cryptography/curve25519-dalek/blob/main/src/edwards.rs#L518
- https://github.com/zkcrypto/jubjub/blob/main/src/lib.rs#L457

**Note**

This is in contrast to the Weierstrass curve, where I usually see the `x` co-ordinate and the sign of the `y` co-ordinate being taken.

PR #345, opened 2021-11-02
Change serialisation format for Twisted Edwards Curves

kevaundray:
## Description

This changes the way that we serialise and deserialize Twisted Edwards points.

closes: #330

---

write a little note why.

- [ ] Wrote unit tests


kevaundray:
Minor note:

There is a method called `from_random_bytes` however there is no guarantee that the bytes will be _random_ looking from the user and it is a public method.

The documentation for that method states:

```rust
 /// Returns a group element if the set of bytes forms a valid group element,
 /// otherwise returns None. This function is primarily intended for sampling
 /// random group elements from a hash-function or RNG output.
```

I believe that this method does not reduce because you don't want to introduce any bias onto the bytes produced by the hash function or RNG output, so maybe a better name would be `from_bytes_no_reduce`. The problem with this suggestion is that users may not be aware that they do not want to introduce bias and may call `from_be_bytes_mod_order` on the RNG output instead.

Another suggestion is to split this into two methods:

**hash_from_bytes**

```rust
 // User passes in their non-random looking bytes and it is 
// hashed using their chosen hash function, then mapped to a point
// using some hash_to_curve method
//
// Could return an Option<Point> here also to mimic current behaviour.
 pub fn hash_from_bytes<HashTrait>(bytes: &[u8]) -> Point;
 ```

**from_rng**
```rust
// User passes in a PRNG and the number of times to call the PRNG in order to 
// derive a valid point.
pub fn from_prng<PRNGTrait>(tries : Option<u32>) -> Option<Point>
```
Another suggestion is to do nothing which is also completely valid as I've not seen anyone open an issue regarding `from_random_bytes` or the API surrounding it.

drskalman:
I second, the second proposal (and I had issue with "from_random_bytes" but I didn't know if I should open an issue). I can't see any use for "from_random_bytes" which is not addressed by either of proposed functions "hash_from_bytes" and "hash_rng". Otherwise it seems useless on its own (without for example, being included in a loop till it succeed which is basically what "hash_rng" is supposed to do).

I like the idea of replacing it with those two functions or at least not exposing it to the user.

ec/src/models/twisted_edwards_extended.rs
```diff
@@ -836,3 +842,65 @@ where
...
+            // Serialize 0.
+            P::BaseField::zero().serialize_with_flags(writer, flags)
+        } else {
+            // Note: although this says `from_x_sign` and we are
+            // using the sign of `y`. The logic works the same.
+            let flags = EdwardsFlags::from_x_sign(self.y > -self.y);
+            self.x.serialize_with_flags(writer, flags)
+        }
+    }
+    pub fn deserialize_old<R: Read>(mut reader: R) -> Result<Self, SerializationError> {
```
Pratyush, line 881: Could you also add the uncompressed versions of these? Thanks!

Pratyush (commented):
Thanks for this! It looks good modulo the comment about `serialize_uncompressed_old`.

ec/src/models/twisted_edwards_extended.rs
kevaundray: Yep :) I left it out initially as it doesn't change in the new serialisation format

kevaundray:
I was going to double check that this new format does match the specified format that we usually do for Edwards Curves, in particular the masks are the correct values used here: https://datatracker.ietf.org/doc/html/rfc8032#section-3.1

kevaundray:
Should this line say y_sign: https://github.com/arkworks-rs/algebra/blob/master/serialize/src/flags.rs#L126 ?

I can't add a review comment as it's not been edited

kevaundray:
Summarising the changes:

For Twisted Edwards:
-  We now encode the Y coordinate and take the sign bit of the X co-ordinate
- The default flag is now the positive co-ordinate, following https://datatracker.ietf.org/doc/html/rfc8032#section-3.1 

      >  y concatenated with one bit that is 1 if x is negative and 0 if x is not negative.

Pratyush:
Sounds good, can you add a note in the CHANGELOG with the summary? And re: the `x_sign` in `SWFlags`, yeah, that looks like it should be `y_sign`.

Also, if you could undo the formatting changes in unrelated files, that would be great. I think I should just remove `wrap_comments` from `rustfmt.toml`

kevaundray:
Done! Let me know if any other changes are needed :) The linter issue seems orthogonal to this PR

Pratyush:
Amazing, thanks a lot for the great PR! Maybe the only thing to add would be a short documentation note on the `_old` methods that specifies that they're there for compatibility with the old serialization format, and will be deprecated and removed in a future version. Adding a note in the CHANGELOG pointing to these methods would also be great =). Thanks again!

Pratyush (approved):

Issue #210, opened 2021-02-06
Precompute `2.inverse()` in `Fq2::sqrt`


Pratyush:
Right now in `QuadExtension::sqrt`, we compute `2.inverse()`, where `2` is in the basefield. An inverse in `bls12_381::Fq` is ~10,000ns, while `Fq2::sqrt` is 75,000ns, so this is quite a significant cost.

Pratyush:
Maybe one "easy" way would be to make a `const` version of `inverse`, and use that to precompute the constant at compile time.

jon-chuang:
Unfortunately, it's hard to define const ops over slices at the moment.

jon-chuang:
It might however, be possible to do so utilising the framework of const generics. Will investigate.

jon-chuang:
There is actually an easier way to compute any number div 2 than multiplying by 1/2. In montgomery form, one does a div2 on the bigint rep, then conditionally add 1/2 depending on whether the lowest bit had been 1.

Pratyush:
> then conditionally add a montgomery rep 1/2

Wouldn't we have to compute `2.inverse()` anyway?

jon-chuang:
Yes, but this is an orthogonal "optimisation" of the actual multiplication.

Pratyush:
Wait, isn't `2.inverse() == p+1/2`, and we can calculate `p+1/2` easily, as p is odd, so `p+1` is even, and so `p+1/2` is just a `div2()`. So basically we compute `add_no_carry` on the modulus, and use that?

jon-chuang:
Well yes. That's great 🥳

jon-chuang:
I think one nice thing about making biginteger a const generic struct rather than a trait is that one should be able to define as many const fns as one would like, without having to wait for this feature to become stabilised on rustc

Pratyush:
That's true, but the downside is that unrolling the `add_nocarry` and sub_noborrow` loops would again be a pain >.<

PR #352, opened 2021-11-24
Update `QuadExtField::sqrt` for better performance

vlopes11:
## Description

Currently, the QuadExtField uses an expensive runtime inverse of two
calculation defined as `P::BaseField::one().double().inverse()`.

With the proposed change, we compute the BigInt `(p+1)/2` that is ~15%
cheaper than the previous method.

Alternatively, we could require a compile-time constant provided by the
user that represents `1/2`. However, having a constant requirement to
satisfy a single use-case isn't ideal.

closes: #210

---

write a little note why.

- [ ] Wrote unit tests (tests are already covering this section: `bls12_381::tests::test_fq2`)

### DIFF
--- a/ff/src/fields/models/quadratic_extension.rs
+++ b/ff/src/fields/models/quadratic_extension.rs
@@ -19,6 +19,7 @@ use ark_std::rand::{
 };
 
 use crate::{
+    biginteger::BigInteger,
     bytes::{FromBytes, ToBytes},
     fields::{Field, LegendreSymbol, PrimeField, SquareRootField},
     ToConstraintField, UniformRand,
@@ -284,7 +285,6 @@ impl<P: QuadExtParameters> Field for QuadExtField<P> {
 
             // result.c1 = 2 * c0 * c1
             self.c1 = v2.double();
-            // result.c0 = (v0) + ((beta + 1) * v2)
             // result.c0 = (c0^2 - beta * c0 * c1 - c0 * c1 + beta * c1^2) + ((beta + 1) c0 * c1)
             // result.c0 = (c0^2 - beta * c0 * c1 + beta * c1^2) + (beta * c0 * c1)
             // result.c0 = c0^2 + beta * c1^2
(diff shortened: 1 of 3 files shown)


ff/src/fields/models/quadratic_extension.rs
```diff
@@ -355,11 +357,18 @@ where
...
         let alpha = self.norm();
-        // TODO: Precompute this
-        let two_inv = P::BaseField::one()
-            .double()
-            .inverse()
-            .expect("Two should always have an inverse");
+
+        // Compute `(p+1)/2` as `1/2`.
+        // This is cheaper than `P::BaseField::one().double().inverse()`
+        let mut two_inv = <<P as QuadExtParameters>::BasePrimeField as PrimeField>::Params::MODULUS;
```
mmagician, line 363:
```suggestion
        let mut two_inv = <P::BasePrimeField as PrimeField>::Params::MODULUS;
```

mmagician (changes requested):
Cool! I believe you can skip `<P as QuadExtsParameters>` since it's already dictated by the trait bounds.

Pratyush (approved):

ff/src/fields/models/quadratic_extension.rs
```diff
@@ -331,6 +332,7 @@ impl<P: QuadExtParameters> Field for QuadExtField<P> {
 impl<'a, P: QuadExtParameters> SquareRootField for QuadExtField<P>
 where
     P::BaseField: SquareRootField,
+    P::BaseField: From<<<P as QuadExtParameters>::BasePrimeField as PrimeField>::BigInt>,
```
mmagician, line 335:
```suggestion
    P::BaseField: From<<P::BasePrimeField as PrimeField>::BigInt>,
```

mmagician: same here

mmagician (approved):

daira (approved):
Untested ACK.

vlopes11:
Better rebase on top of https://github.com/arkworks-rs/algebra/pull/354 before merging so the performance difference will be clear and documented

PR #364, opened 2021-12-10
Add `ChunkedPippenger` to variable-base MSM

huyuncong:
## Description

ChunkedPippenger is a core algorithm for streaming MSM.
We could add streaming MSM on top of it after the stream utils are imported.
The test of this algorithm is added into tests macros in test-templates.

---

write a little note why.

- [x] Wrote unit tests


weikengchen:
Interesting: internal compiler error: unexpected panic
This is often due to the nightly being unstable.

The other bug: https://github.com/arkworks-rs/algebra/runs/4478158408?check_suite_focus=true
This could be fixed by using the full name of the struct.

ec/src/msm/mod.rs
```diff
@@ -3,6 +3,8 @@ mod variable_base;
 pub use fixed_base::*;
 pub use variable_base::*;
 
+pub mod stream_pippenger;
```
Pratyush, line 6: I would move this into `variable_base`, since it's a variable-base MSM

ec/src/msm/stream_pippenger.rs
```diff
@@ -0,0 +1,65 @@
...
+use crate::AffineCurve;
+use ark_ff::{PrimeField, Zero};
+
+use ark_std::{borrow::Borrow, ops::AddAssign, vec::Vec};
+
+/// Struct for the chunked Pippenger algorithm.
+pub struct ChunkedPippenger<G: AffineCurve> {
+    pub scalars_buffer: Vec<<G::ScalarField as PrimeField>::BigInt>,
+    pub bases_buffer: Vec<G>,
+    pub result: G::Projective,
```
Pratyush, line 11:
```suggestion
    pub result: G,
```

Pratyush, line 11:
```suggestion
    pub result: G,
```

ec/src/msm/stream_pippenger.rs
```diff
@@ -0,0 +1,65 @@
...
+    pub buf_size: usize,
+}
+
+impl<G: AffineCurve> ChunkedPippenger<G> {
+    /// Initialize a chunked Pippenger instance with default parameters.
+    pub fn new(max_msm_buffer: usize) -> Self {
+        Self {
+            scalars_buffer: Vec::with_capacity(max_msm_buffer),
+            bases_buffer: Vec::with_capacity(max_msm_buffer),
+            result: G::Projective::zero(),
```
Pratyush, line 21:
```suggestion
            result: G::zero(),
```

ec/src/msm/stream_pippenger.rs
```diff
@@ -0,0 +1,65 @@
...
+            buf_size: max_msm_buffer,
+        }
+    }
+    /// Initialize a chunked Pippenger instance with the given buffer size.
+    pub fn with_size(buf_size: usize) -> Self {
+        Self {
+            scalars_buffer: Vec::with_capacity(buf_size),
+            bases_buffer: Vec::with_capacity(buf_size),
+            result: G::Projective::zero(),
```
Pratyush, line 30:
```suggestion
            result: G::zero(),
```

Pratyush:
A micro-optimization that you might want to consider is to build up a buffer of results, and then sum that buffer up in parallel.

ec/src/msm/mod.rs
huyuncong: This is the core internal algorithm of streaming MSM. But we do have another streaming MSM function for public use that has not been imported yet because it relies on streaming utilities. The plan is to first decide which repo to put streaming utilities and then import streaming MSM into variable-base MSM.

weikengchen: Pratyush, I think here involves a decision, that is where to put this newer stream variable-base MSM---it clearly belongs to variable base. Does it replace the original variable base algorithm? If not, how should we name then?

huyuncong:
> Interesting: internal compiler error: unexpected panic This is often due to the nightly being unstable.

It seems the compiler error is not caused by the imported code. Any idea on fixing that?

weikengchen:
We can either wait for one more day or ignore that. Sometimes, nightly is just unstable.

mmaker:
> A micro-optimization that you might want to consider is to build up a buffer of results, and then sum that buffer up in parallel.

We can't. The chunk size `max_msm_buffer` is a constant and, if we want to keep log space, we can't store `N/max_msm_buffer` elements. We can, however, build them up in chunks again, but that seems.. overkill? After all it's just group addition. What would you expect from this? 

As another note: when/if you merge, can you please credit me? This code comes from [mmaker/gemini](https://github.com/mmaker/gemini)

Pratyush:
You can build up a `max_msm_buffer`-sized buffer of results, and then sum that. This would at most double your memory consumption. However that might still be out of your memory budget, and the time savings might not be sufficient to warrant it...

Re: the authorship, we'll merge it via `squash`-ing the PR, so if you add an empty commit then you should be automatically credited with co-authorship in the squash commit.

ec/src/msm/mod.rs
Pratyush: I mean that we add this algorithm as another file in the `variable_base` module, instead of making it a top-level member of the `msm` module.

mmaker:
Yah, that's what I meant; I'd stick with this "naïve" version for now. 
Added a marker, thanks!

weikengchen:
It seems that `pub mod variable_base;` can be changed back to `mod variable_base;`. So I change that.

weikengchen:
I will wait for Pratyush to do a final pass.

weikengchen (approved):

Pratyush (approved):

Issue #326, opened 2021-10-07
Move `COFACTOR`, `COFACTOR_INV`, and `is_in_correct_subgroup_assuming_on_curve()` to `ModelParameters`


Pratyush:
Currently they live in `SWModelParameters` and `TEModelParameters`, but they really should just live in `ModelParameters` as these don't depend on details of the model.

PR #365, opened 2021-12-11
Move common fields into `ModelParameters`

alexander-zw:
## Description

In addition to some simple changes to move `COFACTOR`, `COFACTOR_INV`, and `is_in_correct_subgroup_assuming_on_curve()`, a new trait `MultipliablePoint` had to be created because each type of curve has its own `GroupAffine` and `GroupProjective`. As a result, `is_in_correct_subgroup_assuming_on_curve()` now has more generic types, but fortunately if the corresponding `GroupAffine` struct has an identically named function, that one does not require any generics. Because of this change, some duplicate code in `ec/src/models/twisted_edwards_extended.rs` was removed.

closes: #326

---

write a little note why.

- [ ] Wrote unit tests


Pratyush:
Hm I think you don't need to introduce a new trait; just add `Affine` and `Projective` types into the `ModelParameters` trait.

alexander-zw:
> Hm I think you don't need to introduce a new trait; just add `Affine` and `Projective` types into the `ModelParameters` trait.

Ah I see, still getting used to Rust

alexander-zw:
Well I used the `AffineCurve` trait, but now I get the compile error
```
expected associated type, found struct `models::short_weierstrass_jacobian::GroupAffine`

note: expected reference `&<P as models::ModelParameters>::Affine`
         found reference `&models::short_weierstrass_jacobian::GroupAffine<P>`
help: consider constraining the associated type `<P as models::ModelParameters>::Affine` to `models::short_weierstrass_jacobian::GroupAffine<P>` or calling a method that returns `<P as models::ModelParameters>::Affine`
```
On the line
```
P::is_in_correct_subgroup_assuming_on_curve(self)
```

ec/src/lib.rs
```diff
@@ -274,10 +274,16 @@ pub trait AffineCurve:
...
+    /// big-endian bit-wise decomposition of the scalar.
+    fn mul_bits(&self, bits: impl Iterator<Item = bool>) -> Self::Projective;
+
     /// Performs scalar multiplication of this element with mixed addition.
     #[must_use]
-    fn mul<S: Into<<Self::ScalarField as PrimeField>::BigInt>>(&self, other: S)
-        -> Self::Projective;
+    #[inline]
+    fn mul<S: Into<<Self::ScalarField as PrimeField>::BigInt>>(&self, by: S) -> Self::Projective {
+        self.mul_bits(BitIteratorBE::new(by.into()))
```
Pratyush, line 285:
```suggestion
        self.mul_bits(BitIteratorBE::without_leading_zeros(by.into()))
```

ec/src/models/mod.rs
```diff
@@ -9,16 +9,29 @@ pub mod mnt6;
 pub mod short_weierstrass_jacobian;
 pub mod twisted_edwards_extended;
 
+/// Model parameters for an elliptic curve.
 pub trait ModelParameters: Send + Sync + 'static {
     type BaseField: Field + SquareRootField;
     type ScalarField: PrimeField + SquareRootField + Into<<Self::ScalarField as PrimeField>::BigInt>;
+    type Affine: AffineCurve;
```
Pratyush, line 16:
```suggestion
    type Affine: AffineCurve<BaseField = Self::BaseField, ScalarField = Self::ScalarField>;
```

ec/src/models/twisted_edwards_extended.rs
```diff
@@ -296,7 +293,7 @@ mod group_impl {
 
 //////////////////////////////////////////////////////////////////////////////
 
-/// `GroupProjective` implements Extended Twisted Edwards Coordinates
+/// `GroupProjective` implements Extended Twisted Edwards (Jacobian) Coordinates
```
Pratyush, line 296:
```suggestion
/// `GroupProjective` implements Extended Twisted Edwards Coordinates
```

alexander-zw: It says "Jacobian" coordinates for the SW `GroupProjective`. Are they not the same for TE?

Pratyush:
> Well I used the `AffineCurve` trait, but now I get the compile error
> 
> ```
> expected associated type, found struct `models::short_weierstrass_jacobian::GroupAffine`
> 
> note: expected reference `&<P as models::ModelParameters>::Affine`
>          found reference `&models::short_weierstrass_jacobian::GroupAffine<P>`
> help: consider constraining the associated type `<P as models::ModelParameters>::Affine` to `models::short_weierstrass_jacobian::GroupAffine<P>` or calling a method that returns `<P as models::ModelParameters>::Affine`
> ```
> 
> On the line
> 
> ```
> P::is_in_correct_subgroup_assuming_on_curve(self)
> ```

You have to constrain the `Parameters` associated type to match the respective `Affine` struct. You can do that by changing the definition of `TEParameters` to `trait TEParameters: ModelParameters<Affine = twisted_edwards::GroupAffine<Self>>`, and similarly for `SWParameters`

ec/src/models/mod.rs
```diff
@@ -9,16 +9,29 @@ pub mod mnt6;
...
+    /// Requires type parameters G: the type of point passed in, and H: the type of
+    /// point that results from multiplying G by a scalar.
+    fn is_in_correct_subgroup_assuming_on_curve(item: &Self::Affine) -> bool {
+        item.mul_bits(BitIteratorBE::new(Self::ScalarField::characteristic()))
+            .is_zero()
+    }
 }
 
+/// Model parameters for a Short Weierstrass curve.
 pub trait SWModelParameters: ModelParameters {
```
Pratyush, line 32:
```suggestion
pub trait SWModelParameters: ModelParameters<Affine = short_weierstrass_jacobian::GroupAffine> {
```

ec/src/models/mod.rs
```diff
@@ -37,21 +50,12 @@ pub trait SWModelParameters: ModelParameters {
...
-    where
-        Self: Sized,
-    {
-        item.mul_bits(BitIteratorBE::new(Self::ScalarField::characteristic()))
-            .is_zero()
-    }
 }
 
+/// Model parameters for a Twisted Edwards curve.
 pub trait TEModelParameters: ModelParameters {
```
Pratyush, line 56:
```suggestion
pub trait TEModelParameters: ModelParameters<Affine = twisted_edwards_extended::GroupAffine<Self>> {
```

ec/src/models/twisted_edwards_extended.rs
Pratyush: No, the non-affine struct for Twisted Edwards is not using the jacobian model. See [here](https://iacr.org/archive/asiacrypt2008/53500329/53500329.pdf) for details.

For SW curves, the `Projective` part is actually a bit of a misnomer; the Jacobian coordinate system is a bit different to the projective system; compare [this](https://en.wikibooks.org/wiki/Cryptography/Prime_Curve/Standard_Projective_Coordinates) with [this](https://en.wikibooks.org/wiki/Cryptography/Prime_Curve/Jacobian_Coordinates).

alexander-zw:
> > Well I used the `AffineCurve` trait, but now I get the compile error
> > ```
> > expected associated type, found struct `models::short_weierstrass_jacobian::GroupAffine`
> > 
> > note: expected reference `&<P as models::ModelParameters>::Affine`
> >          found reference `&models::short_weierstrass_jacobian::GroupAffine<P>`
> > help: consider constraining the associated type `<P as models::ModelParameters>::Affine` to `models::short_weierstrass_jacobian::GroupAffine<P>` or calling a method that returns `<P as models::ModelParameters>::Affine`
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > On the line
> > ```
> > P::is_in_correct_subgroup_assuming_on_curve(self)
> > ```
> 
> You have to constrain the `Parameters` associated type to match the respective `Affine` struct. You can do that by changing the definition of `TEParameters` to `trait TEParameters: ModelParameters<Affine = twisted_edwards::GroupAffine<Self>>`, and similarly for `SWParameters` (i.e., see the suggestions below).

Oh I was going to change every instance of `Parameters` into `Parameters<Affine = Self>`, your suggestion is much easier

ec/src/models/mod.rs
alexander-zw: It gives me
```
the size for values of type `Self` cannot be known at compilation time
```

alexander-zw: Huh adding `+ Sized` to `ModelParameters` seems to fix it

ec/src/models/mod.rs
```diff
@@ -9,16 +9,29 @@ pub mod mnt6;
...
     type BaseField: Field + SquareRootField;
     type ScalarField: PrimeField + SquareRootField + Into<<Self::ScalarField as PrimeField>::BigInt>;
+    type Affine: AffineCurve<BaseField = Self::BaseField, ScalarField = Self::ScalarField>;
+
+    const COFACTOR: &'static [u64];
+    const COFACTOR_INV: Self::ScalarField;
+
+    /// Checks that the current point is in the prime order subgroup given
+    /// the point on the curve.
+    fn is_in_correct_subgroup_assuming_on_curve(item: &Self::Affine) -> bool {
```
Pratyush, line 23:
```suggestion
    /// Check that `element` is in the prime-order subgroup corresponding to `Self::ScalarField`.
    fn is_in_correct_subgroup_assuming_on_curve(element: &Self::Affine) -> bool {
```

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@
 - [\#345](https://github.com/arkworks-rs/algebra/pull/345) (arc-ec, ark-serialize) Change the serialization format for Twisted Edwards Curves. We now encode the Y coordinate and take the sign bit of the X co-ordinate, the default flag is also now the Positive X value. The old methods for backwards compatibility are located [here](https://github.com/arkworks-rs/algebra/pull/345/files#diff-3621a48bb33f469144044d8d5fc663f767e103132a764812cda6be6c25877494R860)
 - [\#348](https://github.com/arkworks-rs/algebra/pull/348) (arc-ec) Rename `msm:{Fixed,Variable}BaseMSM:multi_scalar_mul` to `msm:{Fixed,Variable}:msm` to avoid redundancy.
 - [\#359](https://github.com/arkworks-rs/algebra/pull/359) (ark-test-templates) Simplify the field and curve test macros.
+- [\#365](https://github.com/arkworks-rs/algebra/pull/365) (arc-ec) Move `COFACTOR`, `COFACTOR_INV`, and `is_in_correct_subgroup_assuming_on_curve()` from `SWModelParameters` etc into `ModelParameters`. As a side effect, `mul_bits` is now required in `AffineCurve` but `mul` is implemented.
```
alexander-zw, line 15: How detailed does the changelog need to be, because if I listed all my changes to the API it would take about ten lines

ec/src/models/mod.rs
```diff
@@ -9,16 +9,29 @@ pub mod mnt6;
 pub mod short_weierstrass_jacobian;
 pub mod twisted_edwards_extended;
 
-pub trait ModelParameters: Send + Sync + 'static {
+/// Model parameters for an elliptic curve.
+pub trait ModelParameters: Send + Sync + Sized + 'static {
```
alexander-zw, line 13: I still don't exactly understand what `Sized` does, this doesn't break anything does it?

ec/src/lib.rs
```diff
@@ -274,10 +274,16 @@ pub trait AffineCurve:
     /// random group elements from a hash-function or RNG output.
     fn from_random_bytes(bytes: &[u8]) -> Option<Self>;
 
+    /// Multiplies `self` by the scalar represented by `bits`. `bits` must be a
+    /// big-endian bit-wise decomposition of the scalar.
+    fn mul_bits(&self, bits: impl Iterator<Item = bool>) -> Self::Projective;
```
Pratyush, line 279: You can provide a default implementation of `mul_bits` that uses double-and-add.

CHANGELOG.md
Pratyush, line 15:
```suggestion
- [\#365](https://github.com/arkworks-rs/algebra/pull/365) (arc-ec) 
	- Move `COFACTOR`, `COFACTOR_INV`, and `is_in_correct_subgroup_assuming_on_curve()` from `{SW,TE}ModelParameters` into `ModelParameters`. 
	- Add `mul_bits` to `AffineCurve` and provide a default implementation of `mul` using this.
```

ec/src/models/mod.rs
Pratyush: Nah, that's fine. All `ModelParameters` are `Sized`. `Sized` just means that the struct implementing the trait has a size known at compile time (i.e. it's not dynamically sized like a slice type)

ec/src/lib.rs
alexander-zw: Is this more efficient? Why isn't it the default `new` implementation?

ec/src/models/short_weierstrass_jacobian.rs
```diff
@@ -80,20 +80,6 @@ impl<P: Parameters> GroupAffine<P> {
         self.mul_bits(cofactor)
     }
 
-    /// Multiplies `self` by the scalar represented by `bits`. `bits` must be a
-    /// big-endian bit-wise decomposition of the scalar.
-    pub(crate) fn mul_bits(&self, bits: impl Iterator<Item = bool>) -> GroupProjective<P> {
```
alexander-zw, line 85: I also don't understand what `pub(crate)` does, does it matter that it was removed in the new version of `mul_bits`?

ec/src/lib.rs
alexander-zw: It doesn't work because `double_in_place` is not required in `AffineCurve`

alexander-zw: I also considered moving `mul_by_cofactor_to_projective` and `mul_by_cofactor_inv` into `AffineCurve` but it doesn't work because `AffineCurve` doesn't have the `P: ModelParameters` generic

ec/src/models/short_weierstrass_jacobian.rs
Pratyush: It makes `mul_bits` public outside the crate, when previously it was private.

ec/src/lib.rs
Pratyush:
```suggestion
    fn mul_bits(&self, bits: impl Iterator<Item = bool>) -> Self::Projective {
        let mut res = Self::Projective::zero();
        // Skip leading zeros.
        for i in bits.skip_while(|b| !b) {
            res.double_in_place();
            if i {
                res.add_assign_mixed(&self)
            }
        }
        res
    }
```

Pratyush: ^ this should work

ec/src/models/twisted_edwards_extended.rs
```diff
@@ -140,8 +130,17 @@ impl<P: Parameters> AffineCurve for GroupAffine<P> {
...
+    /// big-endian bit-wise decomposition of the scalar.
+    fn mul_bits(&self, bits: impl Iterator<Item = bool>) -> GroupProjective<P> {
+        let mut res = GroupProjective::zero();
+        for i in bits.skip_while(|b| !b) {
+            res.double_in_place();
+            if i {
+                res.add_assign_mixed(&self)
+            }
+        }
+        res
```
Pratyush, line 143:
```suggestion
```

ec/src/models/short_weierstrass_jacobian.rs
```diff
@@ -228,10 +216,18 @@ impl<P: Parameters> AffineCurve for GroupAffine<P> {
...
+        let mut res = GroupProjective::zero();
+        // Skip leading zeros.
+        for i in bits.skip_while(|b| !b) {
+            res.double_in_place();
+            if i {
+                res.add_assign_mixed(&self)
+            }
+        }
+        res
     }
```
Pratyush, line 231:
```suggestion
```

ec/src/models/twisted_edwards_extended.rs
```diff
@@ -140,8 +130,17 @@ impl<P: Parameters> AffineCurve for GroupAffine<P> {
...
     }
```
Pratyush, line 144:
```suggestion
```

ec/src/models/short_weierstrass_jacobian.rs
alexander-zw: So it's no longer needed because now it's in the trait, not the impl?

ec/src/lib.rs
alexander-zw: But the `double_in_place`?

Pratyush: `Self::Projective` should have a `double_in_place`, so this should work?

ec/src/models/short_weierstrass_jacobian.rs
Pratyush: Yeah, by moving it into the trait it's publicly visible

ec/src/lib.rs
alexander-zw: Oooh I didn't read the error clearly, it says `multiple double_in_place found`

ec/src/lib.rs
alexander-zw: It doesn't seem that this does anything in this case, because `mul_bits()` calls `skip_while()` anyway

ec/src/models/short_weierstrass_jacobian.rs
```diff
@@ -228,12 +216,6 @@ impl<P: Parameters> AffineCurve for GroupAffine<P> {
...
 
-    #[inline]
-    fn mul<S: Into<<Self::ScalarField as PrimeField>::BigInt>>(&self, by: S) -> GroupProjective<P> {
-        let bits = BitIteratorBE::new(by.into());
-        self.mul_bits(bits)
-    }
-
     #[inline]
     fn mul_by_cofactor_to_projective(&self) -> Self::Projective {
         self.scale_by_cofactor()
```
alexander-zw, line 221: Why do they have both `mul_by_cofactor_to_projective` and `scale_by_cofactor` if they do the exact same thing?

ec/src/models/short_weierstrass_jacobian.rs
```diff
@@ -133,9 +113,11 @@ impl<P: Parameters> GroupAffine<P> {
...
         }
     }
+}
 
-    /// Checks if `self` is in the subgroup having order that equaling that of
-    /// `P::ScalarField`.
+impl<P: Parameters> GroupAffine<P> {
+    /// Checks if `self` is in the subgroup having order equaling that of
+    /// `P::ScalarField` given it is on the curve.
     pub fn is_in_correct_subgroup_assuming_on_curve(&self) -> bool {
```
alexander-zw, line 121: Why aren't `is_on_curve()` and `is_in_correct_subgroup_assuming_on_curve()` part of the `AffineCurve` trait?

ec/src/models/short_weierstrass_jacobian.rs
alexander-zw: Update: I removed `scale_by_cofactor`

ec/src/models/short_weierstrass_jacobian.rs
Pratyush: Because they're supposed to represent points that are already on the curve and in the appropriate subgroup.

ec/src/models/short_weierstrass_jacobian.rs
```diff
@@ -74,26 +74,6 @@ impl<P: Parameters> GroupAffine<P> {
         Self { x, y, infinity }
     }
 
-    /// Multiply `self` by the cofactor of the curve, `P::COFACTOR`.
-    pub fn scale_by_cofactor(&self) -> GroupProjective<P> {
-        let cofactor = BitIteratorBE::new(P::COFACTOR);
-        self.mul_bits(cofactor)
-    }
-
```
Pratyush, line 82: Hm why'd you delete this?

ec/src/models/short_weierstrass_jacobian.rs
```diff
@@ -228,15 +210,9 @@ impl<P: Parameters> AffineCurve for GroupAffine<P> {
...
-    fn mul<S: Into<<Self::ScalarField as PrimeField>::BigInt>>(&self, by: S) -> GroupProjective<P> {
-        let bits = BitIteratorBE::new(by.into());
-        self.mul_bits(bits)
-    }
-
     #[inline]
     fn mul_by_cofactor_to_projective(&self) -> Self::Projective {
-        self.scale_by_cofactor()
+        self.mul_bits(BitIteratorBE::new(P::COFACTOR))
     }
```
Pratyush, line 216: I guess now that `mul_bits` is available in the trait, we can move this implementation to the trait impl. This would further deduplicate code across the TE and SW affine impls.
```rust
pub trait AffineCurve {
	// rest
	
	#[inline]
    fn mul_by_cofactor_to_projective(&self) -> Self::Projective {
        self.mul_bits(BitIteratorBE::new(P::COFACTOR))
    }
}
```

(And the same for `mul_by_cofactor_inverse`.).

ec/src/models/short_weierstrass_jacobian.rs
Pratyush: Ah I see now, it was duplicated with the trait impl. Thanks!

Pratyush (approved):

PR #366, opened 2021-12-17
Fix Cubic Ext norm

mmagician:
## Description

Previously, the `norm()` method didn't work on towered extension fields, like 3_over_2.
This PR fixes it.

---

write a little note why.

- [x] Wrote unit tests


test-curves/src/mnt6_753/mod.rs
```diff
@@ -0,0 +1,9 @@
+#[cfg(feature = "mnt6_753")]
```
Pratyush, line 1: This feature flag only flags the existence of the `fr` module; not the other modules or re-exports; is this the intended behaviour?

Pratyush (commented):
Thanks for the PR! This looks good modulo the feature flag comment

test-curves/src/mnt6_753/mod.rs
mmagician: good catch! thx

Pratyush (approved):

PR #369, opened 2021-12-22
Move model-specific methods and types back into model-specific traits

Pratyush:
## Description

Makes some changes to arkworks-rs/algebra#365, since we can't actually make `Affine` common to `ModelParameters`. In particular, the same curve can have different models, and hence different `Affine` representations.

---

write a little note why.

- [ ] Wrote unit tests

### DIFF
--- a/ec/src/models/mod.rs
+++ b/ec/src/models/mod.rs
@@ -55,7 +55,6 @@ pub trait SWModelParameters: ModelParameters {
 pub trait TEModelParameters: ModelParameters {
     const COEFF_A: Self::BaseField;
     const COEFF_D: Self::BaseField;
-    type Affine: AffineCurve<BaseField = Self::BaseField, ScalarField = Self::ScalarField>;
     const AFFINE_GENERATOR_COEFFS: (Self::BaseField, Self::BaseField);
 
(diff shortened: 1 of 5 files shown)


ec/src/models/mod.rs
```diff
@@ -68,6 +65,15 @@ pub trait TEModelParameters:
...
+    /// the point on the curve.
+    fn is_in_correct_subgroup_assuming_on_curve(
+        item: &twisted_edwards_extended::GroupAffine<Self>,
+    ) -> bool {
+        item.mul_bits(BitIteratorBE::new(Self::ScalarField::characteristic()))
+            .is_zero()
+    }
 }
 
 /// Model parameters for a Montgomery curve.
```
alexander-zw, line 79: Why isn't there one for a Montgomery curve?

Pratyush: Mostly because we only ever instantiate the Montgomery model of a curve in a very specific circumstance (inside a particular hash function), and so we don't need it generally).

alexander-zw (approved):

alexander-zw (approved):

PR #371, opened 2021-12-29
Add array serialization impls

ryanleh:
## Description

This PR impls `CanonicalSerialize` and `CanonicalDeserialize` for arrays of up to size 32. This is useful for deriving these traits on structs which contain arrays.

I put `Default` and `Copy` bounds on `CanonicalDeserialize` in order to initialize the initial empty array. We could get rid of these bounds and use an `Option<T>`, but I believe this necessitates heap allocation in order to unwrap the `Option<T>` for the final result which I thought best to avoid. Let me know what y'all think is best.

I'll complete the other TODOs once I get some form of approval from maintainers to move forward (not sure who to ask for review 😁 )

---

write a little note why.

- [x] Wrote unit tests
- [x] Updated relevant documentation in the code (NA)

### DIFF
--- a/serialize/src/lib.rs
+++ b/serialize/src/lib.rs
@@ -45,8 +45,8 @@ pub trait CanonicalSerializeWithFlags: CanonicalSerialize {
 /// The serialization format must be 'length-extension' safe.
 /// e.g. if T implements Canonical Serialize and Deserialize,
 /// then for all strings `x, y`, if `a = T::deserialize(Reader(x))` and `a` is
-/// not an error, then it must be the case that `a = T::deserialize(Reader(x || y))`,
-/// and that both readers read the same number of bytes.
+/// not an error, then it must be the case that `a = T::deserialize(Reader(x ||
+/// y))`, and that both readers read the same number of bytes.
 ///
 /// This trait can be derived if all fields of a struct implement
 /// `CanonicalSerialize` and the `derive` feature is enabled.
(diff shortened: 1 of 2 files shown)


mmagician:
@ryanleh I recently started re-working [an oldish PR](#186) to replace some parts of the code with const generics: I have a draft of the changes to bigint crate here: #372. 
Perhaps it would make sense for this PR to also utilise const generics? I believe it would be analogous to [this impl](https://github.com/mmagician/algebra/blob/bigint-const-generics/ff/src/biginteger/mod.rs#L283).

Otherwise `Copy` and `Default` trait bounds sound reasonable. I think most of the types from algebra that would make sense to put in an array have these implemented.

ryanleh:
@mmagician Ah, had totally missed that const generics had stabilized!! 😁 That definitely makes more sense -- I updated the PR appropriately.

serialize/src/lib.rs
```diff
@@ -313,6 +313,75 @@ impl<T: CanonicalSerialize> CanonicalSerialize for [T] {
...
+
+    #[inline]
+    fn uncompressed_size(&self) -> usize {
+        self.iter()
+            .map(|item| item.uncompressed_size())
+            .sum::<usize>()
+    }
+}
+
+impl<T: CanonicalDeserialize + Default + Copy, const N: usize> CanonicalDeserialize for [T; N] {
```
Pratyush, line 356: I believe we can use new array methods to eliminate the `Default + Copy` bounds:
```rust
Ok([(); N].map(|_| T::deserialize(&mut reader).unwrap()))
```

There is a downside that this panics immediately if there's an error, but that's not necessarily a bad thing. This might be the best we can do until [this feature](https://github.com/rust-lang/rust/issues/89379) is stabilized.

ryanleh: Awesome! I changed de-serialization to do this, removed the extra bounds, and added a `TODO` to update things once the feature is stabilized.

ryanleh:
For some reason the `CanonicalSerialize` impl is causing serialization tests for BTreeMap, BTreeSet, Tuple, Option, and Rc to panic..? Not sure what's going on lol I'll take a look in a bit.

Pratyush:
> For some reason the `CanonicalSerialize` impl is causing serialization tests for BTreeMap, BTreeSet, Tuple, Option, and Rc to panic..? Not sure what's going on lol I'll take a look in a bit.

Should be fixed in the latest commit

Pratyush (approved):

PR #372, opened 2021-12-30
BigInt const generics

mmagician:
## Description

This is an updated PR based on the subset of #186, i.e. replacing the macros with const generics only for BigInt crate. As such, it should be easier to merge.
Breaking change -> will need an accompanying PR in the curves repo.

---

write a little note why.

- [ ] Wrote unit tests


burdges:
cc @tarcieri

tarcieri:
I talked with @Pratyush about [`crypto-bigint`](https://github.com/rustcrypto/crypto-bigint) in the past.

This PR is doing something somewhat similar, but the main difference would be `crypto-bigint` uses `[Limb; N`], where `Limb` is a newtype wrapper for `u32` on 32-bit architectures and `u64` on 64-bit architectures, which seems like it'd be a bigger change than what's in this PR, so perhaps this is a good incremental step if there's an eventual desire to use `crypto-bigint`.

Pratyush (approved):

Pratyush:
One additional thing we might want to consider doing is moving all the impls in the traits directly to inherent methods, and having the trait forward to those inherent impls. This would enable calling the `BigInt` methods in const contexts.

test-curves/src/bn384_small_two_adicity/fr.rs
```diff
@@ -102,7 +103,7 @@ impl FpParameters for FrParameters {
     ]);
 
     // (mod - 1) / 2 = 2972938801625915898129258746014768257744324656783561314222519104145798272973804394996417217026588261812051162269696
-    const MODULUS_MINUS_ONE_DIV_TWO: BigInteger = BigInteger([
+    const MODULUS_MINUS_ONE_DIV_TWO: BigInteger = BigInt::<6>([
```
Pratyush, line 106: I think we can use the `BigInteger::new` method here to avoid specifying the number of limbs. (It's unfortunate that we can't use the type alias as the constructor =()

Pratyush: Ditto for the fixes in `curves`

tarcieri:
> One additional thing we might want to consider doing is moving all the impls in the traits directly to inherent methods, and having the trait forward to those inherent impls. This would enable calling the BigInt methods in const contexts.

This is what `crypto-bigint` does but it does have a downside: it prevents the usage of intrinsics.

One workaround is to keep the intrinsics in the trait impls, and const pure Rust.

Pratyush:
> This is what `crypto-bigint` does but it does have a downside: it prevents the usage of intrinsics.

Good point. 

> One workaround is to keep the intrinsics in the trait impls, and const pure Rust.

Yeah, that's a good trade-off, but in that case we might need to name the inherent and trait methods differently, to avoid accidentally using the slower one. Maybe it's just better to wait for `const` in trait impls...

test-curves/src/bn384_small_two_adicity/fr.rs
weikengchen: One repo needs to move first. It can be algebra.

Pratyush (approved):

PR #396, opened 2022-03-08
add `mul` to ModelParam traits

zhenfeizhang:
## Description

implement part of the solution suggested via 
https://github.com/arkworks-rs/curves/pull/66#issuecomment-949055796

---

write a little note why.

- [ ] ~~Wrote unit tests~~

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,7 @@
 - [\#383](https://github.com/arkworks-rs/algebra/pull/383) (`ark-ff`) Rename `BigInteger::add_nocarry` to `add_with_carry` and `sub_noborrow` to `sub_with_borrow`.
 - [\#386](https://github.com/arkworks-rs/algebra/pull/386) (`ark-ff`) Remove `PrimeField::GENERATOR`, since it already exists on `FftField`.
 - [\#393](https://github.com/arkworks-rs/algebra/pull/393) (`ark-ec`, `ark-ff`) Rename `FpXParams` to `FpXConfig` and `FpXParamsWrapper` to `FpXConfigWrapper`.
+- [\#396](https://github.com/arkworks-rs/algebra/pull/396) (`ark-ec`) Remove `mul_bits` feature, and remove default implementations of `mul` and `mul_by_cofactor_to_projective`.
 
 ### Features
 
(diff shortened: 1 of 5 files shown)


Pratyush:
@zhenfeizhang Thanks! do you mind adding undoing the formatting changes? (I think a git reset + git add of just the model parameters file should work)

Also, I think @alexander-zw was also looking into similar things.

CHANGELOG.md
```diff
@@ -33,6 +33,7 @@
 
 ### Features
 
+- [\#396](https://github.com/arkworks-rs/algebra/pull/396) (`ark-ec`) Add a default `mul` function to `TE/SWModelParameter` trait definition.
```
Pratyush, line 36: Could you move this to the end of the section? (We have reverse chronological order)

ec/src/models/mod.rs
```diff
@@ -64,21 +66,32 @@ pub trait SWModelParameters: ModelParameters {
...
             .is_zero()
     }
+
+    /// Default implementation of group multiplication for projective
+    /// coordinates
+    fn mul(
+        base: &short_weierstrass_jacobian::GroupProjective<Self>,
+        scalar: &Self::ScalarField,
+    ) -> short_weierstrass_jacobian::GroupProjective<Self> {
+        base.mul(scalar.into_bigint())
```
Pratyush, line 88: What I had in mind was actually something even more aggressive: to move the actual `mul` code here, and call `P::mul` from `GroupProjective::mul`. That is, the impl of this method in `GroupProjective` changes to
```rust
fn mul(...) {
    P::mul(...)
}
```

Also, we might want to add both `mul_projective` and `mul_affine` methods here.

zhenfeizhang: I just tried your suggestion.

- we cannot add it directly to `ModelParameter`: we need to specify if it is TE or SW
- we also cannot directly invoke it from `AffineCurve` or `ProjectiveCurve`, the trait is bounded by `ModelParameter` that does not specify `TE` or `SW`

working on the rest of the changes now

zhenfeizhang: a workaround is to add an associated type for `ModelParameters` but it will be ugly and is a breaking change

Pratyush: Ah sorry, should have been clearer, I meant that in the current strategy, instead of calling `base.mul(...)` inside `SWModelParameters`, we should change the implementation as follows:
```rust
// inside SW/TEModelParameters:
pub trait SW/TEModelParameters {
	fn mul_projective(g: GroupProjective<Self>, scalar: ...) -> GroupProjective<Self> {
	    // default double-and-add impl
	} 
}

// inside `ProjectiveCurve`:
pub trait ProjectiveCurve {
	fn mul(...); // <- no default impl anymore
}

// inside sw/te::GroupProjective:
impl ProjectiveCurve for GroupProjective {
    ...
	
	fn mul() {
		P::mul_projective(...)
	}
}
```

zhenfeizhang: Got you. Should be fixed now.

Pratyush (approved):

PR #397, opened 2022-03-11
Add MSM with chunks and HashMapPippenger  

huyuncong:
## Description

Add steaming multi-scalar multiplication algorithm with chunks.
Add Pippenger algorithm using hash map struct.

Note that we use #![feature(iter_advance_by)] in the MSM with chunks.

---

write a little note why.

- [x] Wrote unit tests


weikengchen:
It seems that the biggest problem is that we cannot use unstable features in arkworks-rs. And `iter_advance_by` does not seem to become stable very soon.

Any workaround for now?

ec/src/msm/variable_base/stream_pippenger.rs
```diff
@@ -65,3 +65,66 @@ impl<G: AffineCurve> ChunkedPippenger<G> {
...
 }
+
+/// Hash map struct for Pippenger algorithm.
+pub struct HashMapPippenger<G: AffineCurve> {
+    pub buffer: HashMap<G, G::ScalarField>,
+    pub result: G::Projective,
+}
+
+impl<G: AffineCurve> HashMapPippenger<G> {
+    /// Producce a new hash map with the maximum msm buffer size.
```
weikengchen, line 76:
```suggestion
    /// Produce a new hash map with the maximum msm buffer size.
```

ec/src/msm/variable_base/stream_pippenger.rs
```diff
@@ -65,3 +65,66 @@ impl<G: AffineCurve> ChunkedPippenger<G> {
...
+        B: Borrow<G>,
+        S: Borrow<G::ScalarField>,
+    {
+        // update the entry, guarding the possibility that it has been already set.
+        let entry = self
+            .buffer
+            .entry(*base.borrow())
+            .or_insert(G::ScalarField::zero());
+        *entry += *scalar.borrow();
+        if self.buffer.len() == self.buffer.capacity() {
```
weikengchen, line 97: @huyuncong @mmaker I have a concern about whether `capacity()` will run in the expected way. Rust did not promise that the capacity of the map would not grow on its own, and Rust did not promise that it will allocate exactly the capacity as specified during `with_capacity`.

Note: Rust's hashmap is often constructed through the RawTable here, which may specify a slightly different capacity compared with the one from `with_capacity`.
https://docs.rs/hashbrown/latest/src/hashbrown/raw/mod.rs.html#191

ec/src/msm/variable_base/mod.rs
```diff
@@ -142,13 +142,12 @@ impl VariableBase {
...
-        let mut bases = bases_stream.iter();
+        let bases_init = bases_stream.iter();
         let mut scalars = scalars_stream.iter();
 
         // align the streams
-        bases
-            .advance_by(bases_stream.len() - scalars_stream.len())
-            .expect("bases not long enough");
+        let mut bases = bases_init.skip(bases_stream.len() - scalars_stream.len());
+        // .expect("bases not long enough");
```
Pratyush, line 151:
```suggestion
```

weikengchen: This line was sort of left for the future change back to `advance_by`. But I think maybe we can just remove it for now.

mmaker: I'd rather write explicitly the issue on rust-lang that is keeping track of the implementation of `advance_by`

Pratyush: That indeed sounds better!

ec/src/msm/variable_base/mod.rs
```diff
@@ -127,4 +128,43 @@ impl VariableBase {
...
+        J::Item: Borrow<G>,
+    {
+        assert!(scalars_stream.len() <= bases_stream.len());
+
+        // remove offset
+        let bases_init = bases_stream.iter();
+        let mut scalars = scalars_stream.iter();
+
+        // align the streams
+        // NOTE: Change skip() to iter_advance_by() when it becomes stable.
```
mmaker, line 149:
```suggestion
        // TODO: change `skip` to `advance_by` once rust-lang/rust#7774 is fixed. 
        // See <https://github.com/rust-lang/rust/issues/77404>
```

ec/src/msm/variable_base/stream_pippenger.rs
```diff
@@ -65,3 +65,66 @@ impl<G: AffineCurve> ChunkedPippenger<G> {
         self.result
     }
 }
+
+/// Hash map struct for Pippenger algorithm.
+pub struct HashMapPippenger<G: AffineCurve> {
+    pub buffer: HashMap<G, G::ScalarField>,
```
mmaker, line 71:
```suggestion
    buffer: HashMap<G, G::ScalarField>,
```

ec/src/msm/variable_base/mod.rs
weikengchen: Actually, I feel it is not a bug (``fixed''), it is just a feature that is pending to be stable.

Pratyush:
If you could run `cargo fmt` that'd be great!

mmaker:
Done! (staging only the changes in ec/msm/)

ec/Cargo.toml
```diff
@@ -23,6 +23,7 @@ digest = { version = "0.10", default-features = false }
 num-traits = { version = "0.2", default-features = false }
 rayon = { version = "1", optional = true }
 zeroize = { version = "1", default-features = false, features = ["zeroize_derive"] }
+hashbrown = "0.11.2"
 
 [dev-dependencies]
 hashbrown = { version = "0.11.1" }
```
Pratyush, line 29:
```suggestion
```

Pratyush (approved):

Issue #308, opened 2021-08-26
Enable specialization of subgroup checks, scalar mul using GLV, deserialization, etc via `SW/TEModelParameter` traits


Pratyush:
In many cases, certain elliptic curves have more efficient algorithms for operations like subgroup checks and scalar multiplication routines. I'm going to propose a method to enable specializing these in a backwards compatible way.

We'll use the example of specializing subgroup checks on Short Weierstrass curves for this example, but it should apply equally to scalar mulsTE curves.

First, we modify SWModelParameters to include a new method
```rust
pub trait SWModelParameters {
    // New method
    // It has a default impl based on the existing one, but can be specialized by implementors of this trait.
    fn is_in_correct_subgroup_assuming_on_curve(item: &SWProjective/Affine<Self>) -> bool {
        // Default impl based on scalar mul
    }
    
    fn scalar_mul(item: &SWProjective/Affine<Self>) -> bool {
        // Default impl
    }
    
    fn serialize(item: &SWProjective/Affine<Self>) -> Vec<u8> {
        // Default impl based on existing canonical serialize impl
    }
    
    fn deserialize(item: &SWProjective/Affine<Self>) -> bool {
        // Default impl based on existing canonical deserialize impl
    }
}
```
Then, we update `SWProjective/Affine` to invoke these method:
```rust
// in ark_ec::short_weierstrass_jacobian

impl<P: SWModelParameters> SWProjective/Affine<P> {
    fn is_in_correct_subgroup_assuming_on_curve(&self) -> bool {
        P:: is_in_correct_subgroup_assuming_on_curve(self)
    }
    // same for others
}
```
So far, this is a backwards compatible change: implementers of `SWModelParameters` don't have to change anything if they don't want to (because of the default impl), but can do so if they wish.

----------------------------------------------------------------------------------------

The foregoing is good enough to specialize individual curves. However, for some families of elliptic curves, the specializations apply to *all* curves in that family, and so we would like all these curves to inherit the specializations. 

For example, the BLS12 curves have special subgroup checks and scalar multiplication routines owing to their j-invariant (0). We can change our architecture to specialize these, as follows:

```rust
// In ark-ec:

pub trait Bls12Parameters {
    // Existing stuff, but remove G1Parameters and G2Parameters.
}

pub trait Bls12G1Parameters {
    type P: Bls12Parameters;
    const COEFF_A: Self::P::Fq;
    const COEFF_B: Self::P::Fq;
[...]

PR #403, opened 2022-04-04
Deduplicate curve sample methods for SW/TE

alexander-zw:
## Description

- There are no breaking changes, but ~now users can use their own implementation of `sample()`~ minor refactor to remove duplicate code. There should be no change in behavior or even performance.
- Simple documentation.
- I did not add unit tests because it's difficult to test randomness.

~closes: #308~

---

write a little note why.

- [ ] Wrote unit tests

### DIFF
--- a/ec/src/models/short_weierstrass/affine.rs
+++ b/ec/src/models/short_weierstrass/affine.rs
@@ -186,16 +186,10 @@ impl<P: SWCurveConfig> Zeroize for Affine<P> {
 }
 
 impl<P: SWCurveConfig> Distribution<Affine<P>> for Standard {
+    /// Generates a uniformly random instance of the curve.
     #[inline]
     fn sample<R: Rng + ?Sized>(&self, rng: &mut R) -> Affine<P> {
-        loop {
-            let x = P::BaseField::rand(rng);
-            let greatest = rng.gen();
-
-            if let Some(p) = Affine::get_point_from_x_unchecked(x, greatest) {
-                return p.mul_by_cofactor();
-            }
-        }
+        Projective::sample(self, rng).into()
     }
 }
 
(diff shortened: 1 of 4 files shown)


ec/src/models/mod.rs
```diff
@@ -111,6 +113,24 @@ pub trait SWModelParameters: ModelParameters {
 
         res
     }
+
+    /// Generates a uniformly random affine instance of the group.
+    /// Implements `Distribution<GroupProjective<Self>>` for `Standard`.
+    #[inline]
```
alexander-zw, line 119: I moved the inline directive to the method in the ModelParameters, but removed the ones in each group impl. Not sure if that's acceptable.

ec/src/models/short_weierstrass_jacobian.rs
```diff
@@ -10,12 +15,6 @@ use ark_std::{
     vec::Vec,
 };
```
alexander-zw, line 17: Not sure why there was a blank line here, I thought it made more sense to remove it, then ran cargo fmt.

mmagician (commented):
What's the rationale of customising `sample`? From my understanding of [this issue](https://github.com/arkworks-rs/algebra/issues/308), we would want to allow downstream implementers of {SW/TE}ModelParameters to have a specialised implementation due to more efficient algorithms being known for some curves. I can't really see why we'd want this for `sample` - I'd have thought it'd mostly be used for testing and experimentation.

alexander-zw:
@Pratyush ?

alexander-zw:
Since we don't want to allow customization of sample(), I scaled down the changes to only make a small refactor.

alexander-zw:
Never mind

PR #405, opened 2022-04-19
IETF hash-to-field

mmagician:
## Description

This is the second iteration of the hash-to-field efforts.
In this take, we have:
- refactored the traits a little to more closely follow the specifications in the [IETF draft](https://datatracker.ietf.org/doc/draft-irtf-cfrg-hash-to-curve/?include_text=1)
- made expander into its own, contained and easily testable module
- added a comprehensive test suite

We will be following a similar approach for the map-to-curve part (especially a full test suite), but have decided to split the work over two, more manageble and independent PRs, with the second one to follow once ready.
@drskalman 

---

write a little note why.

- [x] Wrote unit tests


ec/src/hashing/field_hashers/mod.rs
```diff
@@ -1,106 +1,106 @@
...
+/// use ark_ec::hashing::field_hashers::DefaultFieldHasher;
+/// use sha2::Sha256;
+/// use crate::ark_ec::hashing::map_to_curve_hasher::HashToField;
+///
+/// let hasher = <DefaultFieldHasher<Sha256, 128> as HashToField<Fq>>::new_hash_to_field(&[1, 2, 3]).unwrap();
+/// let field_elements: Vec<Fq> = hasher.hash_to_field(b"Hello, World!", 2).unwrap();
+///
+/// assert_eq!(field_elements.len(), 2);
+/// ```
+pub struct DefaultFieldHasher<H: Default + DynDigest + Clone, const SEC_PARAM: usize> {
```
Pratyush, line 28: Do we want to have a default `SEC_PARAM`?  Default const generics are stable since 1.59.

ec/src/hashing/map_to_curve_hasher.rs
```diff
@@ -20,10 +20,10 @@ pub trait HashToField<F: Field>: Sized {
     ///
     /// * `domain` - bytes that get concatenated with the `msg` during hashing, in order to separate potentially interfering instantiations of the hasher.
     /// * `count` - number of elements in field `F` to output.
-    fn new_hash_to_field(domain: &[u8], count: usize) -> Result<Self, HashToCurveError>;
+    fn new_hash_to_field(domain: &[u8]) -> Result<Self, HashToCurveError>;
```
Pratyush, line 23: We can just rename this to `new`, no? Ditto for `new_map_to_curve` above.

ec/src/hashing/tests/mod.rs
```diff
@@ -343,11 +348,10 @@ impl WBParams for TestWBF127MapToCurveParams {
 /// and make simple hash
 #[test]
 fn hash_arbitary_string_to_curve_wb() {
-    use blake2::Blake2bVar;
-
+    use sha2::Sha256;
     let test_wb_to_curve_hasher = MapToCurveBasedHasher::<
         GroupAffine<TestWBF127MapToCurveParams>,
-        DefaultFieldHasher<Blake2bVar, 128>,
+        DefaultFieldHasher<Sha256, 128>,
```
Pratyush, line 354:
```suggestion
        DefaultFieldHasher<sha256::Sha256, 128>,
```

ec/src/hashing/tests/mod.rs
```diff
@@ -343,11 +348,10 @@ impl WBParams for TestWBF127MapToCurveParams {
 /// and make simple hash
 #[test]
 fn hash_arbitary_string_to_curve_wb() {
-    use blake2::Blake2bVar;
-
+    use sha2::Sha256;
```
Pratyush, line 351:
```suggestion
```

ec/src/hashing/tests/mod.rs
```diff
@@ -120,11 +125,11 @@ fn checking_the_hashing_parameters() {
 /// simple hash
 #[test]
 fn hash_arbitary_string_to_curve_swu() {
-    use blake2::Blake2bVar;
+    use sha2::Sha256;
```
Pratyush, line 128:
```suggestion
```

ec/src/hashing/tests/mod.rs
```diff
@@ -120,11 +125,11 @@ fn checking_the_hashing_parameters() {
...
 
     let test_swu_to_curve_hasher = MapToCurveBasedHasher::<
         GroupAffine<TestSWUMapToCurveParams>,
-        DefaultFieldHasher<Blake2bVar, 128>,
+        DefaultFieldHasher<Sha256, 128>,
```
Pratyush, line 132:
```suggestion
        DefaultFieldHasher<sha256::Sha256, 128>,
```

ec/src/hashing/field_hashers/mod.rs
```diff
@@ -1,106 +1,106 @@
...
+        let m = F::extension_degree() as usize;
+
+        // The user imposes a `count` of elements of F_p^m to output per input msg,
+        // each field element comprising `m` BasePrimeField elements.
+        let len_in_bytes = count * m * self.len_per_base_elem;
+        let uniform_bytes = self.expander.expand(message, len_in_bytes);
+
+        let mut output: Vec<F> = Vec::with_capacity(count);
+        for i in 0..count {
+            let mut base_prime_field_elems: Vec<F::BasePrimeField> = Vec::new();
```
Pratyush, line 63:
```suggestion
            let mut base_prime_field_elems: Vec<F::BasePrimeField> = Vec::with_capacity(m);
```

ec/src/hashing/field_hashers/mod.rs
mmagician: Yeah, probably makes sense. Will adapt

Pratyush:
This looks quite good! As a suggestion, I think we should move the field_hashing module to `ark-ff`, since it makes more sense there, and has nothing to do with curves. What do you think?

mmagician:
Don't have anything against that. There was an idea, at some point, to have each curve define its own hash2curve implementation(s) (perhaps as a Trait that curves would implement directly), but it might not be ideal if you wanna have one curve be compatible with multiple protocols which all use a different h2c algorithm. But in its current state, sure, might make more sense under `ff`.

ec/src/hashing/field_hashers/mod.rs
```diff
@@ -1,106 +1,106 @@
...
+            for j in 0..m {
+                let elm_offset = self.len_per_base_elem * (j + i * m);
+                let val: F::BasePrimeField = F::BasePrimeField::from_be_bytes_mod_order(
+                    &uniform_bytes[elm_offset..(elm_offset + self.len_per_base_elem)],
+                );
+                base_prime_field_elems.push(val);
+            }
+            let f: F = F::from_base_prime_field_elems(&base_prime_field_elems).unwrap();
+            output.push(f);
+        }
```
Pratyush, line 73:
```suggestion
        let mut output: Vec<F> = Vec::with_capacity(count);
        let mut base_prime_field_elems: Vec<F::BasePrimeField> = Vec::with_capacity(m);
        for i in 0..count {
            base_prime_field_elems.clear();
            for j in 0..m {
                let elm_offset = self.len_per_base_elem * (j + i * m);
                let val: F::BasePrimeField = F::BasePrimeField::from_be_bytes_mod_order(
                    &uniform_bytes[elm_offset..(elm_offset + self.len_per_base_elem)],
                );
                base_prime_field_elems.push(val);
            }
            let f: F = F::from_base_prime_field_elems(&base_prime_field_elems).unwrap();
            output.push(f);
        }
```

Pratyush:
> Don't have anything against that. There was an idea, at some point, to have each curve define its own hash2curve implementation(s) (perhaps as a Trait that curves would implement directly), but it might not be ideal if you wanna have one curve be compatible with multiple protocols which all use a different h2c algorithm. But in its current state, sure, might make more sense under `ff`.

I think that's not incompatible with the suggestion to move the field hashing to `ark-ff`? Or am I missing something?

mmagician:
Right, it wouldn't be compatible, but in the end I think that by NOT going down the route of curves defining their own h2c, we get more flexibility (i.e. with having a dedicated Hasher struct like now). I was just generally discussing an alternative approach.

mmagician:
I just realised you were talking about field hashing only! In my head they always went together but sure, makes sense that field hashing goes to ff, and curve maps can stay in ec. Sorry for the confusion

mmagician:
@Pratyush I've addressed all the points from feedback - that was very valuable and I think makes the interface better!
Aside from that, I've introduced three minor changes:
1. Since the hash-to-field doesn't fail due to the implementation (could return no field elements at worst), when moving it to `ff` crate I've gotten rid of some HashToCurveErrors and instead of putting them in hash-to-field mod instead, I've changed the interface to not return the `Result` but rather the underlying data structure directly (i.e. `Vec<F>`)
2. (temporarily) removed the unit test for `gen_len_per_elem` [here](https://github.com/arkworks-rs/algebra/pull/405/commits/e8f91d4e6b3a0421c0891e012e6ea5eecfd1597f). I've no clue why, but after the refactor the compiler doesn't recognize `Fq` (from bls12-381 test-curves) as a `Field` anymore. If you or @drskalman have an idea of why/how to fix this, by all means please contribute. Otherwise, as much as I'd hate removing a unit test, this particular one is simple enough that I'd be williing to drop it in order to move ahead with the PR.
3. Since I've added `serde` into `ff` crate as a result of the refactor, I had to give [this](https://github.com/mmagician/algebra/commit/ac2aa67ebf17d12e8f5372c34d4cdce9c8b072da#diff-71d865cbfbf26318ff6a2805c02bcdb77630262eda2bf48e593880a03c8cadfbL321) an explicit type, or else it was ambigious whether to use PartialEq from `core` of from `serde`

ff/src/fields/field_hashers/mod.rs
```diff
@@ -0,0 +1,106 @@
...
+use self::expander::ExpanderXmd;
+
+/// Trait for hashing messages to field elements.
+pub trait HashToField<F: Field>: Sized {
+    /// Initialises a new hash-to-field helper struct.
+    ///
+    /// # Arguments
+    ///
+    /// * `domain` - bytes that get concatenated with the `msg` during hashing, in order to separate potentially interfering instantiations of the hasher.
+    /// * `count` - number of elements in field `F` to output.
```
Pratyush, line 18:
```suggestion
```

PR #417, opened 2022-05-12
Remove `ToBytes` and `FromBytes`

### LINKED ISSUE #390: Projective Coordinates Leak
I know projective curves implement canonical de/serialize traits via affine conversions, so the to/from bytes traits exists only for internal usage, but their projective implementations could easily be miss-used:  https://github.com/arkworks-rs/algebra/blob/master/ec/src/models/short_weierstrass_jacobian.rs#L348-L365  

I'd suggest a comment on the from/ to bytes traits referencing that projective coordinates leak secrets and reference  https://www.researchgate.net/publication/221348015_Projective_Coordinates_Leak / https://eprint.iacr.org/2003/191 or maybe https://eprint.iacr.org/2020/432 

It could be relevant in trusted setups too, not just secret keys.

Pratyush:
## Description

As per title. These traits are insecure, often incorrect, and have been superseded by `CanonicalSerialize` and `CanonicalDeserialize`. Closes #390 

---

write a little note why.

- [ ] Wrote unit tests


mmagician (approved):
The CI job will complain for now, until corresponding fixes are made afterwards in curves, so should be good to go 👍🏼

weikengchen:
A reason why we previously keep them is that they are, at least expected, to match the constraint counterparts, `ToBytesGadget`, which would be useful for certain serialization.

What do you think?

Pratyush:
I think for usage in constraint world, we need a `to_bits` variant anyway, as that is more efficient in circuit. Another option is to just bite the bullet and use canonical serialization inside circuits. So, e.g., we make a `CanonicalSerializeGadget` trait, and use `serialize_uncompressed` in circuit.

PR #418, opened 2022-05-13
Add `sum_of_products`

### LINKED ISSUE #121: Field Trait: Consider adding static linear combination operation
SNARK Protocols (in particular verifiers) often have expensive linear combinations they have to evaluate. The most expensive term in a field multiplication is the modular reduction. However, if we're doing a linear combination, we should be able to batch the modular reductions by adding all of the products in their un-reduced forms.

e.g. if we have an M limb field, and we want to compute `c = sum_{i in 0..N} a_i * b_i`, we can compute this as (pseudocode ignoring montgomery reductions):
```
let mut unreduced_c: 2M + 1 limb bigint;
for i in 0..N {
    let term: 2M limb bigint = unreduced_product(a_i, b_i);
    unreduced_c += term;
}
let c = modular_reduce(c);
c
```
We're guaranteed unreduced_c won't overflow if N < 2^64, since it has one extra limb.

To handle that we're in montgomery form, as a first method, we could make `modular_reduce` just reduce by `p` and then multiply by `R'` at the end. I think its possible to adapt REDC to combine the final multiplication by `R'` into the reduction, but the speedup of that should be marginal, and likely has much higher code complexity.

This could potentially be a significant speedup.

Pratyush:
## Description

Adapts https://github.com/zkcrypto/bls12_381/pull/84 to `ark-ff`. Resolves #121.

I believe the code in `bls12_381` has a bug wherein it does not work for any arbitrary number of inputs (additions overflow the allocated space). So I've modified it to chunk up the inputs into smaller batches where we *can* guarantee that we won't have overflow issues.

I've also used the new method in short weierstrass point additions. This shaves off ~20ns for additions:

Old:
```
test bls12_381::g1::add_assign         ... bench:         726 ns/iter (+/- 28)
test bls12_381::g1::add_assign_mixed   ... bench:         546 ns/iter (+/- 13)
```

New:
```
test bls12_381::g1::add_assign         ... bench:         704 ns/iter (+/- 33)
test bls12_381::g1::add_assign_mixed   ... bench:         526 ns/iter (+/- 18)
```
---

write a little note why.

- [x] Wrote unit tests

### DIFF
--- a/ec/src/models/short_weierstrass.rs
+++ b/ec/src/models/short_weierstrass.rs
@@ -584,7 +584,7 @@ impl<P: SWCurveConfig> ProjectiveCurve for Projective<P> {
             self.z.double_in_place();
 
             // X3 = F-2*D
-            self.x = f - &d - &d;
+            self.x = f - &d.double();
 
             // Y3 = E*(D-X3)-8*C
             self.y = (d - &self.x) * &e - &*c.double_in_place().double_in_place().double_in_place();
@@ -663,7 +663,7 @@ impl<P: SWCurveConfig> ProjectiveCurve for Projective<P> {
                     i.double_in_place().double_in_place();
 
                     // J = H*I
-                    let mut j = h * &i;
+                    let j = h * &i;
 
                     // r = 2*(S2-Y1)
                     let r = (s2 - &self.y).double();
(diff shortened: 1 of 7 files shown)


Pratyush:
Heh @ValarDragon currently we have a weird bug where the results are off-by-one. If you have any insight that would be helpful =)

weikengchen (approved):

PR #420, opened 2022-05-24
Introduce an explicit clear-cofactor method

mmagician:
There are faster ways to clear the cofactor that multiplying by `h`, see e.g. https://eprint.iacr.org/2021/1130.pdf.
In this PR I expose a method `clear_cofactor` on `AffineCurve`, which SW/TE-ModelParameters implement with a default `mul_by_cofactor()` call, but which allows upstream implementers of these traits to override with a more efficient implementation.

As an example, I've added a concrete speedup for bls12-381 G1 (see [companion PR in curves](https://github.com/arkworks-rs/curves/pull/103)) which employs multiplication by a smaller scalar than the cofactor itself.
## Description

related to https://github.com/arkworks-rs/algebra/issues/308

---

write a little note why.

- [x] Wrote unit tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -46,6 +46,7 @@
 - [\#386](https://github.com/arkworks-rs/algebra/pull/386) (`ark-ff-macros`, `ark-ff`) Add a macro to derive `MontConfig`.
 - [\#396](https://github.com/arkworks-rs/algebra/pull/396) (`ark-ec`) Add a default `mul` function to `{TE,SW}ModelParameters` trait definition.
 - [\#397](https://github.com/arkworks-rs/algebra/pull/397) (`ark-ec`) Add `HashMapPippenger` to variable-base MSM.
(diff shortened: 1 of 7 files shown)


Pratyush:
The effect of these methods is to `mul_by_cofactor`, right? I.e., if I have a group element `g` that's already in the prime order subgroup, does this assertion hold: `assert_eq!(g.clear_cofactor(), g.mul_by_cofactor())`?

mmagician:
Not necessarily. AFAIU the point of clearing the cofactor is to return a point in the prime order subgroup, but different techniques for clearing will return different points with the same order.
So if there's a non-default `clear_cofactor` implementation (like [here](https://github.com/mmagician/algebra/blob/dbd429f7375b3a34d6a2a6c8727d280581cde59a/test-curves/src/bls12_381/g1.rs#L51)),  the above assertion won't hold, just as `assert_eq!(g, g.mul_by_cofactor())` won't hold now.

Pratyush (approved):

PR #387, opened 2022-02-07
Unify SquareRootField and Field

solomonjoseph:
## Description

closes: #XXXX

---

Before we can merge this PR, please make sure that all the following items have been
checked off. If any of the checklist items are not applicable, please leave them but
write a little note why.

- [ ] Targeted PR against correct branch (master)
- [ ] Linked to GitHub issue with discussion and accepted design OR have an explanation in the PR that describes this work.
- [ ] Wrote unit tests
- [ ] Updated relevant documentation in the code
- [ ] Added a relevant changelog entry to the `Pending` section in `CHANGELOG.md`
- [ ] Re-reviewed `Files changed` in the GitHub PR explorer

PR #422, opened 2022-05-25
Remove `SquareRootField`, and move functionality to `Field`

mmagician:
## Description

Rather than implementing optimised algorithms for each case from the start, let's begin with Tonelli-Shanks, which should work in each instance, then benchmark that and continue to implement optimisations for each case one-by-one. While not fastest from the start, it already unifies the `SqrtField` and `Field` traits and makes using of `sqrt()` method cleaner.

In this draft, I propose moving the following to `CubicExtField` (prev. only on `Fp3`):
- `TWO_ADICITY`
- `TRACE_MINUS_ONE_DIV_TWO`
- `QUADRATIC_NONRESIDUE_TO_T`

This puts a small overhead on implementers of `CubicExtField` which are not `Fp3` (e.g. [`Fq6`](https://github.com/arkworks-rs/curves/blob/master/bls12_381/src/fields/fq6.rs)), requiring these to define the above constants.

TODOs:
- [ ] verify the approach (based on feedback)
- [ ] implement the above constants for test-curves
- [ ] implement the above for [all curves](https://github.com/arkworks-rs/curves/search?q=%22impl+Fp6Config%22)
- [ ] (?) generate the constants like done [here](https://github.com/arkworks-rs/algebra/blob/master/ff/src/fields/models/fp/mod.rs#L362)

Credit to @solomonjoseph for the initial work on this.

closes: #387 

---

write a little note why.

- [ ] Wrote unit tests


ff/src/fields/models/quadratic_extension.rs
```diff
@@ -446,7 +411,7 @@ where
         two_inv.div2();
 
         let two_inv = P::BasePrimeField::from(two_inv);
-        let two_inv = P::BaseField::from(two_inv);
+        let two_inv = P::BaseField::from_base_prime_field_elems(&[two_inv]).unwrap();
```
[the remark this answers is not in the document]
mmagician: The calling method, `sqrt`, was previously on a `SquareRootField` trait, and in the implementation for quadratic extension we were restricting `P: QuadExtConfig` to have:
```
where P::BaseField: From<P::BasePrimeField>
```

Now that the method is on `Field` trait, we don't have these bounds, so I need to explicitly construct a base field element from `BasePrimeField` elements.

ff/src/fields/mod.rs
```diff
@@ -486,6 +485,31 @@ impl LegendreSymbol {
     }
 }
 
+pub enum SqrtPrecomputation<F: Field> {
```
Pratyush, line 488: Let's mark this enum as `#[non_exhaustive]`, so that in the future when we add more variants, it won't cause breakage.

ff/src/fields/models/fp6_3over2.rs
```diff
@@ -24,6 +24,8 @@ impl<P: Fp6Config> CubicExtConfig for Fp6ConfigWrapper<P> {
     type BaseField = Fp2<P::Fp2Config>;
     type FrobCoeff = Fp2<P::Fp2Config>;
 
+    const PRECOMP: Option<SqrtPrecomputation<CubicExtField<Self>>> = None;
```
Pratyush, line 27: Let's change this to `const PRECOMP: ... = P::PRECOMP;`, and add a `const PRECOMP: Option<SqrtPrecomputation<CubicExtField<Self>>> = None;` to `Fp6Config`. This way, if someone downstream wishes to specify a  sqrt algorithm, they can do it easily.

This would also allow us to write `Fp6<Self>` instead of `CubicExtField<Self>`

Pratyush: Also, let's call this `SQRT_PRECOMP` like elsewhere.

ff/src/fields/models/quadratic_extension.rs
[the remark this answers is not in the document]
Pratyush: Let's add a `BaseField: From<BasePrimeField>` bound in the config struct; maybe as a different PR?

Pratyush (approved):

PR #427, opened 2022-06-10
Add katex support in documentation.

mmaker:
This commit adds latex support within rustdoc for arkworks algebra.
In order to add latex code in textmode, use delimiters `$` and `$`.
In order to add latex code in displaymode, use delimiters `$$` and `$$`.
To compile the documentation, use `cargo rustdoc`.

This is achieved via the option `--html-in-header` set via
- the environment variable `RUSTDOCFLAGS`, or
- the variable `build.rustdocflags` of `.cargo/config.toml`
I opted for the latter making it a default compilation flag to use with
rustdoc.

IMPORTANT: for some reason (independent from us) `cargo rustdoc` fails
on rust nightly and this is due to a failure concerning `cfg-if` and
`rand-chacha`. I did not investigate the reason why this happened.

### DIFF
--- a/doc/katex-header.html
+++ b/doc/katex-header.html
@@ -1,0 +1,31 @@
+<link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/katex@0.16.0/dist/katex.min.css" integrity="sha384-Xi8rHCmBmhbuyyhbI88391ZKP2dmfnOl4rT9ZfRI7mLTdk1wblIUnrIq35nqwEvC" crossorigin="anonymous">
+
+<!-- The loading of KaTeX is deferred to speed up page rendering -->
+<script defer src="https://cdn.jsdelivr.net/npm/katex@0.16.0/dist/katex.min.js" integrity="sha384-X/XCfMm41VSsqRNQgDerQczD69XqmjOOOwYQvr/uuC+j4OPoNhVgjdGFwhvN02Ja" crossorigin="anonymous"></script>
+
+<!-- To automatically render math in text elements, include the auto-render extension: -->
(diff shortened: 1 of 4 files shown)


mmaker:
Great, right now doctests (`cargo test`) fail because the documentation path is incorrect: 
instead of `doc/katex-header.html` it expects `../doc/katex-header.html`. 
However, `cargo rustdoc` expects `doc/katex-header.html` instead of `../doc/katex-header.html`.

We have two options: 
- we YOLO it, and let the user set the option `--html-in-header` or the env variable `RUSTDOCFLAGS` manually.
- we create a `makefile` inside *each package* with the following content:
```
doc:
	cargo rustdoc -- --html-in-header doc/katex-header.html
```
The latter is what's being used in `curve25519-dalek`, but would pollute the build environment :(
This affects only local compilation and shouldn't concern upload to docs.rs, which is already dealt with.

Pratyush:
@mmaker would using the solution in this issue work? https://github.com/rust-lang/cargo/issues/9427

We would have to switch to running doctests on nightly only, but that's fine IMO

mmaker:
Ok after some investigation: 
```rustdocflags = "-Zdoctest-in-workspace --html-in-header doc/katex-header.html"```
clearly does not work because it's not a rustdoc option. After some hacking around, the following option is what we want:
```[unstable]
doctest-in-workspace = true```

Now `.cargo/config.html` looks like the following:
```
# Add support for katex, and work around rust-lang issue #8097.
# <https://github.com/rust-lang/cargo/issues/8097>
[target.'cfg(nightly)']
rustdocflags = "--html-in-header doc/katex-header.html"


# Without this line, one among `cargo test --doc` and `cargo rustdoc`
# will fail due to the relative path in the `--html-in-header` option for rustdoc.
[unstable]
doctest-in-workspace = true```

Note the `target.cfg(nightly)`: without it, on rust stable we'd have `cargo test --doc` fail because of the relative path issue...
However, now we have nother problem: `target.<cfg>` sections only define `rustflags` and not `rustdocflags`... therefore we're back to square one.

mmaker:
Force-pushing this pull request with a makefile-based workaround (similarly to what's done in `curve25519-dalek`...

PR #443, opened 2022-07-18
Improve ergonomics of scalar field multiplication.

mmaker:
Hello! This pull request aims to improve the ergonomics of arkworks scalar multiplication by eradicating the plague that `into_bigint()` has become without sacrificing efficiency. 

This change will affect `Affine::mul` and `Projective::mul`: instead of accepting as input bits packed into limbs (`AsRef<[u64]>` or `PrimeField::BigInt`), we accept a`ScalarField` element. Another function `mul_bigint` is made available for implementations where conversion for Montgomery form is not necessary.

We believe this will greately improve ergonomicity of arkworks, and since this change will probably affect also other crates in the arkworks ecosystem we'd like to know what's the best way to proceed aligning also the other repositories.

This is partially related (but doesn't conflict) to #294.


asn-d6:
Some questions that appeared while coding the PR:
- Do we need `#[muse_use]`? If yes, do we also want it in the projective curve?
- Do we like the name `mul_bigint()`? It actually multiplies by a ref to u64, so maybe `mul_limbs()` could be an alternative name.
- Should we move forward with implementing the `Mul` trait for the curves, so that we can do `B = b*A` in arkworks? I know that in the past there were concerns about `Mul` allowing scalar multiplication from the right (i.e. `B = A*b`) which is not so nice, but maybe these concerns are not as important as the potential ergonomic benefits?

weikengchen:
I would say it is very likely that we need to keep the old one as well.

And `Mul` for different ways is probably better.

mmaker:
@weikengchen what does it mean "the old one"? Mul by bigints? also for implementing the trait `Mul`, this will be a separate pull request. See https://github.com/mmaker/arkworks-algebra/tree/feature/mul-operand-affine

weikengchen:
Got it. Just found out this

```
    + for<'a> Mul<&'a Self::ScalarField, Output = Self>
```

Then it makes sense.

Pratyush (approved):

Issue #97, opened 2020-11-27
Split up affine and projective files


jon-chuang:
They are too long

ValarDragon:
We're considering re-architecting the curve interfaces entirely (see #41). Considering that proposed API, I think it does make sense to separate out the canonical repr into its own file. (Though the Canonical Repr should hopefully be smaller than it is right now)

PR #447, opened 2022-08-03
Refactor curve and pairing traits

Pratyush:
## Description

closes:

* #294 
* #268 
* #41 
* #27
* #97

---

write a little note why.

- [ ] Wrote unit tests


Pratyush:
This is blocked on merging #446

weikengchen (approved):

ec/src/pairing.rs
```diff
@@ -0,0 +1,331 @@
...
+        + for<'a> From<&'a Self::G2Affine>
+        + From<Self::G2>
+        + From<Self::G2Affine>;
+
+    /// The extension field that hosts the target group of the pairing.
+    type TargetField: CyclotomicMultSubgroup;
+
+    /// Computes the product of Miller loops for some number of (G1, G2) pairs.
+    #[must_use]
+    fn multi_miller_loop(
```
mmagician, line 77: I think I understand the reason for naming it `multi_miller_loop` (base & parallel configuration squashed into one?), but it could be confusing - it almost sounds like multiple complete runs of Miller loop are computed. Why not stick to the old `miller_loop`?

ec/src/pairing.rs
```diff
@@ -0,0 +1,331 @@
...
+    Copy(bound = "P: Pairing"),
+    Clone(bound = "P: Pairing"),
+    Debug(bound = "P: Pairing"),
+    PartialEq(bound = "P: Pairing"),
+    Eq(bound = "P: Pairing"),
+    PartialOrd(bound = "P: Pairing"),
+    Ord(bound = "P: Pairing")
+)]
+#[must_use]
+pub struct MillerLoopOutput<P: Pairing>(pub P::TargetField);
```
mmagician, line 311: Are you wrapping `TargetField` in another struct just to be able to avoid Rust's orphan rule, or is there an underlying API design decision?
It seems somewhat awkward that we're mixing a very high-level API names  (`MillerLoopOutput`) with names of algebraic structures in the miller loop method signature. For consistency, I'd suggest either:
```
    fn multi_miller_loop(
        a: impl IntoIterator<Item = impl Into<Self::MillerLoopInputA>>,
        b: impl IntoIterator<Item = impl Into<Self::MillerLoopInputB>>,
    ) -> MillerLoopOutput<Self>;
```
or something along the lines of:
```
    fn multi_miller_loop(
        a: ...G1Prepared...,
        b: ...G2Prepared...,
    ) -> FqkEqClass;
```
of which I'd go for the latter, though maybe `FqkEqClass` is a bit too explicit? I don't think any of our cuves implement the Weil pairing though, so that should always be the output of the miller loop.

Same for `PairingOutput`, maybe `Gt` instead?

ec/src/scalar_mul/variable_base/mod.rs
```diff
@@ -1,73 +1,49 @@
...
-    + for<'a> AddAssign<&'a Self>
-    + for<'a> Add<&'a Self, Output = Self>
-{
-    type MSMBase: Sync + Copy;
+use super::ScalarMul;
 
-    type Scalar: PrimeField;
-
-    #[doc(hidden)]
-    fn _double_in_place(&mut self) -> &mut Self;
```
mmagician, line 32: nice!

ec/src/models/bn/g1.rs
```diff
@@ -23,14 +22,32 @@ impl<P: BnParameters> From<G1Affine<P>> for G1Prepared<P> {
     }
 }
 
+impl<P: BnParameters> From<G1Projective<P>> for G1Prepared<P> {
```
mmagician, line 25: this pattern of `impl<> From<> for G1Prepared<>` repeats in each model - could we maybe have a macro to generate this impls?

ec/src/pairing.rs
weikengchen: I think it is because the implementation here does not strictly follow what a Miller loop should be.

We often say Miller loop as f_{T, P)(Q) or its variants. In this case, we are actually doing the following, rather than a common Miller loop.
f_{T, P_1}(Q_1) \cdot f_{T, P_2}(Q_2) \cdot ...

The difference will be evident when one does a large number of Miller loops, likely in the case of **batch verification of zero-knowledge proofs** in order to improve the TPS, because of the parallelization support here.

ec/src/models/bls12/g2.rs
```diff
@@ -24,11 +21,15 @@ pub type G2Projective<P> = Projective<<P as Bls12Parameters>::G2Parameters>;
...
+    pub ell_coeffs: Vec<EllCoeff<P>>,
     pub infinity: bool,
 }
 
-pub(crate) type EllCoeff<F> = (F, F, F);
+pub(crate) type EllCoeff<P> = (
+    Fp2<<P as Bls12Parameters>::Fp2Config>,
+    Fp2<<P as Bls12Parameters>::Fp2Config>,
+    Fp2<<P as Bls12Parameters>::Fp2Config>,
+);
```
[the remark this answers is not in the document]
weikengchen: I think this is a small change that we may just skip---we now have to be selective.

The main issue here is probably the naming of this local type. `F2` could be confusing, and maybe the easiest way is just to not create this local type since it is a one-time use.

ec/src/pairing.rs
weikengchen: I think `FqkEqClass` is unlikely going to be used in another place, and for one-time use, I prefer to dismiss it.

The output is an element in `Fqk`, we just need to do a final exp after that. Note that we do not always use it just for equality check. Pairing could also be used, without a equality check in the middle, for **group commitments**.

For this reason, I would not recommend `FqkEqClass`.

weikengchen: By the way, I think dedicating a struct for this is good `MillerLoopOutput` as it prevents developers from forgetting to do final exp.

For this purpose, I recommend we drop the derivation of `PartialEq, Eq, PartialOrd, Ord`.

ec/src/models/short_weierstrass/group.rs
```diff
@@ -0,0 +1,578 @@
...
+
+impl<P: SWCurveConfig> Default for Projective<P> {
+    #[inline]
+    fn default() -> Self {
+        Self::zero()
+    }
+}
+
+impl<P: SWCurveConfig> Projective<P> {
+    /// Construct a new group element without checking whether the coordinates
```
weikengchen, line 113:
```suggestion
    /// Constructs a new group element without checking whether the coordinates
```

ec/src/models/short_weierstrass/group.rs
```diff
@@ -0,0 +1,578 @@
...
+    /// specify a point in the subgroup.
+    pub const fn new_unchecked(x: P::BaseField, y: P::BaseField, z: P::BaseField) -> Self {
+        Self { x, y, z }
+    }
+
+    /// Construct a new group element in a way while enforcing that points are in
```
weikengchen, line 119:
```suggestion
    /// Constructs a new group element in a way while enforcing that points are in
```

weikengchen (approved):

ec/src/models/bn/g1.rs
weikengchen: Good points. And it seems that only BN, BLS, BW also implements `is_zero`, but not MNT46. Any reason why?

Pratyush: It depends entirely on what the corresponding pairing implementations require; the MNT46 impls don't seem to require that method.

ec/src/pairing.rs
Pratyush: I'd be fine with `MillerLoopInput1`, but that it seems a bit less informative than `G1Prepared`.

The reason to have `MillerLoopOutput` be not just `Self::TargetField` is to make it explicit that it's not meant to be used by itself.

ec/src/pairing.rs
Pratyush: we could also call this `batch_miller_loop` and the corresponding pairing method `batch_pairing`. Does that sound better?

weikengchen: Batching usually refers to **checking** a bunch, likely involving a randomizer. In this Miller loop we just add them together.

Pratyush:
I think this PR is good to merge. I have some more things I'd like to add, but let's do that in a followup PR.

weikengchen:
I think we can merge. I probably will fix the downstream in a few days.

PR #450, opened 2022-08-09
The sign trick for multi-scalar multiplication

mmaker:
Hello! @horacepan helped me move in an optimized algorithm for multi-scalar multiplications with`ScalarField` elements.


Pratyush:
Are there any benchmarks here? It should be a non-trivial speedup, right?

Issue #456, opened 2022-08-20
DenseMultilinearExtension swap first and last variable


montekki:
```rust
    let mut poly = DenseMultilinearExtension::rand(4, &mut rng);
    poly.relabel_inplace(0, 3, 1)
```

relabeling first and last fails, the second branch of check in this assert does not look right:

https://github.com/arkworks-rs/algebra/blob/6bf24dd0dc6e97ce25b11815c7ac38a2aeb212eb/poly/src/evaluations/multivariate/multilinear/dense.rs#L60-L63

PR #457, opened 2022-08-20
fix assert in relabel_inplace

montekki:
## Description

It looks like relabeling first and last values in the `DenseMultilinearExtension` triggers a wrong assert check.
closes: #456 

---


poly/src/evaluations/multivariate/multilinear/dense.rs
```diff
@@ -58,7 +58,7 @@ impl<F: Field> DenseMultilinearExtension<F> {
             ark_std::mem::swap(&mut a, &mut b);
         }
         assert!(
-            a + k < self.num_vars && b + k < self.num_vars,
+            a + k < self.num_vars && b + k <= self.num_vars,
```
mmagician, line 61: That looks good. A small optimisation:
if we move the `a==b` check above the first assert, then we can even skip the first part of the assert (`a + k < self.num_vars`), since at that point `b` would be strictly larger than `a`, and so `b + k <= self.num_vars` would be a sufficient condition.

mmagician: https://github.com/montekki/algebra/pull/1

montekki: Thanks!

weikengchen (approved):

weikengchen:
I checked the issue. Yes, the condition should actually check b + k - 1 < num_elements, which equals to b + k <= num_elements. We do not need to place a condition for a because a <= b.

PR #461, opened 2022-08-28
Allow running BLS12-381 curve test in the field-only mode

weikengchen:
## Description

At this moment, if one does the following, which is the case of ark-snark (!)
```
[dev-dependencies]
ark-test-curves = { version = "^0.3.0", default-features = false, features = [ "bls12_381_scalar_field" ] }
```

It would fail because the conditional compilation for bls12_381_scalar_field is not complete. It would report that several modules are missing. 

This PR fixes so.

---

write a little note why.

N/A:
- [ ] Wrote unit tests

### DIFF
--- a/test-curves/src/bls12_381/mod.rs
+++ b/test-curves/src/bls12_381/mod.rs
@@ -20,11 +20,11 @@ pub mod g1_swu_iso;
 pub mod g2;
 #[cfg(feature = "bls12_381_curve")]
 pub mod g2_swu_iso;
-
 #[cfg(feature = "bls12_381_curve")]
 pub use {fq::*, fq12::*, fq2::*, fq6::*, g1::*, g1_swu_iso::*, g2::*, g2_swu_iso::*};
 
 #[cfg(test)]
+#[cfg(feature = "bls12_381_curve")]
 mod tests;
 
 #[cfg(feature = "bls12_381_curve")]
--- a/test-curves/src/bls12_381/mod.rs
+++ b/test-curves/src/bls12_381/mod.rs
@@ -4,13 +4,13 @@ pub use fr::*;
 #[cfg(feature = "bls12_381_curve")]
 use ark_ec::bls12::{Bls12, Bls12Parameters, TwistType};
 
-#[cfg(feature = "bls12_381_scalar_field")]
+#[cfg(feature = "bls12_381_curve")]
 pub mod fq;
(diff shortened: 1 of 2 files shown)


test-curves/src/bls12_381/mod.rs
```diff
@@ -22,12 +24,16 @@ pub mod g2_swu_iso;
...
 mod tests;
 
+#[cfg(feature = "bls12_381_curve")]
 pub type Bls12_381 = Bls12<Parameters>;
 
+#[cfg(feature = "bls12_381_curve")]
 pub struct Parameters;
 
+#[cfg(feature = "bls12_381_curve")]
 impl Bls12Parameters for Parameters {
```
weikengchen, line 37: The main reason for this change is that `Parameters` requires group definitions.

test-curves/src/bls12_381/tests.rs
```diff
@@ -1,16 +1,13 @@
-#![allow(unused_imports)]
-use ark_ec::{
-    models::short_weierstrass::SWCurveConfig, AffineCurve, PairingEngine, ProjectiveCurve,
-};
+use ark_ec::AffineCurve;
 use ark_ff::{Field, One, UniformRand, Zero};
 
 use crate::bls12_381::{g1, Fq, Fq2, Fq6, FqConfig, Fr, FrConfig, G1Affine, G1Projective};
+
```
weikengchen, line 5:
```suggestion
```

test-curves/src/bls12_381/mod.rs
```diff
@@ -22,12 +24,16 @@ pub mod g2_swu_iso;
 pub use {fq::*, fq12::*, fq2::*, fq6::*, g1::*, g1_swu_iso::*, g2::*, g2_swu_iso::*};
 
 #[cfg(test)]
+#[cfg(feature = "bls12_381_curve")]
```
weikengchen, line 27: Since Fq is only loaded when `bls12_381_curve`, this effectively means that all the tests would require `bls12_381_curve`.

weikengchen:
Note: this should not be merged before https://github.com/arkworks-rs/algebra/pull/447

because #447 will have a conflict with the code here, and it is better at this moment not to block #447 or add new conflicts in any way.

Pratyush:
I've included the fix for this in #463. If you could check and confirm that it works, that would be great.

PR #462, opened 2022-08-28
Add canonical serialization for G2Prepared

weikengchen:
## Description

This is due to Groth16's implementation that uses `PreparedVerifyingKey`. Now it no longer has `ToBytes` and there is a preference to do `CanonicalSerialize`.

To derive prepared VK directly, we would like G2Prepared to implement CanonicalSerialize and CanonicalDeserialize.

This, however, adds a restriction. So, it would be desirous to know if we want to place this requirement.

---

write a little note why.

N/A:
- [ ] Wrote unit tests


weikengchen:
-> Still, we should prioritize #447.

weikengchen:
Ready for a review. I think the next step is about r1cs_std.

Pratyush (approved):

Issue #145, opened 2020-12-17
Serialization with modes


weikengchen:
When I was implementing serialization for certain data structures (manually), there appears to be a lot of repeated code that just calls the underlying function differently (e.g., `deserialize` vs `deserialize_uncompressed`).

This is similar to what has happened to `AllocVar` (https://github.com/arkworks-rs/r1cs-std/blob/master/src/alloc.rs#L9), which we eventually introduce modes:
```
/// Describes the mode that a variable should be allocated in within
/// a `ConstraintSystem`.
#[derive(Eq, PartialEq, Ord, PartialOrd, Debug, Copy, Clone)]
pub enum AllocationMode {
    /// Indicate to the `ConstraintSystem` that the high-level variable should
    /// be allocated as a constant. That is, no `Variable`s should be
    /// generated.
    Constant = 0,

    /// Indicate to the `ConstraintSystem` that the high-level variable should
    /// be allocated as a public input to the `ConstraintSystem`.
    Input = 1,

    /// Indicate to the `ConstraintSystem` that the high-level variable should
    /// be allocated as a private witness to the `ConstraintSystem`.
    Witness = 2,
}
```

Pratyush:
See also the discussion here: https://github.com/arkworks-rs/algebra/pull/130#issuecomment-745746326

PR #463, opened 2022-08-30
More flexible and deduplicated serialization

Pratyush:
## Description

closes: #130, #145, #461

depends on #447 

---

write a little note why.

- [ ] Wrote unit tests


ec/src/models/short_weierstrass/affine.rs
```diff
@@ -0,0 +1,420 @@
...
+    }
+}
+
+impl<P: SWCurveConfig> Neg for Affine<P> {
+    type Output = Self;
+
+    /// If `self.is_zero()`, returns `self` (`== Self::zero()`).
+    /// Else, returns `(x, -y)`, where `self = (x, y)`.
+    #[inline]
+    fn neg(mut self) -> Self {
```
weikengchen, line 245: Just a confirmation that we are on the same page that if one does `negq = q.neg()`, an implicit copy happens, and q will remain unchanged, right? @confuseSUN mentioned this to me a few hours ago.

weikengchen (approved):
I went through all the files.

ec/src/models/short_weierstrass/affine.rs
Pratyush: Yes, it’s not a `neg_in_place` (and in fact there’s no such operation)

Pratyush:
@weikengchen There are a few changes here.

* Serialization for Short Weierstrass curves now sets the flags to 0 if the y-coordinate is positive, 1 << 6 if it's a point at infinity, and 1 << 7 if the y-coordinate is negative.
* Serialization for Twisted Edwards curves now sets the flags to 0 if the y-coordinate is positive, and 1 if it's negative. 
I will add this information to the CHANGELOG.

I intend to add a PR that allows users to specify custom serialization/deserialization routines. This is useful, for example, for decoding bls12_381 points, since the standardized format for those curves differs from our standardized format. It would also be useful for users of 0.3 that need to maintain compatibility with the old way of serialization/deserialization.

We should also prepare a migration guide that explains how to go about upgrading to 0.4 once we've finished the refactors.

Pratyush:
I have updated the CHANGELOG now.

PR #467, opened 2022-09-04
Customizable serialization for `SWCurveConfig` and `TECurveConfig`

alexander-zw:
## Description

closes: #308

---

write a little note why.

- [ ] Wrote unit tests

### DIFF
--- a/ec/src/models/short_weierstrass/affine.rs
+++ b/ec/src/models/short_weierstrass/affine.rs
@@ -1,6 +1,5 @@
 use ark_serialize::{
-    CanonicalDeserialize, CanonicalDeserializeWithFlags, CanonicalSerialize,
-    CanonicalSerializeWithFlags, Compress, SerializationError, Valid, Validate,
+    CanonicalDeserialize, CanonicalSerialize, Compress, SerializationError, Valid, Validate,
 };
 use ark_std::{
     borrow::Borrow,
@@ -125,7 +124,7 @@ impl<P: SWCurveConfig> Affine<P> {
     ///
     /// The results are sorted by lexicographical order.
     /// This means that, if `P::BaseField: PrimeField`, the results are sorted as integers.
-    fn get_ys_from_x_unchecked(x: P::BaseField) -> Option<(P::BaseField, P::BaseField)> {
+    pub fn get_ys_from_x_unchecked(x: P::BaseField) -> Option<(P::BaseField, P::BaseField)> {
         // Compute the curve equation x^3 + Ax + B.
         // Since Rust does not optimise away additions with zero, we explicitly check
         // for that case here, and avoid multiplication by `a` if possible.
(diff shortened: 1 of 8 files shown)


ec/src/models/short_weierstrass/affine.rs
```diff
@@ -125,7 +124,7 @@ impl<P: SWCurveConfig> Affine<P> {
     ///
     /// The results are sorted by lexicographical order.
     /// This means that, if `P::BaseField: PrimeField`, the results are sorted as integers.
-    fn get_ys_from_x_unchecked(x: P::BaseField) -> Option<(P::BaseField, P::BaseField)> {
+    pub fn get_ys_from_x_unchecked(x: P::BaseField) -> Option<(P::BaseField, P::BaseField)> {
```
alexander-zw, line 127: Deserialize uses `get_ys_from_x_unchecked`. Should I move `get_ys_from_x_unchecked` and `get_point_from_x_unchecked` to the curve config too? If not, I need to make this public.

ec/src/models/short_weierstrass/affine.rs
```diff
@@ -344,38 +343,18 @@ impl<P: SWCurveConfig> From<Projective<P>> for Affine<P> {
 }
 
 impl<P: SWCurveConfig> CanonicalSerialize for Affine<P> {
-    #[allow(unused_qualifications)]
     #[inline]
     fn serialize_with_mode<W: Write>(
         &self,
-        mut writer: W,
```
alexander-zw, line 351: TIL that you can turn an immutable variable into a mutable one

ec/src/models/short_weierstrass/affine.rs
```diff
@@ -344,38 +343,18 @@ impl<P: SWCurveConfig> From<Projective<P>> for Affine<P> {
 }
 
 impl<P: SWCurveConfig> CanonicalSerialize for Affine<P> {
-    #[allow(unused_qualifications)]
```
alexander-zw, line 347: I'm not sure what this does, but there are no lint errors after removing it. Let me know if it's wrong to remove

ec/src/models/short_weierstrass/affine.rs
```diff
@@ -390,47 +369,12 @@ impl<P: SWCurveConfig> Valid for Affine<P> {
 }
```
alexander-zw, line 369: Should I move `check()` into the curve config too?

ec/src/models/twisted_edwards/mod.rs
```diff
@@ -78,6 +84,63 @@ pub trait TECurveConfig: super::CurveConfig {
...
+        }
+        Ok(point)
+    }
+
+    #[inline]
+    fn serialized_size(compress: Compress) -> usize {
+        let zero = Self::BaseField::zero();
+        match compress {
+            Compress::Yes => zero.serialized_size_with_flags::<TEFlags>(),
+            Compress::No => zero.uncompressed_size() + zero.uncompressed_size(),
```
alexander-zw, line 141: I changed this from `x.uncompressed_size() + y.uncompressed_size()` to `zero.uncompressed_size() + zero.uncompressed_size()`. Is there a difference? Aren't they all the same field? With this change this function is independent of the affine element.

ec/src/models/short_weierstrass/affine.rs
alexander-zw: I dug into the history and it's been here for at least three years, I'm guessing it was kept long after no longer being needed

ec/src/models/twisted_edwards/mod.rs
alexander-zw: The tests passed after this change, though I'm not sure how well serialize size is tested

ec/src/models/short_weierstrass/affine.rs
Pratyush: You can make it `pub`

ec/src/models/short_weierstrass/affine.rs
Pratyush: Yeah, we can remove it if it doesn't trigger any linter errors.

ec/src/models/short_weierstrass/affine.rs
Pratyush: I think for the time being we can keep it outside; for elliptic curves, the checks are basically `is_in_subgroup_assuming_is_on_curve`, and we already specialize that.

ec/src/models/twisted_edwards/mod.rs
Pratyush: It's the same right now, but if it's simpler we can keep the old behaviour.

alexander-zw: The new behavior is simpler. In the new behavior the size doesn't depend on the curve element, in the old behavior it referenced the curve values. So should we keep the behavior in this PR right now?

Pratyush (approved):

PR #469, opened 2022-09-05
Let field tests work with 256-bit curves like secp256k1

weikengchen:
## Description

When implementing secp256k1, we find that the field tests do not work well with curves whose modulus bit numbers are 64 * K.

This PR fixes so.

---

write a little note why.

N/A:
- [ ] Wrote unit tests


serialize/src/lib.rs
```diff
@@ -204,15 +204,15 @@ impl<T: CanonicalSerialize> CanonicalSerializeHashExt for T {}
 
 #[inline]
 pub fn buffer_bit_byte_size(modulus_bits: usize) -> (usize, usize) {
-    let byte_size = buffer_byte_size(modulus_bits);
+    let byte_size = buffer_byte_size(modulus_bits + 1);
```
weikengchen, line 207:
```suggestion
    let byte_size = buffer_byte_size(modulus_bits);
```

test-templates/src/fields.rs
```diff
@@ -54,7 +54,7 @@ macro_rules! __test_field {
                     let buf_size = <$field>::zero().serialized_size(compress);
 
                     let buffer_size =
-                        buffer_bit_byte_size(<$field as Field>::BasePrimeField::MODULUS_BIT_SIZE as usize).1 *
+                        buffer_bit_byte_size(<$field as Field>::BasePrimeField::MODULUS_BIT_SIZE as usize + 1).1 *
```
weikengchen, line 57:
```suggestion
                        buffer_bit_byte_size(<$field as Field>::BasePrimeField::MODULUS_BIT_SIZE as usize).1 *
```

test-templates/src/fields.rs
```diff
@@ -395,7 +403,11 @@ macro_rules! __test_field {
             assert_eq!(BigUint::from(<$field>::MODULUS_MINUS_ONE_DIV_TWO), &modulus_minus_one / 2u32);
             assert_eq!(<$field>::MODULUS_BIT_SIZE as u64, modulus.bits());
             if let Some(SqrtPrecomputation::Case3Mod4 { modulus_plus_one_div_four }) = <$field>::SQRT_PRECOMP {
-                assert_eq!(modulus_plus_one_div_four, &((&modulus + 1u8) / 4u8).to_u64_digits());
+                let mut expected = ((&modulus + 1u8) / 4u8).to_u64_digits();
+                if <$field>::MODULUS_BIT_SIZE as u64 % 64 == 0 {
```
Pratyush, line 407: Why is this necessary? Modulus_plus_one_div_four has always 255 bits at most, and so should fit in 4 limbs

serialize/src/lib.rs
```diff
@@ -212,7 +212,7 @@ pub fn buffer_bit_byte_size(modulus_bits: usize) -> (usize, usize) {
 /// into the number of bytes required to represent it.
 #[inline]
 pub const fn buffer_byte_size(modulus_bits: usize) -> usize {
-    (modulus_bits + 7) / 8
+    (modulus_bits + 1 + 7) / 8
```
Pratyush, line 215: This isn’t necessarily correct; it adds a byte to every representation, no? Can we add a test to ensure that it maintains the old behaviour for smaller moduli?

and is the reason the test is failing because you’re trying to use the existing arithmetic infrastructure for 256 bit moduli?

Pratyush: If you’re trying to use the existing infrastructure for secp256k1, I think a better idea would be to port over the algorithms from here: https://github.com/pornin/crrl/blob/3a3e625d784b4234d53829771fc3e18abab16913/src/backend/w64/modint.rs#L141

that way we introduce the extra limb only when necessary.

weikengchen:
I will come back and cross-check on this.

serialize/src/lib.rs
weikengchen: Let me take a look at this new algorithm---it would be a cleaner fix and is also for the future.

weikengchen:
This PR is no longer the solution---instead of allowing an additional limb, we now generalize the code into supporting full limbs. Closed.

PR #421, opened 2022-05-25
`QUADRATIC_NONRESIDUE` is not used

mmagician:
Either the sqrt algorithm was changed or the variable was never used.

## Description

Slightly simplifies the interface for fp2.

---

Before we can merge this PR, please make sure that all the following items have been
checked off. If any of the checklist items are not applicable, please leave them but
write a little note why.

- [x] Targeted PR against correct branch (master)
- [ ] Linked to GitHub issue with discussion and accepted design OR have an explanation in the PR that describes this work.
- [ ] Wrote unit tests
- [ ] Updated relevant documentation in the code
- [ ] Added a relevant changelog entry to the `Pending` section in `CHANGELOG.md`
- [x] Re-reviewed `Files changed` in the GitHub PR explorer

Issue #276, opened 2021-05-12
Move multiplicative_generator from FFTField to Field


ValarDragon:
Currently `multiplicative_generator` is a method defined on FFTField. We should instead move this to Field, since all fields have a multiplicative generator, not just FFTField's. (Furthermore, there are contexts where one may not have an FFT-friendly field, but still needs the multiplicative generator for it)

This would be a breaking change, that I think would require a change to every curve in curves.

(Thanks to @mikerah for pointing this out in telegram)

Pratyush:
How easy is it to test if an element is a generator? It's not clear that factorizing (p^n - 1) is easy for arbitrary p and n. (p - 1 is a factor of p^n - 1, but there must be other prime factors too.)

daira:
See https://trac.sagemath.org/ticket/31686 for factoring p<sup>n</sup> - 1. Summary: yes it is easy in the cases likely to arise in Arkworks, using the code at https://trac.sagemath.org/ticket/31686#comment:7 .

Pratyush:
That’s perfect, thanks!

pwang00:
Ah yeah, I was going to suggest some possibility of factorization by cyclotomic polynomials since $ x^n - 1 = \Prod_{d | n} Φ_d(x) $, but it seems like the link covers this.  

I'm assuming this hasn't been implemented into arkworks yet?  If not I will start here.

Pratyush:
@pwang00 indeed it hasn't been implemented yet, so you can start there. Please feel free to ask any questions!

PR #472, opened 2022-09-09
Move to `criterion` benchmarks

Pratyush:
## Description

Moves from the old `bencher` based infrastructure to a more accurate `criterion` based infrastructure. This allows us to minimize the time spent in generating material for initializing field and group elements, and also enables us to add better profilers in the future.

---

write a little note why.

- [x] Wrote unit tests


bench-templates/src/macros/field.rs
```diff
@@ -1,542 +1,468 @@
...
+                    let mut i = 0;
+                    b.iter(|| {
+                        i = (i + 1) % SAMPLES;
+                        let j = (i + 1) % SAMPLES;
+                        field_elements_left[i] * field_elements_left[j]
+                            + field_elements_right[j] * field_elements_right[i]
+                    })
+                },
+            );
+            // arithmetic.bench_function(
```
weikengchen, line 146: What is the plan?

Pratyush: Oh oops, good catch. Adding these back now.

bench-templates/src/macros/field.rs
```diff
@@ -1,542 +1,468 @@
...
-            let v: Vec<$f_type> = (0..SAMPLES)
-                .map(|_| {
-                    let mut tmp = $f::rand(&mut rng);
-                    tmp.square_in_place();
-                    tmp
+            let f = (0..SAMPLES)
+                .map(|_| <$F>::rand(&mut rng))
+                .collect::<Vec<_>>();
+            let qrs = f.iter().map(|s| s.square()).collect::<Vec<_>>();
+            let qnrs = f.iter().map(|s| s.square()).collect::<Vec<_>>();
```
weikengchen, line 274: Would `qnrs` be NRS?

bench-templates/src/macros/field.rs
```diff
@@ -1,542 +1,468 @@
...
+                    v1[i] == v2[i]
+                })
             });
         }
 
-        fn into_repr(b: &mut $crate::bencher::Bencher) {
-            const SAMPLES: usize = 1000;
-
-            let mut rng = ark_std::test_rng();
+        // fn into_repr(b: &mut $crate::bencher::Bencher) {
```
weikengchen, line 438: Plan?

weikengchen (approved):

bench-templates/src/macros/field.rs
Pratyush: `qrs` = quadratic residues, and `qnrs` = quadratic non-residues. I think this is in line with our terminology elsewhere.

weikengchen: The problem is that
```
 let qnrs = f.iter().map(|s| s.square()).collect::<Vec<_>>();
```
here qnrs are also qrs?

Pratyush: I see. I was planning originally on making it `generator * s.square()`, but the generator doesn't yet exist for all fields. So for the time being I will remove it, as I don't know of another way to easily generate a non-square.

weikengchen: Is it true that for a lot of the field we have access to a QNRS?

Pratyush: Yeah, for Fp the generator is always a QNR, but for extension fields we don't necessarily have QNRs

bench-templates/src/macros/field.rs
Pratyush: Fixed in [`b0e862b` (#472)](https://github.com/arkworks-rs/algebra/pull/472/commits/b0e862b405e3114ea353432a4c273defbb7e7c88)

bench-templates/src/macros/field.rs
Pratyush: Fixed in [`b0e862b` (#472)](https://github.com/arkworks-rs/algebra/pull/472/commits/b0e862b405e3114ea353432a4c273defbb7e7c88)

bench-templates/src/macros/field.rs
weikengchen: I feel maybe one day we will add a mandatory QNR field for extension fields. Thoughts?

Pratyush: This is related to #276

mmagician: Some of the extension fields used to have QNRs which were not used anywhere, see https://github.com/arkworks-rs/algebra/pull/421/.

Pratyush: Hm good catch, still not sure how we could use this generically.

PR #474, opened 2022-09-12
Tracking PR for moving generators from `FFTField` to `Field`

weikengchen:
## Description

We plan to move generators from `FFTField` to `Field`, which is for #276 and #472.

The current progress of this PR is that we have finished moving, but we need to supply the generators. I am currently factoring q^k-1 both by hand-waving and Sage, first for those curves in `test-curves`.

closes: #276

---

write a little note why.

- [ ] Wrote unit tests

### DIFF
--- a/ff/src/fields/mod.rs
+++ b/ff/src/fields/mod.rs
@@ -141,6 +141,9 @@ pub trait Field:
     /// The multiplicative identity of the field.
     const ONE: Self;
 
+    /// The generator of the multiplicative group of the field
+    const GENERATOR: Self;
+
     /// Returns the characteristic of the field,
     /// in little-endian representation.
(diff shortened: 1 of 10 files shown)


weikengchen:
After discussion, we decided to give up---it takes a long time to factor q^2 + q + 1 for BLS12-381.

Mikerah:
Since this is closed, are there any alternatives to exposing the generators so that even the `Field` trait has access to it? The only way I can think of is to combine both the `Field` and `FFTField` trait to create an overarching trait in order to get access to `multiplicative generator`

weikengchen:
I think a more comprehensive solution is to make it optional to specify a generator for a field.

Mikerah:
That could work as well.

PR #477, opened 2022-09-17
Fix benchmark names

andrewmilson:
## Description

Field benchmarks are incorrectly named. 
Criterion outputs benchmark names as "Conversions for $curve::...".
Names have been fixed to replace "$curve" with the actual intended name.

**Edit:** Additionally I went ahead and changed the metavariable name from "$curve_name" to "$field_name".

### DIFF
--- a/bench-templates/src/macros/field.rs
+++ b/bench-templates/src/macros/field.rs
@@ -63,7 +63,7 @@ macro_rules! f_bench {
 macro_rules! field_common {
     ($curve_name:expr, $F:ident) => {
         fn arithmetic(c: &mut $crate::criterion::Criterion) {
-            let name = format!("{}::{}", stringify!($curve), stringify!($F));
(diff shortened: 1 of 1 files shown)


weikengchen (approved):

Pratyush:
I think the commit [`eeef817` (#477)](https://github.com/arkworks-rs/algebra/pull/477/commits/eeef817052963b1e4309d4613925eeb19910c591) isn't quite what we want, since our benchmark framework associates each field with a particular curve. So calling the variable `field_name` isn't quite accurate. (See, e.g. the same file in this PR: https://github.com/arkworks-rs/algebra/blob/b6c2c04d64fa138f0756b33871e7feab2aed2fc0/bench-templates/src/macros/field.rs)

I think a good compromise would be to change the `field_name` there to `bench_group_name`, or something along those lines. I'll add that to my PR.

Issue #88, opened 2020-02-01
Change evaluation domain API for cosets


ValarDragon:
Could the struct for an evaluation domain contain a Field element for the affine shift, which then enables it for use as any coset. This not existing required me to re-implement a subset of this logic. (The current API only supports cosets with shift `g`, for a subset of methods)

My suggested API:
* Have two constructors, one for the subgroup case, one for the coset case.
* Remove cosetFFT, instead have the FFTs switch "if affine_shift == Field::one()"
* Add coset support in the lagrange interpolations. This is simple to do (see libiop: https://github.com/alexchmit/libiop-dev/blob/master/libiop/algebra/lagrange.tcc#L177. I also suggest switching to its method of doing arithmetic as it avoids doing one more pass over the array, with the same number of arithmetic ops)
* Update the vanishing polynomial, this just requires setting the constant term to `affine_offset^|H|`

ValarDragon:
Had a discussion with @Pratyush regarding an API for this.

1) Make Domain have two constructors, `new_subgroup(subgroup_size)` and `new_coset(coset_size, offset)`. 
2) Remove cosetFFT / cosetIFFT methods
3) Add a convenience method `domain.get_coset(offset)`, to make it easy to go from subgroup -> coset.

andrewmilson:
Are these changes still wanted? If so I'm happy to submit a pull request for this

Pratyush:
Yes!

PR #487, opened 2022-09-25
Change evaluation domain API for cosets

andrewmilson:
## Description

closes: #88

---

* Added two constructors `new_subgroup(subgroup_size)` and `new_coset(coset_size, offset)`. I kept the old constructor new(num_coeffs) since it contains logic to choose a subgroup size
* Removed cosetFFT / cosetIFFT methods
* Added get_coset(offset) convenience method 
* Removed `divide_by_vanishing_poly_on_coset_in_place()`. It wasn't clear if this method was appropriate anymore given domains can be coset domains.
* Removed generator_inv() method and added offset() and offset_inv() methods
* Moved `evaluate_all_lagrange_coefficients()` into a default implementation. **Is there any reason not to do the same for the vanishing_polynomial and element methods?**

---

write a little note why.

- [x] Wrote unit tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,14 +2,6 @@
 
 ## Pending
 
-- [\#487](https://github.com/arkworks-rs/algebra/pull/487) (`ark-poly`) Refactor `EvaluationDomain` trait for cosets:
-    - Add constructors `new_subgroup` and `new_coset`.
-    - Add convenience method `get_coset`.
-    - Add methods `offset` and `offset_inv`.
-    - Remove method `generator_inv`.
-    - Remove method `divide_by_vanishing_poly_on_coset_in_place`.
-    - Remove coset fft methods: `coset_fft`, `coset_fft_in_place`, `coset_ifft`, `coset_ifft_in_place`.
-
 ### Breaking changes
 
 - [\#300](https://github.com/arkworks-rs/algebra/pull/300) (`ark-ec`) Change the implementation of `Hash` trait of `GroupProjective` to use the affine coordinates.
@@ -134,6 +126,10 @@
     - New serialization format for Twisted Edwards curves:
         - Points with a "positive" x-coordinate are serialized with the sign bit set to zero.
         - Points with a "negative" x-coordinate are serialized with the sign bit set to one.
+- [\#487](https://github.com/arkworks-rs/algebra/pull/487) (`ark-poly`) Refactor `EvaluationDomain` trait for cosets:
+    - Remove method `generator_inv`.
+    - Remove method `divide_by_vanishing_poly_on_coset_in_place`.
+    - Remove coset fft methods: `coset_fft`, `coset_fft_in_place`, `coset_ifft`, `coset_ifft_in_place`.
 
 ### Features
 
@@ -154,6 +150,10 @@
 - [\#446](https://github.com/arkworks-rs/algebra/pull/446) (`ark-ff`) Add `CyclotomicMultSubgroup` trait and impl for extension fields
 - [\#467](https://github.com/arkworks-rs/algebra/pull/467) (`ark-ec`)
     - Move implementation of `serialize_with_mode()`, `deserialize_with_mode()`, and `serialized_size()` into `{SW,TE}CurveConfig` to allow customization.
+- [\#487](https://github.com/arkworks-rs/algebra/pull/487) (`ark-poly`) Refactor `EvaluationDomain` trait for cosets:
+    - Add constructors `new_subgroup` and `new_coset`.
+    - Add convenience method `get_coset`.
+    - Add methods `coset_offset` and `coset_offset_inv`.
 
 ### Improvements
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -151,9 +151,9 @@
 - [\#467](https://github.com/arkworks-rs/algebra/pull/467) (`ark-ec`)
     - Move implementation of `serialize_with_mode()`, `deserialize_with_mode()`, and `serialized_size()` into `{SW,TE}CurveConfig` to allow customization.
 - [\#487](https://github.com/arkworks-rs/algebra/pull/487) (`ark-poly`) Refactor `EvaluationDomain` trait for cosets:
-    - Add constructors `new_subgroup` and `new_coset`.
+    - Add constructor `new_coset`.
     - Add convenience method `get_coset`.
-    - Add methods `coset_offset` and `coset_offset_inv`.
+    - Add methods `coset_offset`, `coset_offset_inv` and `coset_offset_pow_size`.
 
 ### Improvements
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,14 @@
 
 ## Pending
 
+- [\#487](https://github.com/arkworks-rs/algebra/pull/487) (`ark-poly`) Refactor `EvaluationDomain` trait for cosets:
+    - Add constructors `new_subgroup` and `new_coset`.
+    - Add convenience method `get_coset`.
+    - Add methods `offset` and `offset_inv`.
+    - Remove method `generator_inv`.
+    - Remove method `divide_by_vanishing_poly_on_coset_in_place`.
+    - Remove coset fft methods: `coset_fft`, `coset_fft_in_place`, `coset_ifft`, `coset_ifft_in_place`.
+
 ### Breaking changes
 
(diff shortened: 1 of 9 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,14 @@
 
 ## Pending
 
+- [\#487](https://github.com/arkworks-rs/algebra/pull/487) (`ark-poly`) Refactor `EvaluationDomain` trait for cosets:
```
Pratyush, line 5: Let's move parts of these to "Breaking changes" and "Features" (there's usually nothing directly under `Pending`, only under the other subheadings.

poly/src/domain/mod.rs
```diff
@@ -46,6 +46,17 @@ pub trait EvaluationDomain<F: FftField>:
     /// having `num_coeffs` coefficients.
     fn new(num_coeffs: usize) -> Option<Self>;
 
+    /// Construct a domain
+    fn new_subgroup(subgroup_size: usize) -> Option<Self>;
```
Pratyush, line 50: Is this method different from `new`? That is, is the output checked to be a multiplicative subgroup of `self`? If so, we should make this consistent with `get_coset`, and name the method `get_subgroup`

poly/src/domain/mod.rs
```diff
@@ -70,8 +81,11 @@ pub trait EvaluationDomain<F: FftField>:
     /// Return the group inverse of `self.group_gen()`.
     fn group_gen_inv(&self) -> F;
 
-    /// Return the inverse of the multiplicative generator of `F`.
-    fn generator_inv(&self) -> F;
+    /// Return the offset that defines this domain.
+    fn offset(&self) -> F;
```
Pratyush, line 85: Would it be accurate to call this `coset_offset`, or `offset_from_group`? Otherwise the name is kind of mysterious

poly/src/domain/radix2/mod.rs
```diff
@@ -127,87 +149,16 @@ impl<F: FftField> EvaluationDomain<F> for Radix2EvaluationDomain<F> {
...
-            // Invert the lagrange coefficients inverse, to get the actual coefficients,
-            // and return these
-            batch_inversion(lagrange_coefficients_inverse.as_mut_slice());
-            lagrange_coefficients_inverse
-        }
-    }
-
     fn vanishing_polynomial(&self) -> crate::univariate::SparsePolynomial<F> {
-        let coeffs = vec![(0, -F::one()), (self.size(), F::one())];
+        let coeffs = vec![(0, -self.offset.pow([self.size])), (self.size(), F::one())];
```
Pratyush, line 153: We should specialize this in the case that `self.offset.is_one()`, to avoid unnecessary work.
```suggestion
        let constant_coeff = if self.offset.is_one() { self.offset } else { self.offset.pow([self.size]) };
        let coeffs = vec![(0, -constant_coeff), (self.size(), F::one())];
```

poly/src/domain/radix2/mod.rs
```diff
@@ -127,87 +149,16 @@ impl<F: FftField> EvaluationDomain<F> for Radix2EvaluationDomain<F> {
...
         crate::univariate::SparsePolynomial::from_coefficients_vec(coeffs)
     }
 
     /// This evaluates the vanishing polynomial for this domain at tau.
     /// For multiplicative subgroups, this polynomial is `z(X) = X^self.size -
     /// 1`.
     fn evaluate_vanishing_polynomial(&self, tau: F) -> F {
-        tau.pow(&[self.size]) - F::one()
+        tau.pow([self.size]) - self.offset.pow([self.size])
```
Pratyush, line 161:
```suggestion
        let constant_coeff = if self.offset.is_one() { self.offset } else { self.offset.pow([self.size]) };
        tau.pow([self.size]) - constant_coeff
```

poly/src/domain/utils.rs
```diff
@@ -202,7 +203,7 @@ impl<F: FftField> Iterator for Elements<F> {
             let cur_elem = self.cur_elem;
             self.cur_elem *= &self.group_gen;
             self.cur_pow += 1;
-            Some(cur_elem)
+            Some(self.offset * cur_elem)
```
Pratyush, line 206: We should just set `cur_elem` to be multiplied by `self.offset` at construction time, to avoid the extra multiplication on every call to `next()`

Pratyush:
Thanks for the PR! These changes all look great. Left some comments!

Pratyush:
Regarding moving `elements` and `vanishing_polynomial` and `evaluate_vanishing_polynomial` also into the default impl, I think it makes sense!

poly/src/domain/radix2/mod.rs
Pratyush: Another way around this would be to just store `self.offset.pow([self.size])` in the domain when constructing it (though this could hinder a default impl in the trait)

andrewmilson: Yeah, I prefer storing it in the domain

poly/src/domain/mod.rs
andrewmilson: Not sure I fully understand. 

My understanding is that new(num_coefficients) actually determines a subgroup size and new_subgroup(subgroup_size) is provided with a subgroup size.

Pratyush: It's effectively the same; the `new_subgroup` seems to assume that its input is a power of 2, whereas `new` does the rounding. Since the two are basically the same, I'd prefer to avoid duplicating the API, and just keep `new` which rounds up.

poly/src/domain/radix2/mod.rs
```diff
@@ -127,102 +153,22 @@ impl<F: FftField> EvaluationDomain<F> for Radix2EvaluationDomain<F> {
...
-    /// their power of the generator which they correspond to.
-    /// e.g. the `i`-th element is g^i
-    fn element(&self, i: usize) -> F {
-        // TODO: Consider precomputed exponentiation tables if we need this to be
-        // faster.
-        self.group_gen.pow(&[i as u64])
+        tau.pow([self.size]) - self.offset_pow_size
     }
 
     /// Return an iterator over the elements of the domain.
```
Pratyush, line 168: I guess this can be made generic too, if you're so inclined.

poly/src/domain/mixed_radix.rs
```diff
@@ -37,8 +37,12 @@ pub struct MixedRadixEvaluationDomain<F: FftField> {
     pub group_gen: F,
     /// Inverse of the generator of the subgroup.
     pub group_gen_inv: F,
-    /// Multiplicative generator of the finite field.
-    pub generator_inv: F,
+    /// Offset that specifies the coset.
+    pub offset: F,
+    /// Inverse of the offset that specifies the coset.
+    pub offset_inv: F,
+    /// Constant coefficient for the vanishing polynomial.
```
Pratyush, line 44:
```suggestion
    /// Constant coefficient for the vanishing polynomial. 
    /// Equals `self.offset^self.size`.
```

poly/src/domain/mixed_radix.rs
```diff
@@ -365,7 +325,7 @@ pub(crate) fn serial_mixed_radix_fft<T: DomainCoeff<F>, F: FftField>(
 
         // Doing the q_adicity passes.
         for _ in 0..q_adicity {
-            let w_m = omega.pow(&[(n / (q * m)) as u64]);
+            let w_m = omega.pow([(n / (q * m)) as u64]);
```
Pratyush, line 328: As a future microoptimization, we can precompute `w_m` as a table outside the `for` loop, and save some multiplications. (Not clear if this would be worth it at all)

poly/src/domain/mod.rs
```diff
@@ -131,71 +146,102 @@ pub trait EvaluationDomain<F: FftField>:
...
+    fn evaluate_vanishing_polynomial(&self, tau: F) -> F {
+        // TODO: Consider precomputed exponentiation tables if we need this to be
+        // faster.
+        tau.pow([self.size() as u64]) - self.coset_offset_pow_size()
+    }
 
     /// Returns the `i`-th element of the domain.
-    fn element(&self, i: usize) -> F;
+    fn element(&self, i: usize) -> F {
+        self.coset_offset() * self.group_gen().pow([i as u64])
```
Pratyush, line 239:
```suggestion
        let mut result = self.group_gen().pow([i as u64]);
        if !self.coset_offset().is_one() {
            result *= self.coset_offset()
        }
        result
```

PR #493, opened 2022-09-30
Add degree aware FFT

andrewmilson:
## Description

closes: #83

---

write a little note why.

- [ ] Wrote unit tests

### DIFF
--- a/poly/src/domain/radix2/fft.rs
+++ b/poly/src/domain/radix2/fft.rs
@@ -37,7 +37,7 @@ impl<F: FftField> Radix2EvaluationDomain<F> {
         // Therefore those first i rounds have the effect of copying the evaluations into more locations,
         // so we handle this in initialization, and reduce the number of loops that are performing arithmetic.
         // The number of times we copy each initial non-zero element is as below:
-        let duplicity_of_initials = 1 << (log_n - log_d);
+        let duplicity_of_initials = 1 << log_n.checked_sub(log_d).expect("domain is too small");
         coeffs.resize(n, T::zero());
 
         // swap coefficients in place
@@ -50,16 +50,17 @@ impl<F: FftField> Radix2EvaluationDomain<F> {
 
         // duplicate initial values
         if duplicity_of_initials > 1 {
-            for i in (0..n).step_by(duplicity_of_initials) {
-                let v = coeffs[i];
-                coeffs[i + 1..i + duplicity_of_initials].fill(v);
-            }
+            ark_std::cfg_chunks_mut!(coeffs, duplicity_of_initials).for_each(|chunk| {
+                let v = chunk[0];
+                chunk[1..].fill(v);
+            });
         }
 
         let start_gap = duplicity_of_initials;
         self.oi_helper(&mut coeffs[..], self.group_gen, start_gap);
     }
 
+    #[allow(unused)]
     pub(crate) fn in_order_fft_in_place<T: DomainCoeff<F>>(&self, x_s: &mut [T]) {
         if !self.offset.is_one() {
             Self::distribute_powers(x_s, self.offset);
--- a/poly/src/domain/radix2/fft.rs
+++ b/poly/src/domain/radix2/fft.rs
@@ -20,6 +20,46 @@ enum FFTOrder {
 }
 
 impl<F: FftField> Radix2EvaluationDomain<F> {
+    /// Degree aware FFT that runs in O(n log d) instead of O(n log n)
+    /// Implementation copied from libiop.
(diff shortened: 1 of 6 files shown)


poly/src/domain/radix2/fft.rs
```diff
@@ -20,6 +20,46 @@ enum FFTOrder {
...
+        let num_coeffs = coeffs.len();
+        let log_d = ark_std::log2(num_coeffs);
+        // When the polynomial is of size k*|coset|, for k < 2^i,
+        // the first i iterations of Cooley Tukey are easily predictable.
+        // This is because they will be combining g(w^2) + wh(w^2), but g or h will always refer
+        // to a coefficient that is 0.
+        // Therefore those first i rounds have the effect of copying the evaluations into more locations,
+        // so we handle this in initialization, and reduce the number of loops that are performing arithmetic.
+        // The number of times we copy each initial non-zero element is as below:
+        let duplicity_of_initials = 1 << (log_n - log_d);
```
Pratyush, line 40: Let's make this is a `checked_sub`, or maybe handle the case where `d >= n` separately.

poly/src/domain/radix2/fft.rs
```diff
@@ -20,6 +20,46 @@ enum FFTOrder {
...
+            if i < ri {
+                coeffs.swap(i as usize, ri as usize);
+            }
+        }
+
+        // duplicate initial values
+        if duplicity_of_initials > 1 {
+            for i in (0..n).step_by(duplicity_of_initials) {
+                let v = coeffs[i];
+                coeffs[i + 1..i + duplicity_of_initials].fill(v);
```
Pratyush, line 55: I think this is equivalent? 

It's also parallelizable, so might be a tad bit faster. There are also potentially other parallelizations that could be better (e.g. just chunking `coeffs` up into `num_cpu` chunks, and allocating each chunk to a thread to process serially.)

```suggestion
            ark_std::cfg_chunks_mut!(coeffs, duplicity_of_initials).for_each(|c| {
            	let v = c[i];
                c[i + 1][.. duplicity_of_initials].fill(v);
            });
```

Pratyush:
Thanks, this PR is awesome! If you'd like to add benchmark numbers that would be great. 

For the `unused` errors, you can add `#[allow(unused)]` on the relevant items.

andrewmilson:
@Pratyush  sad to say the benchmarks are rather depressing. ~7% performance regression on the FFT tests. I added a couple of benchmarks for polynomial multiplication. Multiplication of two polynomials with equal degree also has ~7% regression in performance. The only situation this implementation performs better is when multiplying polynomials of substantially different degree i.e. `|poly_a| = |poly_b| * 32` - I actually tested this specific case and it only had a 10% gain over the O(n log n) implementation. 

I think the performance degradation is due to `fft_in_place` using oi_helper (now) instead of io_helper (before). The master branch IFFT benchmarks are 10% slower than the master branch FFT benchmarks and I'm pretty sure it's attributed to IFFTs using `oi_helper` and FFTs using `io_helper`

Pratyush:
Wow that's quite surprising. Let me take a look and see if we can optimize the out_of_order code.

Pratyush:
@ValarDragon any ideas? I've been looking, but can't find a fix.

Pratyush:
@andrewmilson I've changed it to only invoke the degree_aware_fft when `domain.size() >= p.degree() * 4`. I'm seeing perf improvements from doing that. Let me know if this looks good to you!

Pratyush (approved):

poly/src/domain/radix2/mod.rs
```diff
@@ -142,8 +142,17 @@ impl<F: FftField> EvaluationDomain<F> for Radix2EvaluationDomain<F> {
...
+            coeffs.len()
+        } else {
+            coeffs.len().next_power_of_two()
+        };
+        if self.size() >= (num_coeffs << 2) {
+            self.degree_aware_fft_in_place(coeffs);
+        } else {
+            coeffs.resize(self.size(), T::zero());
+            self.in_order_fft_in_place(coeffs);
+        }
```
andrewmilson, line 155: Could this just be

```rust
if coeffs.len() * 4 <= self.size() {
    self.degree_aware_fft_in_place(coeffs);
} else {
    coeffs.resize(self.size(), T::zero());
    self.in_order_fft_in_place(coeffs);
}
```

?

andrewmilson: Pushed these changes. Marking resolved

andrewmilson:
Awesome thanks. LGTM

PR #495, opened 2022-10-11
if "greatest" is selected, second coordinate should be picked

mmagician:
## Description

If I'm not mistaken, then according to the docs and code of `get_ys_from_x_unchecked`, the second value should be selected if `greatest` flag is true. 

---

write a little note why.

- [ ] Wrote unit tests


Pratyush:
Thanks for the fix, it is indeed inconsistent with the documentation. Could you check against the call sites to see if this fix works? I'm asking because the code recently changed.

mmagician:
It's only used in `sample` and `get_random_bytes`. These methods didn't care for anything specific regarding the y-coordinate, as long as it was a valid point - which it is no matter which coordinate is picked.

Pratyush (approved):

PR #503, opened 2022-10-25
Fix line function evaluation when adding for BLS12 G2

Pratyush:
## Description

Reported by user codeblooded in discord. Crosschecked against Eqn(13) in https://eprint.iacr.org/2013/722

---

write a little note why.

- [ ] Wrote unit tests


weikengchen (approved):

weikengchen:
How breaking would this change be?

simonmasson:
I think the code in `main` is actually right, and it is probably a typo on the paper you mentioned. See [this paper](https://eprint.iacr.org/2009/243.pdf) where it is indeed λ and not -λ. This explains why the tests do not pass with this modification. I am curious if this code was written following [this paper](https://eprint.iacr.org/2013/722) (with the typo) because it seems to be very similar (same names of variables, etc.), but without the typo :-) .

weikengchen:
I think the best solution at this moment is to drop an email to the four authors.

confuseSUN:
I agree that equation (13) is probably a typo, we can verify it with the tangent equation, in fact, for M types (j, -theta, lambda) and (-j, theta, -lambda) are equivalent.

Pratyush:
Let's close this for the time being, seems like the paper was incorrect.

PR #506, opened 2022-10-29
add serde to Field

mimoo:
I presume this is no good, but let's start a discussion o.o 

It is currently really painful to use arkworks in real-world projects because as soon as you want to serialize to disk or to send something over the wire, you can't use serde. Or at least, you have to resort to very verbose tricks (using serde_with, for example) to be able to serialize/deserialize any struct you have that contains `Field`.

### DIFF
--- a/ff/Cargo.toml
+++ b/ff/Cargo.toml
@@ -22,6 +22,7 @@ derivative = { version = "2", features = ["use_core"] }
 num-traits = { version = "0.2", default-features = false }
 paste = "1.0"
 rayon = { version = "1", optional = true }
+serde = { version = "1.0", features = ["derive"] }
 zeroize = { version = "1", default-features = false, features = ["zeroize_derive"] }
 num-bigint = { version = "0.4", default-features = false }
 digest = { version = "0.10", default-features = false, features = ["alloc"] }
--- a/ff/src/biginteger/mod.rs
+++ b/ff/src/biginteger/mod.rs
@@ -18,13 +18,74 @@ use ark_std::{
     vec::Vec,
 };
 use num_bigint::BigUint;
+use serde::{Deserialize, Serialize};
 use zeroize::Zeroize;
 
 #[macro_use]
 pub mod arithmetic;
 
-#[derive(Copy, Clone, PartialEq, Eq, Debug, Hash, Zeroize)]
-pub struct BigInt<const N: usize>(pub [u64; N]);
+mod arrays {
+    use ark_std::vec::Vec;
+    use ark_std::{convert::TryInto, marker::PhantomData};
+
+    use serde::{
+        de::{SeqAccess, Visitor},
+        ser::SerializeTuple,
+        Deserialize, Deserializer, Serialize, Serializer,
+    };
+    pub fn serialize<S: Serializer, T: Serialize, const N: usize>(
+        data: &[T; N],
+        ser: S,
+    ) -> Result<S::Ok, S::Error> {
+        let mut s = ser.serialize_tuple(N)?;
+        for item in data {
+            s.serialize_element(item)?;
+        }
+        s.end()
+    }
+
+    struct ArrayVisitor<T, const N: usize>(PhantomData<T>);
+
+    impl<'de, T, const N: usize> Visitor<'de> for ArrayVisitor<T, N>
+    where
+        T: Deserialize<'de>,
+    {
+        type Value = [T; N];
+
+        fn expecting(&self, formatter: &mut ark_std::fmt::Formatter<'_>) -> ark_std::fmt::Result {
+            formatter.write_str(&format!("an array of length {}", N))
+        }
+
+        #[inline]
+        fn visit_seq<A>(self, mut seq: A) -> Result<Self::Value, A::Error>
+        where
+            A: SeqAccess<'de>,
+        {
+            // can be optimized using MaybeUninit
+            let mut data = Vec::with_capacity(N);
+            for _ in 0..N {
+                match (seq.next_element())? {
+                    Some(val) => data.push(val),
+                    None => return Err(serde::de::Error::invalid_length(N, &self)),
+                }
+            }
+            match data.try_into() {
+                Ok(arr) => Ok(arr),
+                Err(_) => unreachable!(),
+            }
+        }
+    }
+    pub fn deserialize<'de, D, T, const N: usize>(deserializer: D) -> Result<[T; N], D::Error>
+    where
+        D: Deserializer<'de>,
+        T: Deserialize<'de>,
+    {
+        deserializer.deserialize_tuple(N, ArrayVisitor::<T, N>(PhantomData))
+    }
+}
+
+#[derive(Copy, Clone, PartialEq, Eq, Debug, Hash, Zeroize, Serialize, Deserialize)]
+pub struct BigInt<const N: usize>(#[serde(with = "arrays")] pub [u64; N]);
 
 impl<const N: usize> Default for BigInt<N> {
     fn default() -> Self {
--- a/ff/src/fields/mod.rs
+++ b/ff/src/fields/mod.rs
@@ -12,6 +12,7 @@ use ark_std::{
 
 pub use ark_ff_macros;
 use num_traits::{One, Zero};
+use serde::{de::DeserializeOwned, Serialize};
 use zeroize::Zeroize;
 
 pub mod utils;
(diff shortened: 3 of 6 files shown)


weikengchen:
This is highly subject to debate.

mimoo:
BTW how do people deal with this in projects that run in prod? We chose to use serde_with everywhere but I'm thinking of just forking arkworks with this branch and using this atm.

Pratyush:
Thanks for the PR! Couple of points
* Let's make it optional, so that it's not on always
* Should we directly implement it for all types that impl `CanonicalSerialize/Deserialize`, instead of one-off impls?

Pratyush:
Hm I had something more like this in mind:
```rust
// in ark-serialize
pub type Compressed<T: CanonicalSerialize + CanonicalDeserialize>(T);
pub type Uncompressed<T: CanonicalSerialize + CanonicalDeserialize>(T);

#[cfg(feature = "serde")]
impl<T: CanonicalSerialize> Serialize for Compressed<T> {
}

#[cfg(feature = "serde")]
impl<T: CanonicalDeserialize> Deserialize for Compressed<T> {
}

#[cfg(feature = "serde")]
impl<T: CanonicalSerialize> Serialize for Uncompressed<T> {
}

#[cfg(feature = "serde")]
impl<T: CanonicalDeserialize> Deserialize for Uncompressed<T> {
}
```

You can use this in your code as follows:

```rust
#[derive(Serialize, Deserialize)]
pub struct Proof {
    pub type: Compressed<Whatever> // or Uncompressed<Whatever>
}
```
Or like this, at serialization/deserialization boundaries:
```rust
#[derive(CanonicalSerialize, CanonicalDeserialize)]
pub struct Proof { ... }

fn broadcast_to_network(p: Proof, r: &mut Read) {
    Compressed(p).serialize_to_json(r); // or whatever is your serde-interfacing code
}
```
If the latter works for your code, I think it's better because it requires minimal code in arkworks libraries.

mimoo:
Humm, I'm not sure I understand your example. What is `Uncompressed`? 

Also, if I understand correctly:

* either I have to implement `CanonicalSerialize`, `CanonicalDeserialize` on my types, that's probably annoying
* or I have to always handle this wrapper type around my `Field` elements, which could be in arrays, tuples, and other types. That sounds a bit annoying as well : o

As in, if I have the following:

```rust
pub struct Proof {
    pub single: Compressed<Field>,
    pub vec: Vec<Compressed<Field>>,
}
```

then I'd have to do things like `proof.vec.into_iter().map(|x| x.0)...` right? Although if I implement `Deref` it might not be that bad...

Pratyush:
The idea is to define `Uncompressed` and `Compressed` as newtype wrappers around `T` in `ark-serialize`.

Regarding the two options, you can derive `CanonicalSerialize` and `CanonicalDeserialize` easily for your structs; we have `derive` macros for these traits. Probably that would be the least effort approach

mimoo:
hummm I see, but what is the point of Uncompressed?

Pratyush:
Ah `Uncompressed` and `Compressed` together implement the different behaviours of compression. The idea is that the `Serialize` impl of `Compressed<T>` will call `T::serialize_compressed`, while the `Serialize` impl for `Uncompressed<T>` will call `T::serialize_uncompressed`. (ditto for the `Deserialize` impls).

We need these additional wrapper types because a blanket `impl<T: CanonicalSerialize> serde::Serialize for T` is not possible due to Rust's orphan rules

mmaker:
Do we ever need to send an uncompressed point through the wire?

Pratyush:
People definitely use it when serializing to disk.

mimoo:
Sorry for the delayed replayed. 

The problem I see with this approach is that you're creating a wrapper type (I presume you meant `struct Compressed` and not `type Compressed` in your example). And so it will be missing all the implementations that might make it useful (like `Field`)

Pratyush:
@mimoo The usage I had in mind was more along the following lines:

```rust
#[derive(CanonicalSerialize, CanonicalDeserialize)]
pub struct Proof {
    // Your actual proof contents
}

// elsewhere, in your serialization code
let proof = Pickles::prove(stuff);
let compressed = Compressed(proof);
compressed.serialize_json(); // or whatever else serde method you want to use.

// similarly, while deserializing:
let compressed: Compressed<Proof> = deserialize_json();
let proof = compressed_proof.0;
// rest of your code.
```

I *think* that should be easier to handle, no?

mimoo:
oh, but that means we have to implement `Canonical(De)Serialize` on all of our types instead of serde + arkworks has to implement these traits on Rust primitive types (not sure if this is already done). Let me think about this : o

Pratyush:
You wouldn't need to impl `CanonicalSerialize/CanonicalDeserialize` for all your types. For example, a hypothetical `Transaction` type would look like:

```rust
#[derive(Serialize, Deserialize)]
pub struct Transaction {
    state_root: [u8; 32],
    signature: [u8; 64],
    proof: Compressed<Proof>,
}

#[derive(CanonicalSerialize, CanonicalDeserialize)]
pub struct Proof {
    // proof contents
}
```

That is, you can be as granular as you want, and insert the CanonicalSerialize → Serialize transition wherever you want in the stack.

mimoo:
I see, another problem I can see is that I'm assuming the derivers don't come with the helpers the serde ones come with. For example `#[serde(bound)]` or `#[serde(skip)]`

burdges:
I'd suggest this become a separate `ark-serde` crate since you need wrappers anyways.  It'll look exactly like pratyush's code above, but avoid feature flag confusion and provide more visible documentation.

mimoo:
I'm running into another issue now: I can't nicely JSON serialize some of the types in arkworks. For example, a Radix2EvaluationDomain is just going to look like a bytearray instead of a nice JSON struct. I talk more about this here: https://github.com/o1-labs/proof-systems/pull/1044

I think there's no good solution here besides implementing serde for these types of structs. I still don't understand why we don't do this, and just choose compressed format as the default for elliptic curve points.

burdges:
> a Radix2EvaluationDomain is just going to look like a bytearray instead of a nice JSON struct. 

Why is this a problem?  Users cannot read finite field elements anyways.

> I still don't understand why we don't do this, and just choose compressed format as the default for elliptic curve points.

Why do so many serialization formats exist?  Why is serde never going to be part of std?  It's total chaos.. [Inconsistent, incorrect, or inadequately defined behavior](https://github.com/serde-rs/json/labels/bug), incorrect canonicalization, format changes, semver fights, poor memory usage on HSMs, etc.  Are capnproto arenas using unsafe or C code?  Voila, secret leakage risk in serde!

It's basically universal that cryptography libraries only ever serialize as bytes, even [when they support serde](https://github.com/dalek-cryptography/ed25519-dalek/blob/main/src/verifying.rs#L539-L558).  Afaik even JS crypto libraries and the Web Crypto API obey this convention. 

Also, `ark-serialize` has options for both compression and validation, but you might change validation criteria, like by doing a curve check but not a subgroup check, or something even more subtle.  We'd never propagate this through serde correctly.

burdges:
Just fyi, I wrote wrapper types like this for arkworks-scale which currently live at https://github.com/w3f/ring-vrf/tree/master/ark-scale, although likely they'll move so a snapshot is https://github.com/w3f/ring-vrf/tree/eb6c4b24ef0c35a821eb521236277174b054af0e/ark-scale  It's simpler than serde because parity-scale-codec works almost exactly like arkworks' serialization conceptually.  It's straightforward to serialize directly from an iterator into the bytes that'll decode as a `Vec<T>`, which maybe doable with serde since the visitor goes in sequence: https://github.com/w3f/ring-vrf/blob/eb6c4b24ef0c35a821eb521236277174b054af0e/ark-scale/src/lib.rs#L247 and not sure if this is useful for most people anyways.

burdges:
Appears serde_derive is a precompiled binary from someone who works for Palantir.

https://www.bleepingcomputer.com/news/security/rust-devs-push-back-as-serde-project-ships-precompiled-binaries/

https://twitter.com/eddyb_r/status/1693198713286541429
https://twitter.com/eddyb_r/status/1693215683449069981

mmaker:
Hello there just to say we are at zksummit and still thinking about this issue. Hope you guys are all well

mmaker:
Trying to be constructive, here's what it seems to be needed to move forward that issue: 

Add an optional `serde` feature ([ref.](https://github.com/arkworks-rs/algebra/pull/506#issuecomment-1296020231)) that and instruct serde in such a way that it will be able to serialize/deserialized compressed/uncompressed with/without validation ([ref.](https://github.com/arkworks-rs/algebra/pull/506#issuecomment-1298867831)).  **Make ser/deser compressed with validation the default**, but allow either with serde compile flags (or with wrapper structs, [ref.](https://github.com/arkworks-rs/algebra/pull/506#issuecomment-1297847529)) all other options.

mimoo:
sorry I have no time on my hands atm :o

Pratyush:
I think this is closed by #967, so closing now.

Issue #319, opened 2021-09-29
Multiscalar multiplication doesn't enforce lengths of input vectors


gtank:
The variable base `multi_scalar_mul` will silently truncate its inputs if the lengths don't match. This is not usually a desired behavior and would produce incorrect (or at least unexpected) results.

https://github.com/arkworks-rs/algebra/blob/d7fa00e708d76465cc49ed5e33ea42ff17f599b1/ec/src/msm/variable_base.rs#L16-L19

We've run into this a couple of times (https://github.com/celo-org/celo-bls-snark-rs/pull/181, https://github.com/celo-org/celo-bls-snark-rs/pull/222/commits/681cf111cf73459847bacf4378a26ba25a865fcc) and always have to include length checks in the caller code. Instead, the function could return a Result or Option that can express the "unequal inputs" error condition. 

I'm not sure what the preferred idiom for this library would be but I'd be happy to try implementing it if there is one.

weikengchen:
I see. Yes, the code intentionally does the truncation.

Let me ask Pratyush if this implementation is intentional. But I agree that for safety, it is better to fail when it does not match.

@Pratyush

Pratyush:
In downstream usage (eg: the KZG impl in poly-commit), doing that gets cumbersome, because you end up with some variation of `msm(scalars, srs[..scalars.len()])`.

Also, IMO there's a difference between the two cases, stemming from the primary usage of MSMs in commitments:
* when `scalars.len() > group_elements.len()`, the additional scalars get ignored, which means they aren't committed to.
* when `scalars.len() < group_elements.len()`, the additional group elements just get paired with "zero" scalars, and this is still a valid commitment to the original scalars.

Happy to revise this opinion if y'all think that the latter case also has issues.

PR #519, opened 2022-11-20
Switch comments for unchecked and checked msm versions

kevaundray:
## Description

closes: #319

---

write a little note why.

- [ ] Wrote unit tests


weikengchen:
Is it true that `msm_unchecked` should actually be `msm_checked`?

kevaundray:
I am not sure what the default behaviour of `msm` is intended to be here, I only changed the comments to match what the implementation is doing.

It looks like the intended default behaviour is: 

- `msm` should truncate, ie it's basically `msm_unchecked`
- `msm_unchecked` is actually checking the arguments and should be `msm_checked`

I'm having problems with my laptop at the moment, so feel free to make any changes you see fit

mmaker:
On Sun 20 Nov 2022 at 17:02, kevaundray ***@***.***> wrote:

> I am not sure what the default behaviour of msm is intended to be here, I
> only changed the comments to match what the implementation is doing.
>
> It looks like the intended default behaviour is:
>
>    - msm should truncate, ie it's basically msm_unchecked
>    - msm_unchecked is actually checking the arguments and should be
>    msm_checked
>
> I'm having problems with my laptop at the moment, so feel free to make any
> changes you see fit
>

That’s correct, the reason being that in Rust unsafe is always more verbose

Pratyush:
Yes, it should be the case that `msm` checks lengths, and `msm_unchecked` does not.

Pratyush (commented):
This looks good to me. I believe you only need to fix some tests now that the type signatures are switched.

Issue #520, opened 2022-11-23
DensePolynomial::evaluate_over_domain truncates polynomial coefficients


tjade273:
Dense polynomials evaluated over a domain of size `k` are evaluated as if all terms of degree greater than `k` are zero, due to `domain.fft` implicitly truncating the coefficients. 

Sparse polynomials evaluate the full polynomial.

```
    #[test]
    fn evaluate_high_degree_over_domain() {
        let domain = GeneralEvaluationDomain::new(2).unwrap();
        let f = DensePolynomial::from_coefficients_vec(vec![Fr::zero(), Fr::zero(), Fr::one()]);
        let evals = f.evaluate_over_domain(domain);
        let x = evals[0];
        assert_eq!(x, Fr::zero());
    }
```

```
    #[test]
    fn evaluate_high_degree_over_domain() {
        let domain = GeneralEvaluationDomain::new(2).unwrap();
        let f = SparsePolynomial::from_coefficients_vec(vec![(2, Fr::from(1))]);
        let evals = f.evaluate_over_domain(domain);
        let x = evals[0];
        assert_eq!(x, Fr::from(1));
    }
```

https://github.com/arkworks-rs/algebra/blob/e062140351e7775d30c8b0167fbef048fcd4535e/poly/src/polynomial/univariate/mod.rs#L145-L163

PR #521, opened 2022-11-23
Fix evaluation of dense polynomials over domains smaller than the degree

tjade273:
## Fixes the evaluation of dense polynomials over domains smaller than the polynomial degree.

Currently, the coefficients of dense polynomials are truncated to match the size of the evaluation domain. This causes unexpected / incorrect behavior when larger polynomials are used. This PR resolves the issue by reducing the polynomial mod `X^d` where `d` is the size of the domain, before the FFT is performed.

A new unit test is located in `src/polynomial/univariate/sparse.rs`

It may also be a good idea to throw an error in the FFT routine when the coefficient vector is bigger than the FFT size, so that information is not accidentally discarded.

closes: #520

---

write a little note why.

- [ x ] Targeted PR against correct branch (master)
- [ x ] Linked to GitHub issue with discussion and accepted design OR have an explanation in the PR that describes this work.
- [ x ] Wrote unit tests

### DIFF
--- a/poly/src/polynomial/univariate/mod.rs
+++ b/poly/src/polynomial/univariate/mod.rs
@@ -153,9 +153,20 @@ impl<'a, F: 'a + FftField> DenseOrSparsePolynomial<'a, F> {
                 Evaluations::from_vec_and_domain(evals, domain)
             },
             DPolynomial(Cow::Borrowed(d)) => {
-                Evaluations::from_vec_and_domain(domain.fft(&d.coeffs), domain)
+                let chunks = d.coeffs.chunks(domain.size());
+                let reduced = chunks.fold(vec![F::zero(); domain.size()], |x, y| {
+                    x.iter().zip(y).map(|(&x, y)| x + y).collect()
+                });
+                Evaluations::from_vec_and_domain(domain.fft(&reduced), domain)
             },
             DPolynomial(Cow::Owned(mut d)) => {
+                let mut chunks = d.coeffs.chunks_mut(domain.size());
+                let coeffs = chunks.next().unwrap();
+                for chunk in chunks {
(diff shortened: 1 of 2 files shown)


poly/src/polynomial/univariate/mod.rs
```diff
@@ -153,9 +153,20 @@ impl<'a, F: 'a + FftField> DenseOrSparsePolynomial<'a, F> {
                 Evaluations::from_vec_and_domain(evals, domain)
             },
             DPolynomial(Cow::Borrowed(d)) => {
-                Evaluations::from_vec_and_domain(domain.fft(&d.coeffs), domain)
+                let chunks = d.coeffs.chunks(domain.size());
+                let reduced = chunks.fold(vec![F::zero(); domain.size()], |x, y| {
+                    x.iter().zip(y).map(|(&x, y)| x + y).collect()
```
Pratyush, line 158:
```suggestion
                    cfg_iter!(x).zip(y).map(|(&x, y)| x + y).collect()
```

Pratyush: Actually, can avoid extra allocations here by making this a for loop, like so:

```rust
let mut reduced = chunks[0].to_vec();
for chunk in chunks {
    cfg_iter_mut!(reduced).zip(chunk).for_each(|(x, y)| {
    *x += y;
});
}
```

Pratyush (changes requested):
Thanks for the fix. Just left a small comment.

poly/src/polynomial/univariate/mod.rs
tjade273: Thanks for the notes, should be better now.

Pratyush (approved):

PR #526, opened 2022-11-30
Fix incorrect square for `Fp128`

Pratyush:
## Description

closes: #XXXX

---

write a little note why.

- [ ] Wrote unit tests

### DIFF
--- a/ff/src/fields/models/fp/montgomery_backend.rs
+++ b/ff/src/fields/models/fp/montgomery_backend.rs
@@ -198,8 +198,7 @@ pub trait MontConfig<const N: usize>: 'static + Sync + Send + Sized {
         if N == 1 {
             // We default to multiplying with `a` using the `Mul` impl
             // for the N == 1 case
-            let temp = *a;
-            Self::mul_assign(a, &temp);
+            *a *= *a;
             return;
         }
         if Self::CAN_USE_NO_CARRY_SQUARE_OPT
@@ -241,12 +240,10 @@ pub trait MontConfig<const N: usize>: 'static + Sync + Send + Sized {
             r.b1[i] = carry;
             carry = 0;
         }
-        r.b1[N - 1] >>= 63;
-        for i in 0..N {
-            r[2 * (N - 1) - i] = (r[2 * (N - 1) - i] << 1) | (r[2 * (N - 1) - (i + 1)] >> 63);
-        }
-        for i in 3..N {
-            r[N + 1 - i] = (r[N + 1 - i] << 1) | (r[N - i] >> 63);
+
+        r.b1[N - 1] = r.b1[N - 2] >> 63;
+        for i in 2..(2 * N - 1) {
+            r[2 * N - i] = (r[2 * N - i] << 1) | (r[2 * N - (i + 1)] >> 63);
         }
         r.b0[1] <<= 1;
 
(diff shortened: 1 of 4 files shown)


ff/src/fields/models/fp/montgomery_backend.rs
```diff
@@ -241,12 +240,10 @@ pub trait MontConfig<const N: usize>: 'static + Sync + Send + Sized {
...
-            r[2 * (N - 1) - i] = (r[2 * (N - 1) - i] << 1) | (r[2 * (N - 1) - (i + 1)] >> 63);
-        }
-        for i in 3..N {
-            r[N + 1 - i] = (r[N + 1 - i] << 1) | (r[N - i] >> 63);
+
+        r.b1[N - 1] = r.b1[N - 2] >> 63;
+        for i in 2..(2 * N - 1) {
+            r[2 * N - i] = (r[2 * N - i] << 1) | (r[2 * N - (i + 1)] >> 63);
         }
         r.b0[1] <<= 1;
```
mmagician, line 248:
```suggestion
```

ff/src/fields/models/fp/montgomery_backend.rs
```diff
@@ -241,12 +240,10 @@ pub trait MontConfig<const N: usize>: 'static + Sync + Send + Sized {
...
-        r.b1[N - 1] >>= 63;
-        for i in 0..N {
-            r[2 * (N - 1) - i] = (r[2 * (N - 1) - i] << 1) | (r[2 * (N - 1) - (i + 1)] >> 63);
-        }
-        for i in 3..N {
-            r[N + 1 - i] = (r[N + 1 - i] << 1) | (r[N - i] >> 63);
+
+        r.b1[N - 1] = r.b1[N - 2] >> 63;
+        for i in 2..(2 * N - 1) {
+            r[2 * N - i] = (r[2 * N - i] << 1) | (r[2 * N - (i + 1)] >> 63);
```
mmagician, line 246:
```suggestion
        for i in 1..(2 * N) {
            r[2 * N - i] = (r[2 * N - i] << 1) | (r[2 * N - (i + 1)] >> 63);
```

mmagician: Is line 244 outside of the loop for performace reasons?
Otherwise we could remove the explicit cases of r[2*N-1] and r[1] before and after the loop and instead do:

```
for i in 1..(2 * N) {
            r[2 * N - i] = (r[2 * N - i] << 1) | (r[2 * N - (i + 1)] >> 63);
```

mmagician: I ran some benchmarks: https://gist.github.com/mmagician/ee80f60e402a9e29ad8dff0a62424edd. As expected the result is pretty inconclusive, given how small the change would be.

Pratyush: Yeah, those were outside the loop for perf reasons. Let me try your version, also comparing the performance against master.

Pratyush: Hm I can detect a small but non-zero perf difference between the master and your version (the diff b/w my version and master is smaller).

mmagician (approved):
I can approve inasmuch as the new tests for fp128 pass.
Maybe it's worth adding a comment to the method explaining which algorithm is used for squaring? While the multiplication seems to follow [this doc](https://hackmd.io/@gnark/modular_multiplication), I can't immediately see the connection to the squaring one.

Pratyush:
We're not using the advanced squaring algorithm from the gnark folks. (I intend to implement that, but haven't found the cycles yet.)

I can't find a reference for our existing squaring algorithm, but it's the one implemented in bellman originally.

Issue #470, opened 2022-09-05
Should arkworks-rs have hooks?


weikengchen:
A random thought related to https://github.com/ingonyama-zk/cloud-ZK/issues/1

Should we consider one day that we will have some hooks for this library? Note that we should not hook every method/function---we can focus on a method/function that has a legit reason to be constantly hooked. One that comes to my mind is hardware acceleration for MSM and/or NTT.

Note that hooks are becoming more and more common in Rust. We have panic hooks. It seems that OOM hooks are coming soon.  
https://github.com/rust-lang/rust/issues/51245

If we implement the hook, there would be an `AtomicPtr` around the method that is hookable, and the corresponding struct/trait will have implementations that allow runtime hooking by modifying this pointer.

weikengchen:
I was thinking whether or not one can redirect `msm_bigint` to another function not this one: https://github.com/arkworks-rs/algebra/blob/fc7dc3e2a226eade8cda02f6eb2305bdd90b1da8/ec/src/scalar_mul/variable_base/mod.rs#L43

without forking the proof system repo, such as Groth16:
https://github.com/arkworks-rs/groth16/blob/sync-algebra/src/prover.rs#L92

weikengchen:
In other words, I am thinking about a "patches-crates.io` for the default implementation in `VariableBaseMSM`.

burdges:
We want to do basically this in polkadot to have fast elliptic curve host calls from WASM.  We'll build wrapper types around the arkworks curve types, which then implement the host calls.  A priori, I'd think hardware acceleration should similarly use wrapper types that add hooks. 

There exist some overhead in that users must add wrapper traits for their protocol constants like base points, but maybe it'd make more sense to simplify this aspect somehow?  We might for example recommend that people provide constants using polymorphic functions instead of wrapper traits, although not sure if this really makes sense.  

You could also just hide hooks behind a feature flag I guess, maybe that's simplest..

burdges:
As an aside, rand avoided hooks in part for performance but also for security so that malicious dependencies could not corrupt the randomness.  I guess these hooks are less of a vulnerability than bad randomness, since malicious dependencies would still need to exfiltrate any secret key material, but bad randomness is an attack all by itself.

weikengchen:
[...]

PR #528, opened 2022-11-30
Allow to overwrite the default implementation of `msm`

achimcc:
## Description

Multi scalar multiplication is usually an expensive operation on different platforms. It is slow in WebAssembly runtimes and can usually get speed up by special hardware. This makes it interesting to define a custom `msm` function in the model which overwrites the default implementation. This custom function could compute the msm by calling into host functions from a WebAssembly runtime or by delegating it to customized hardware.

This PR:

- reimplements `msm` from the `VariableBaseMSM` trait in the `SWCurveConfig` trait, refers there to `msm_unchecked` from the `VariableBaseMSM` trait. This function ois called `model_msm`
- refers to `model_msm` implementation provided by the `SWCurveConfig` trait in the `VariableBaseMSM` implementation for `Projective<P>`
- now the `model_msm` bigint function can get overwritten by curves which are implementing the `SWCurveConfig` trait.

Similar motivation and technique then the already merged PR #534 . As discussed in issue #537.

---

write a little note why.

~~- [ ] Wrote unit tests~~ non-breaking change, all existing tests are still passing.


achimcc:
It would be great if we could pass a custom `bigint_msm` function to the model which overwrites the default implementation!

It looks like this is not the way to go, since it is producing some recursive compile time loop problem?

ec/src/models/short_weierstrass/group.rs
```diff
@@ -634,6 +634,24 @@ impl<P: SWCurveConfig> ScalarMul for Projective<P> {
     }
 }
 
+#[cfg(feature = "specialization")]
+pub trait CustomMSM: SWCurveConfig {
+    fn msm_bigint(
+        bases: &[Affine<Self>],
+        bigints: &[<Self::ScalarField as PrimeField>::BigInt],
+    ) -> Projective<Self>;
+}
```
Pratyush, line 643: We can also try adding this to method to `SWCurveConfig` directly, with a default impl calling the current variable-base MSM code.

burdges:
I suppose ` mul_projective` and `mul_affine` kinda establish this pattern already?  I think default type parameters could maybe help, which always risks breakage if people have the opportunity.  Anyways..

Are there methods besides `mul_projective`, `mul_affine`, `bigint_msm`, and the pairing calls that'll require replacement?  FFTs?  It's also plausible we're best off being able to maintain forks more easily by going outside the trait system.

Pratyush:
Another idea would be to have a global "registry" that contains function pointers for MSMs and FFTs. Something like the following, perhaps:

```rust
pub struct Registry {
    msm: /* type signature for MSM method */
    fft: /* type signature for FFT methods */
}

lazy_static! { REGISTRY: Registry = /* init */ }; // or use OnceCell

// In MSM impl
fn msm(...) {
    REGISTRY.msm(...);
}
```

We will provide default impls for `REGISTRY`, but binaries that use `arkworks` can override this method. Perhaps we can take inspiration from `tracing`'s subscriber infrastructure to allow fine-grained overrides for specific curves only.

Pratyush:
The foregoing design is basically the idea in https://github.com/arkworks-rs/algebra/issues/470

Pratyush:
Actually, it would be nice to have a design that's a hybrid between `tracing` and `rayon`. I.e. we can support both scoped usage: 
```rust
ark_ec::with_msm(msm_fn1, || { /* scoped code that uses `msm_fn1` */});
ark_ec::with_msm(msm_fn2, || { /* scoped code that uses `msm_fn2` */});
```
and global replacement
```rust
ark_ec::set_global_msm(msm_fn1);
/* all subsequent code uses `msm_fn1` */
```

burdges:
A registry incurs some global cost, unless some default feature disables it, but if your talking about ASICs or FPGA then maybe you'd want it for some curves but not others.

Another part of the story would be what/how gadget code depends upon the models? ala your comments in https://github.com/arkworks-rs/curves/pull/76  Assuming "not really" then..

We could override `mul_projective and mul_affine` already, no?
https://github.com/arkworks-rs/algebra/blob/master/ec/src/models/short_weierstrass/mod.rs#L80-L106

I kinda wish there was some co-git tool that helped maintain branch-like repository differences across repositories.  https://github.com/arkworks-rs/algebra/blob/master/ec/src/models/short_weierstrass/group.rs#L637

Anyways..

We could do this specialization manually if we really like this solution, no?

```
pub trait VariableBaseMSM: ScalarMul {
    ...

    /// Optimized implementation of multi-scalar multiplication.
    fn msm_bigint(
        bases: &[Self::MulBase],
        bigints: &[<Self::ScalarField as PrimeField>::BigInt],
    ) -> Self {
        unhooked_msm_bigint(based,bigints)
    }

    /// Raw optimized implementation of multi-scalar multiplication, only for usage by curve models.
    fn unhooked_msm_bigint(
        bases: &[Self::MulBase],
        bigints: &[<Self::ScalarField as PrimeField>::BigInt],
    ) -> Self {
        ... current msm_bigint body ...
    }

    ...
}
```

```
pub trait SWCurveConfig: super::CurveConfig {
    ...

    fn model_msm_bigint(
        bases: &[Affine<Self>],
        bigints: &[<Self::ScalarField as PrimeField>::BigInt],
    ) -> Projective<Self> {
        Projective<Self>::unhooked_msm_bitint(bases,bigints)
    }

    ...
}
```

```
impl<P: SWCurveConfig> VariableBaseMSM for Projective<P> {
    fn msm_bigint(
        bases: &[Self::MulBase],
        bigints: &[<Self::ScalarField as PrimeField>::BigInt],
    ) -> Self {
        P::model_msm_bigint(bases, bigints)
    }
}
```

Pratyush:
> A registry incurs some global cost, unless some default feature disables it, but if your talking about ASICs or FPGA then maybe you'd want it for some curves but not others.

In practice I suspect the overhead of the registry will be negligible compared to the cost of the MSM.

> We could do this specialization manually if we really like this solution, no? ...

Yes, the "hook"/"registry" idea is compatible with this PR, but I think this PR would be redundant if we implemented the former. A further advantage of the registry idea is that only the default (current) MSM impl needs to live in `ark-ec`; the other MSM impls can live in their own crates.

ec/src/models/short_weierstrass/group.rs
achimcc: This was my initial idea! But this was what caused the recursive compile time loop problem. I just pushed the code here: https://github.com/achimcc/algebra/blob/add-method-to-trait/ec/src/models/short_weierstrass/mod.rs#L114 It builds but execution is extremely slow. That is why I changed the implementation to what it is now. I'd assume we would have to actually re-implement the method in `SWCurveConfig` trait to avoid this recursion?

burdges:
> In practice I suspect the overhead of the registry will be negligible compared to the cost of the MSM.

Yes of course, but what I mean is:  What would need hooks? 

It's clear Miller looks and final exps are costly enough to justify hooks this way, but they can be patched without hooks without much code duplication.   I suppose `mul_projective`, `mul_affine`, point preperation, batch inversion, and serialization would not get hooks?  

> the other MSM impls can live in their own crates.

Yes, this sounds more relevant, and I guess answers my question above.  Are the trade offs in MSMs mostly just size?  

Rust has no polymorphic `static mut`s so hooks might need to be declared in a macro used in curve crates.

achimcc:
I adopted my implementation to what @burdges suggested. This way tests are passing, the recursive compile time loop is gone. If we want to implement this solution, the PR would be ready for review. It would definitely drastically cut down the amount of code that we have to fork for our implementation.

ec/src/scalar_mul/variable_base/mod.rs
```diff
@@ -44,6 +44,14 @@ pub trait VariableBaseMSM: ScalarMul {
     fn msm_bigint(
         bases: &[Self::MulBase],
         bigints: &[<Self::ScalarField as PrimeField>::BigInt],
+    ) -> Self {
+        Self::unhooked_msm_bigint(bases, bigints)
+    }
+
+    /// /// Raw optimized implementation of multi-scalar multiplication, only for usage by curve models.
+    fn unhooked_msm_bigint(
```
mmagician, line 52: I don't think we need to introduce an extra wrapper method. That adds an unnecesary layer of abstraction.

ec/src/models/short_weierstrass/mod.rs
```diff
@@ -105,6 +105,14 @@ pub trait SWCurveConfig: super::CurveConfig {
         res
     }
 
+    /// Default implementation for multi scalar multiplication
+    fn model_msm_bigint(
+        bases: &[Affine<Self>],
+        bigints: &[<Self::ScalarField as PrimeField>::BigInt],
+    ) -> Projective<Self> {
+        VariableBaseMSM::unhooked_msm_bigint(bases, bigints)
```
mmagician, line 113:
```suggestion
        VariableBaseMSM::msm_bigint(bases, bigints)
```
That should be enough no?

achimcc: It seems like there is no way reference the default implementation of a function in a trait explicitly. Once we override it, only the overridden method is available. That's the underlying problem.

It might look nicer though to just rename `unhooked_msm_bigint` into `default_msm_bigint`?

achimcc: I renamed `unhooked_msm_bigint` into `default_msm_bigint`.

achimcc:
I changed the design a bit, now we are re-implementing the `msm` function body in the default implementation of the `model_msm` function in the `SWCurveConfig` trait. This also avoids any recursive compile time bugs and removes the superficial hook function.

ec/src/models/short_weierstrass/mod.rs
```diff
@@ -105,6 +105,16 @@ pub trait SWCurveConfig: super::CurveConfig {
         res
     }
 
+    /// Default implementation for multi scalar multiplication
+    fn model_msm(
```
Pratyush, line 109:
```suggestion
    fn msm(
```

ec/src/models/short_weierstrass/group.rs
```diff
@@ -635,7 +635,11 @@ impl<P: SWCurveConfig> ScalarMul for Projective<P> {
     }
 }
 
-impl<P: SWCurveConfig> VariableBaseMSM for Projective<P> {}
+impl<P: SWCurveConfig> VariableBaseMSM for Projective<P> {
+    fn msm(bases: &[Self::MulBase], bigints: &[Self::ScalarField]) -> Result<Self, usize> {
+        P::model_msm(bases, bigints)
```
Pratyush, line 640:
```suggestion
        P::msm(bases, bigints)
```

ec/src/models/short_weierstrass/mod.rs
```diff
@@ -105,6 +105,16 @@ pub trait SWCurveConfig: super::CurveConfig {
         res
     }
 
+    /// Default implementation for multi scalar multiplication
+    fn msm(
+        bases: &[Affine<Self>],
+        scalars: &[Self::ScalarField],
+    ) -> Result<Projective<Self>, usize> {
+        (bases.len() == scalars.len())
```
Pratyush, line 113: IIUC, this is here because calling `VariableBase::msm` results in infinite recursion?

achimcc: Yes, exactly! The previous solution was to redefine in the `VariableBaseMSM`:

```
/// Optimized implementation of multi-scalar multiplication.
    fn msm_bigint(
        bases: &[Self::MulBase],
        bigints: &[<Self::ScalarField as PrimeField>::BigInt],
    ) -> Self {
        unhooked_msm_bigint(based,bigints)
    }

    /// Raw optimized implementation of multi-scalar multiplication, only for usage by curve models.
    fn unhooked_msm_bigint(
        bases: &[Self::MulBase],
        bigints: &[<Self::ScalarField as PrimeField>::BigInt],
    ) -> Self {
        ... current msm_bigint body ...
    }
```

as suggested by @burdges above. But this means to change the `VariableBaseMSM` trait and people looking at this code might not get the reasoning behind it. That's why I favoured re-implement this tiny function body to break the recursive compile time loop.

Pratyush (approved):

Issue #525, opened 2022-11-29
Add constructor for `IsogenyMap`


Pratyush:
Add a constructor for `IsogenyMap`.
```rust
impl</* bounds */> IsogenyMap<'a, DOMAIN, CODOMAIN> {
	const fn new(
	    x_numerator: &'a [BaseField<CODOMAIN>],
	    x_denominator: &'a [BaseField<CODOMAIN>],
	    ...
	) → Self {
	    Self {
	        x_numerator,
	        x_denominator,
	        ...
	    }
	}
}
```

_Originally posted by @Pratyush in https://github.com/arkworks-rs/algebra/pull/517#discussion_r1035169648_

burdges:
IRTF H2C draft says Edwards curves, including BLS12-377, should use Elligator, not WB, so the isogeny you want for the SW form of BSL12-377 is from the Edwards form.

PR #529, opened 2022-11-30
Skalman make isogeny map with new

drskalman:
## Description

This PR is trying to open discussion around issue #525 and hopefully closes it.

closes: #525
---

write a little note why.

- [x ] Targeted PR against correct branch (master)
- [x] Wrote unit tests


drskalman:
@lamafab has done some tests and thinks this might be a rust bug. Quoting them here:

this works
```
trait Trait {
    type AssocType;
}

struct Test<
    A: Trait,
    B: Trait<AssocType = A>,
> {
    some: B,
} 
```
this does NOT:
```
trait Trait {
    type AssocType;
}

struct Test<
    A: Trait,
    // HERE:
    B: Trait<AssocType = A::AssocType>,
    // Or:
    // B: Trait<AssocType = <A as Trait>::AssocType>,
> {
    some: B,
} 
```
I'm really not sure if this is expected behavior or a bug? 🤔

ec/src/hashing/curve_maps/wb/mod.rs
```diff
@@ -32,18 +32,48 @@ pub struct IsogenyMap<
...
+        y_map_denominator: &'a [BaseField<CODOMAIN>],
+    ) -> Self {
+        Self {
+            x_map_numerator,
+            x_map_denominator,
+            y_map_numerator,
+            y_map_denominator,
+            _phantom_domain: PhantomData::<DOMAIN>,
+        }
+    }
```
Pratyush, line 75: We can also limit this to `'static` lifetimes. That might help with the "temporary dropped while borrowed" issue.

drskalman: do you mean to hard code static life time. I think I did that and the error still persists.

mmagician: I tried it too, couldn't find a way around it...

Pratyush:
I think this is a Rust bug. What seems to be the equivalent code does compile: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=51633acd5ecaf330b62a44216c1b7e41

@drskalman do you want to file a bug against rust-lang/rust?

drskalman:
> @drskalman do you want to file a bug against rust-lang/rust?

Sure, I'll do it but I need find a minimal example that it doesn't work beside arkworks :-) I'll try.

Pratyush:
I think you can just make an issue linking the playground and your branch. Re: minimization tips, the main differences that come to mind are
* we're using macros in the constant
* we're doing some relatively heavy computation at compile time, which might be exhausting the const execution time limit.

PR #530, opened 2022-12-01
Add derived impl of squaring for `MontBackend`

Pratyush:
## Description

Implements unrolled squaring algorithm for prime fields via the `MontConfig` macro. Incorporates the fix from #526.

---

write a little note why.

- [ ] Wrote unit tests

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -116,7 +116,9 @@ jobs:
             command: test
             args: "--workspace \
                    --package ark-test-curves \
-                   --all-features"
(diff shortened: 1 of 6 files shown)


mmagician:
1. Where does the unroll actually happen? Is this only due to the attribute macro on the trait definition? If that's the case then already the default impl should be getting unrolled no?
2. What's the advantage of having two implementations, where the default trait impl and the one derived from macro are the same, as far as I can see? I think I might have missed some optimization step.

mmagician (approved):

Issue #549, opened 2022-12-18
MontFp failure 


weikengchen:
When testing FFT over secp256k1's Fr with mixed radix (with 2^6 * 3), there is a weird bug.

```
let t: Fr = MontFp!("111192936301596926984056301862066282284536849596023571352007112326586892541694");
let rou_biguint: BigUint = t.into();
println!("t = {}", rou_biguint);
```

It outputs 111192936301596926984056301862066282284536849596023571352007112326586892541693. Note: it ends at 93 instead of 94.

We suspect that an additional step might be needed in the Montgomery reduction.

PR #509, opened 2022-11-03
Support exactly 64 * K bits

weikengchen:
## Description

This PR removes the requirement that the top bit in a limb must be zero.

---

Before we can merge this PR, please make sure that all the following items have been
checked off. If any of the checklist items are not applicable, please leave them but
write a little note why.

- [x] Targeted PR against correct branch (master)
- [x] Linked to GitHub issue with discussion and accepted design OR have an explanation in the PR that describes this work.
- [x] Updated relevant documentation in the code
- [ ] Added a relevant changelog entry to the `Pending` section in `CHANGELOG.md`
- [ ] Re-reviewed `Files changed` in the GitHub PR explorer

N/A:
- [ ] Wrote unit tests

PR #550, opened 2022-12-18
Fix MontFp issue in fields with 64 * k bits

weikengchen:
## Description

Previously, we changed the logic of field multiplication to support 64 * k bits, as in #509 and #532. 

We did not, however, change the one that affects the `MontFp` macro. This was previously not detected because `MontFp` uses a special, dedicated function to perform the computation (since it cannot use Fp because Fp has not yet been constructed), while the rest is using other functions.

This PR fixes so, in that `MontFp` will also call a function that is aware of the 64 * k issue.

closes: #549 

---

write a little note why.

N/A:
- [ ] Wrote unit tests

### DIFF
--- a/ff/Cargo.toml
+++ b/ff/Cargo.toml
@@ -28,7 +28,7 @@ digest = { version = "0.10", default-features = false, features = ["alloc"] }
 itertools = { version = "0.10", default-features = false }
 
 [dev-dependencies]
-ark-test-curves = { version = "^0.3.0", path = "../test-curves", default-features = false, features = [ "bls12_381_curve", "mnt6_753"] }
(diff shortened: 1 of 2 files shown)


weikengchen:
Randomized tests appear to be difficult to add, as several functions are private (and are intended to be so).

mmagician:
How about we add regression tests for the specific failure case that we know of? I think it's better to have a few tests not following our test-templates than not having them and risking incorrect computation now & in the future.

weikengchen:
Let me add one.

Pratyush (changes requested):

weikengchen:
Update and cc @Pratyush 

- spare bit is now used
- added a test

Pratyush (approved):

PR #556, opened 2022-12-22
Fix #479: Use workspace inheritance for dependencies

### LINKED ISSUE #479: Use workspace inheritance to reduce duplication in crate `Cargo.toml`s
Rust 1.64 supports workspace inheritance, which means that we can put specify common fields such as `LICENSE` files, homepages, common dependencies, etc. all in the *workspace* `Cargo.toml`, instead of repeating these everywhere. 

We should move to this model once 1.64 is two releases old. (That is, we're at 1.66).

Aphoh:
## Description

closes: #479 

---

write a little note why.

- [ ] Wrote unit tests (no new code)
- [ ] Updated relevant documentation in the code (no docs to change)
- [ ] Added a relevant changelog entry to the `Pending` section in `CHANGELOG.md` (does this require a mention in the changelog? Not super relevant to consumers of the library)


ec/Cargo.toml
```diff
@@ -14,25 +14,25 @@ edition = "2021"
 rust-version = "1.57"
```
weikengchen, line 14: @Pratyush just a quick check, this means that our minimal support version is 1.64 right?

Aphoh:
Merged master and pushed a fix that _should_ work for the failing CI item.

weikengchen:
added that we may put it in "the version" immediately after 0.4.0 due to the need to increase MSVP.

Aphoh:
This failing CI item works for me when I run it locally with [`nektos/act`](https://github.com/nektos/act)... I'm not sure why it now says it can't find one of the curves' sub-directories. Any ideas?

ec/Cargo.toml
Pratyush: Yup

Pratyush:
```suggestion
rust-version = "1.64"
```

poly-benches/Cargo.toml
```diff
@@ -0,0 +1,45 @@
+[package]
```
Pratyush, line 1: After this PR was opened, PR #558 was created and merged. That PR deletes the `poly-benches` crate, so this file shouldn't exist. Please delete this file and the enclosing folder =)

Cargo.toml
```diff
@@ -16,6 +16,46 @@ members = [
     "test-templates",
 ]
```
Pratyush, line 18: Let's also use the [`workspace.package`](https://doc.rust-lang.org/cargo/reference/workspaces.html#the-package-table) feature to inherit common things like `contributors`, license files, version numbers, etc.

Pratyush: I think everything except the package name and description can be lifted up

Pratyush (changes requested):
Thanks for this great PR! Left a couple of comments that should be addressed, but otherwise this looks great!

poly-benches/Cargo.toml
Aphoh: Got it! should be removed now. Must have gotten pulled along in merging or something.

Cargo.toml
Aphoh: Got it. `rust-version` was slightly different between crates (between 1.56 and 1.59). I set it in the workspace to 1.59, but this can of course be changed.

PR #558, opened 2022-12-25
Remove poly-benches.

mmaker:
Less code to maintain, more code to take as an example


mmagician (approved):
Nice. This should actually go into 0.4.0, since poly-benches was messing up the releases a little, it'll make the workflow simpler (`cargo release` command doesn't bump the crates which have `publish = false` like poly-benches, and then you have version mismatches).

👍🏼

poly/benches/common.rs
```diff
@@ -1,15 +1,17 @@
-use ark_std::cmp::min;
-
 // Utility function for getting a vector of degrees to benchmark on.
 // returns vec![2^{min}, 2^{min + interval}, ..., 2^{max}], where:
 // interval = log_interval
 // min      = ceil(log_2(min_degree))
 // max      = ceil(log_2(max_degree))
```
Pratyush, line 5:
```suggestion
/// Utility function for getting a vector of degrees to benchmark on.
/// returns `vec![2^{min}, 2^{min + interval}, ..., 2^{max}]`, where:
/// * `interval = log_interval`
/// * `min      = ceil(log_2(min_degree))`
/// * `max      = ceil(log_2(max_degree))`
```

Pratyush (approved):

Issue #533, opened 2022-12-05
Implement improved squaring algorithms


Pratyush:
Implement the optimized squaring algorithm outlined in https://hackmd.io/@gnark/modular_multiplication#Montgomery-squaring.

An example implementation is available here: https://github.com/mratsim/constantine/blob/e47159e4/constantine/arithmetic/limbs_montgomery.nim#L169-L205

mmagician:
WIP

PR #559, opened 2022-12-25
Improved* Montgomery squaring

mmagician:
## Description

*Right now there's no noticable improvement for the optimised algorithm, even a slight regression.

I've also re-worked the previous separated operand scanning method, s.t. it uses a carry with an extra bit, instead of splitting the squaring across three steps (before, documented: https://github.com/arkworks-rs/algebra/commit/c65952388215406a823dd94e5cc89472c8018bae). See [end of chapter 2.3.1 of this thesis](https://www.microsoft.com/en-us/research/wp-content/uploads/1998/06/97Acar.pdf).

It's quite possible that the new `mul_double_add_with_carry_2` method I added isn't fully optimised in its current state, as I haven't got a whole lot of experience with Montgomery arithmetic, and hence we're not seeing any performance improvements. Perhaps if someone manages to point out a better way to do things we might end up with an overall speedup.

Temporarily removed the macro for squaring

closes: https://github.com/arkworks-rs/algebra/issues/533

---

write a little note why.

- [ ] Wrote unit tests


Pratyush:
That's surprising that the new algorithm is worse. But maybe that's to be expected; seems like other libraries don't use these algorithms right now.

mratsim:
There are 2 reasons why I don't use Montgomery squaring as per Gnark:

1. I caught an off by one while fuzzing, in 32-bit mode. Didn't get any in 64-bit mode but I couldn't track it down so ..., see: https://github.com/mratsim/constantine/blob/495ef44f/constantine/math/arithmetic/limbs_montgomery.nim#L365-L397
2. Then I would need to implement the corresponding assembly. And iirc, the base C code was not convincingly faster than multiplication.

PR #560, opened 2022-12-25
should just check the top bit, not two bits

mmagician:
## Description

While working on #559, I realized that the condiditon for mul and squaring are the same (potenitally a copy paste error?). Since we're using [this optimisation](https://hackmd.io/@gnark/modular_multiplication) ([our method docs](https://github.com/arkworks-rs/algebra/blob/3351d0fe834b397344554aa710461dac556bb77d/ff/src/fields/models/fp/montgomery_backend.rs#L142)), we should probably change the condition to match gnark's.

---

write a little note why.

- [ ] Wrote unit tests


weikengchen:
Let me read the code a bit. The reason is that this no-carry implementation was originally for fields where the top two bits are zero (since, in arkworks-rs, previously, the top bit always must be zero). I need to think about whether this is intended.

Pratyush:
We actually haven't implemented the squaring optimization; right now our squaring code only uses the flag when deciding which assembly routine to call.

Also, our code now supports 64 * k bit moduli, so we need to check if the top bit is zero still, I believe

mmagician:
@Pratyush we have now :) https://github.com/arkworks-rs/algebra/pull/559

Pratyush (approved):

Issue #543, opened 2022-12-13
Optimised `sqrt` variants


mmagician:
Currently there are three sqrt algorithms implemented in arkworks:

For even extension fields:
- Complex method (alg. 8) from https://eprint.iacr.org/2012/685.pdf

For odd extension fields:
- Tonelli-Shanks which works for all fieds no matter the modulus
- Specialized Shanks algorithm for case where modulus is 3 mod 4
 
To be still implemented are the remaining specialized algorithms for odd extension fields. See [Figure 1 here](https://eprint.iacr.org/2012/685.pdf) for the algorithm taxonomy.

Odd extension fields:
- [x] 1 mod 16 (Tonelli Shanks)
- [x] 3 mod 4 (Shanks)
- [ ] 9 mod 16
- [ ] 5 mod 8

Preliminary work was done [here](https://github.com/arkworks-rs/algebra/pull/387/).

Pratyush:
@alexander-zw is thinking of working on this, I believe.

alexander-zw:
Tasks:

  1. Port [Joseph's PR](https://github.com/arkworks-rs/algebra/pull/387) to master
  2. Write general unit tests using small simple fields
  3. Update test-curves to use faster algorithms (need to use Sage script), and benchmark compare Tonelli with new ones and with power algorithm
  4. Check for other algorithms from [the paper](https://eprint.iacr.org/2012/685.pdf) that weren't in Joseph's PR
  5. Use macro to precompute instead of requiring Sage script & hardcode

PR #579, opened 2023-01-16
Add field square root algorithms Shanks, Atkin, and Kong

alexander-zw:
## Description

closes: #543 (finishes task #1 but not whole issue)

---

write a little note why.

- [ ] Wrote unit tests


ff/src/fields/sqrt.rs
```diff
@@ -65,14 +65,55 @@ impl LegendreSymbol {
...
 #[non_exhaustive]
 pub enum SqrtPrecomputation<F: crate::Field> {
-    // Tonelli-Shanks algorithm works for all elements, no matter what the modulus is.
+    /// https://eprint.iacr.org/2012/685.pdf (page 12, algorithm 5).
+    /// Tonelli-Shanks algorithm works for all elements, no matter what the modulus is.
+    /// With _q_ as field order, _p_ as characteristic, and _m_ as extension degree:
+    /// * First factor _q - 1 = 2^s t_ where _t_ is odd.
+    /// * `two_adicity` - _s_.
+    /// * `quadratic_nonresidue_to_trace` - _c^t_, with random _c_ such that _c^2^(s - 1) = 1_.
+    /// * `trace_of_modulus_minus_one_div_two` - _(t - 1)/2 + 1_.
```
Pratyush, line 74:
```suggestion
    /// * `trace_of_modulus_minus_one_div_two` - _(t - 1)/2_.
```

ff/src/fields/sqrt.rs
```diff
@@ -84,66 +125,230 @@ impl<F: crate::Field> SqrtPrecomputation<F> {
...
+    let i = elem_b.double() * b;
+    let x = elem_b * (i - F::one());
+
+    Some(x)
+}
+
+fn kong<F: crate::Field>(
+    elem: &F,
+    trace: &F,
+    c: &F,
```
alexander-zw, line 305: I didn't know if this variable had a name

ff/src/fields/sqrt.rs
```diff
@@ -84,66 +125,230 @@ impl<F: crate::Field> SqrtPrecomputation<F> {
...
+    d: &F,
```
alexander-zw, line 306: I considered expanding out `d` into a description of how it's related to `c` but `c_to_order_minus_nine_div_eight` seems too long

ff/src/fields/sqrt.rs
```diff
@@ -65,14 +65,55 @@ impl LegendreSymbol {
...
+        trace: F,
+        c: F,
+        d: F,
+        c_squared: F,
+        char_minus_nine_div_sixteen: &'static [u64],
+        deg_minus_three_div_two_plus_one: usize,
+    },
+    /// In the case of 3 mod 4, we can find the square root via an exponentiation,
+    /// sqrt(a) = a^(p+1)/4. This can be proved using Euler's criterion, a^(p-1)/2 = 1 mod p.
+    PowerCase3Mod4 {
```
alexander-zw, line 116: Didn't find any name for this algorithm so I just called it "power"

ff/src/fields/sqrt.rs
```diff
@@ -65,14 +65,55 @@ impl LegendreSymbol {
...
     },
-    /// To be used when the modulus is 3 mod 4.
-    Case3Mod4 {
+    /// https://eprint.iacr.org/2012/685.pdf (page 9, algorithm 2).
+    /// With _q_ as field order, _p_ as characteristic, and _m_ as extension degree:
+    /// * `char_minus_three_div_four` - _(p - 3)/4_.
+    /// * `deg_minus_three_div_two_plus_one` - _(m - 3)/2 + 1_.
+    ShanksCase3Mod4 {
+        char_minus_three_div_four: &'static [u64],
+        deg_minus_three_div_two_plus_one: usize,
```
Pratyush, line 86: We don't need to have this as a separate field, since `F::extension_degree()` exists.

ff/src/fields/sqrt.rs
```diff
@@ -65,14 +65,55 @@ impl LegendreSymbol {
...
+    },
+    /// https://eprint.iacr.org/2012/685.pdf (page 10, algorithm 3).
+    /// With _q_ as field order, _p_ as characteristic, and _m_ as extension degree:
+    /// * `trace` - _2^(q - 5)/8_.
+    /// * `char_minus_five_div_eight` - _(p - 5)/8_.
+    /// * `deg_minus_three_div_two_plus_one` - _(m - 3)/2 + 1_.
+    AtkinCase5Mod8 {
+        trace: F,
+        char_minus_five_div_eight: &'static [u64],
+        deg_minus_three_div_two_plus_one: usize,
```
Pratyush, line 96: ditto

ff/src/fields/sqrt.rs
```diff
@@ -65,14 +65,55 @@ impl LegendreSymbol {
...
+    /// * `c_squared` - _c^2_.
+    /// * `char_minus_nine_div_sixteen` - _(p - 9)/16_.
+    /// * `deg_minus_three_div_two_plus_one` - _(m - 3)/2 + 1_.
+    KongCase9Mod16 {
+        trace: F,
+        c: F,
+        d: F,
+        c_squared: F,
+        char_minus_nine_div_sixteen: &'static [u64],
+        deg_minus_three_div_two_plus_one: usize,
```
Pratyush, line 112: same

ff/src/fields/sqrt.rs
```diff
@@ -84,66 +125,230 @@ impl<F: crate::Field> SqrtPrecomputation<F> {
...
+                quadratic_nonresidue_to_trace,
+                trace_of_modulus_minus_one_div_two,
+            ),
+            SqrtPrecomputation::ShanksCase3Mod4 {
+                char_minus_three_div_four,
+                deg_minus_three_div_two_plus_one,
+            } => shanks(
+                elem,
+                char_minus_three_div_four,
+                *deg_minus_three_div_two_plus_one,
```
Pratyush, line 140: Same, we should remove the `deg_minus_...` from function arguments

ff/src/fields/sqrt.rs
Pratyush: We can just clarify that in a comment in `d`.

alexander-zw: I already did

ff/src/fields/sqrt.rs
alexander-zw: I was thinking this would save some computation but sure

ff/src/fields/sqrt.rs
```diff
@@ -65,14 +65,49 @@ impl LegendreSymbol {
...
+    /// Tonelli-Shanks algorithm works for all elements, no matter what the modulus is.
+    /// With _q_ as field order, _p_ as characteristic, and _m_ as extension degree:
+    /// * First factor _q - 1 = 2^s t_ where _t_ is odd.
+    /// * `two_adicity` - _s_.
+    /// * `quadratic_nonresidue_to_trace` - _c^t_, with random _c_ such that _c^2^(s - 1) = 1_.
+    /// * `trace_of_modulus_minus_one_div_two` - _(t - 1)/2_.
     TonelliShanks {
         two_adicity: u32,
         quadratic_nonresidue_to_trace: F,
         trace_of_modulus_minus_one_div_two: &'static [u64],
```
alexander-zw, line 78: Just realized my naming is a bit inconsistent with some existing names, this says modulus but I say characteristic. I feel like "characteristic" is more accurate though. Let me know if you think that should be changed.

Antonio95:
Hello! I've been doing some work trying to get the algorithms @alexander-zw implemented to actually be called during sqrt calls on the fields - so far they are implemented, but never called. This is proving very difficult for a couple of reasons that you might have easy solutions to (especially @Pratyush). I'll explain how I view the situation
- The method `sqrt_precomputation` is generic on a `T: MontConfig<N>`, which corresponds to a _prime field_ F_p (p prime). However, the algorithms cited in the paper are eminently for nontrivial extensions of prime fields F_q with q = q^m a prime power. and, crucially, m > 1. Since I saw this caused some confusion above, the _modulus_ of such an F_q is q, whereas its _characteristic_ is p. The two only coincide for prime fields.
- To support my last claim about the requirement m > 1, cf. the main paper https://eprint.iacr.org/2012/685.pdf and have a look at p. 8, which cointains precisely that condition. For instance, algorithm 3 (Atkin's, for the case q = 5 mod 8) contains an m - 3 (which should be positive). Cf. equation (3) too. In my new tests, this causes the program to crash, since m - 3 overflows as an unsigned integer: since `MontConfig` corresponds to F_p, m is automatically 1.
- My approach was to make the `sqrt_precomputation` generic on a `Field`, which does represent a (finite) extension of a finite field. Infact, it is `Field` that the enum `SqrtPrecomputation` is generic on. This change gives us access to things like `characteristic` (p) and `extension_degree` (m). However, it does not give us access to other necessary elements such as the modulus q. Of course, we could just compute q = p^m, but doing so requires switching between `[u64]` and `BigInt`, and the access to methods to do so, and `pow` is a) partially unavailabie in `Field` and b) unavailable at compile time, where we can only use `const fn` and `static`. 

What is the way to go?
- a) Try to make those methods static so that precomputation can be done at compile time. This would be difficult for me and might introduce breaking changes.
- b) Compute things at runtime. This would be less efficient, although it could be done so that, e.g., `q` is only calculated once (by e.g. keeping it inside an Option)
- c) Some other suggestion?

PR #577, opened 2023-01-13
Add `AdditiveGroup` trait + `Field: AdditiveGroup`

mmaker:
## Description

closes: #507

---

Before we can merge this PR, please make sure that all the following items have been
checked off. If any of the checklist items are not applicable, please leave them but
write a little note why.

- [ ] Targeted PR against correct branch (master)
- [ ] Linked to GitHub issue with discussion and accepted design OR have an explanation in the PR that describes this work.
- [ ] Wrote unit tests
- [ ] Updated relevant documentation in the code
- [ ] Added a relevant changelog entry to the `Pending` section in `CHANGELOG.md`
- [ ] Re-reviewed `Files changed` in the GitHub PR explorer

PR #584, opened 2023-01-21
Remove line links from ff/README

mmaker:
Line links are difficult to maintain (and currently most of them are wrong). Removing them and improving the documentation with same links.

### DIFF
--- a/ff/README.md
+++ b/ff/README.md
@@ -16,26 +16,22 @@ This crate contains two types of traits:
 
 The available field traits are:
 
-- [`Field`](https://github.com/arkworks-rs/algebra/blob/master/ff/src/fields/mod.rs#L66) - Interface for a generic finite field.
-- [`FftField`](https://github.com/arkworks-rs/algebra/blob/master/ff/src/fields/mod.rs#L419) - Exposes methods that allow for performing efficient FFTs on field elements.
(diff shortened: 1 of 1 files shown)


Pratyush (commented):
Thanks, I agree this is a good change! Can you check what the links point to when compiled to docs via cargo doc?

mmaker:
hmm it translates into broken links both with prefixed `/` and without.

Pratyush:
Hm in that case there's three options:
(a) link to the docs.rs pages for these traits/structs (e.g. https://docs.rs/ark-ff/0.3.0/ark_ff/fields/trait.Field.html)
(b) provide a github link to the _file_ containing these traits (but not the link)
(c) just skip the links entirely

mmaker:
good point, going for (a)

Pratyush:
Great, thanks! You can see the latest commit in #577 (https://github.com/arkworks-rs/algebra/pull/577/commits/f22cd4abfe15595bededbfd524c89357f02b4dd4) for an example where I do this.

mmaker:
right, even with the latest commit I see how that breaks the readme..
maybe we should split the doc?

mmaker:
ping. @Pratyush how about leaving the readme alone?

PR #593, opened 2023-02-03
feat: make AffineRepr::xy() more flexible

vmx:
## Description

I decided to open a PR instead of an issue, to also illustrate that the impact on the current code base won't be that big (besides it being a breaking change).

---

Currently [`AffineRepr::xy()` returns the field elements as references]. This is only possible if the underlying data structure that implements that traits owns those field elements and can directly have a reference to it.

Though there are cases, where providing those direct references is not possible. For example if the underlying structure has a different layout, or if the fields elements you're returning need to be wrapped in a newtype struct (that's my case).

With this change, `AffineRepr::xy()` now returns owned data. It's more flexible as now the underlying data structure can have any shape as long as it can return those field elements somehow.

[`AffineRepr::xy()` returns the field elements as references]: https://github.com/arkworks-rs/algebra/blob/4e4fc17e69178ab7f929602f17a5e769ebaf34c7/ec/src/lib.rs#L208

closes: #XXXX

---

write a little note why.

- [ ] Wrote unit tests


Pratyush:
Hm can you leave a snippet of your code here? Would like to see an example use case. Also, I believe the curve traits require the underlying data to be owned (ie any type implementing the AffineRepr trait must not contain any non-static references)

vmx:
My actual code looks like this:
```rust
pub struct G1Affine(blstrs::G1Affine);

impl AffineRepr for G1Affine {
…
    fn xy(&self) -> Option<(Self::BaseField, Self::BaseField)> {
        if self.0.is_identity().into() {
            None
        } else {
            Some((Fp(self.0.x()), Fp(self.0.y())))
        }
    }
…
}
```

What it shows: I indeed own the underlying data. Though I'm not able to return a reference. One thing is that [`self.0.x()` doesn't return a reference](https://docs.rs/blstrs/latest/blstrs/struct.G1Affine.html#method.x), but even if it would, I'd need to wrap it in a type (`Fp` in this case) that implements the arkworks traits. I can't do that directly on the `blstrs` types (without forking) as they would be foreign. This `Fp` cannot be a reference as it is only temporarily created.

Pratyush (commented):
This looks good, could you just add an entry in the `CHANGELOG` under the `Breaking changes` section? Thanks!

Pratyush (approved):

vmx:
I've rebased it. If I should also squash the commits into a single one, let me know.

weikengchen (approved):

PR #612, opened 2023-03-02
MSM bench improved

swasilyev:
This produces a more realistic result

## Description

closes: #XXXX

---

write a little note why.

- [ ] Wrote unit tests


bench-templates/src/macros/ec.rs
```diff
@@ -214,8 +214,9 @@ macro_rules! ec_bench {
                     let name = format!("{}::{}", $curve_name, stringify!($Group));
                     let mut rng = ark_std::test_rng();
 
-                    let g = <$Group>::rand(&mut rng).into_affine();
-                    let v: Vec<_> = (0..SAMPLES).map(|_| g).collect();
+                    let v: Vec<_> = (0..SAMPLES)
```
Pratyush, line 217: Let's construct the projective points, and then use batch inversion to get the affine form. Otherwise the benchmark will be too slow just in the setup phase.

burdges:
We should've a 1024 point benchmark maybe?  It'd give something that runs within the timeouts on more common hardware.

bench-templates/src/macros/ec.rs
Pratyush: Ping on this!

Pratyush (approved):

Issue #616, opened 2023-03-11
Optimize BW6 Miller loop


mmagician:
Contrary to the [inline comment](https://github.com/arkworks-rs/algebra/blob/b71caecb0f31f545b8bd7a8062000c7b6d7b190f/ec/src/models/bw6/mod.rs#L58), the Miller loop for BW6 doesn't implement the optimized algorithm that reuses `f_u` in the second loop, and thus allows shortening the second loop.

Reference paper: https://eprint.iacr.org/2020/351.pdf

swasilyev:
Gnark [implements ](https://github.com/ConsenSys/gnark-crypto/blob/8f7ca09273c24ed9465043566906cbecf5dcee91/ecc/bw6-761/pairing.go#L158) an even more sophisticated approach, that is, though, not backwards compatible.

PR #617, opened 2023-03-11
Bw6 miller loop optimization

mmagician:
## Description

As per Algorithm 5 in https://eprint.iacr.org/2020/351.pdf.

TODO: check whether the inverse should apply to `f_u` or `f_1` in case the loop count is negative.
@yelhousni maybe you could help?

Benchmarks:
Before:
```
Pairing for BW6_761/G2 Preparation for BW6_761
                        time:   [732.74 µs 734.78 µs 737.28 µs]
Pairing for BW6_761/Miller Loop for BW6_761
                        time:   [2.5692 ms 2.5765 ms 2.5844 ms]
Pairing for BW6_761/Full Pairing for BW6_761
                        time:   [6.6391 ms 6.6557 ms 6.6730 ms]
```
After:
```
Pairing for BW6_761/G2 Preparation for BW6_761
                        time:   [591.47 µs 592.55 µs 593.67 µs]
Pairing for BW6_761/Miller Loop for BW6_761
                        time:   [2.0311 ms 2.0358 ms 2.0406 ms]
Pairing for BW6_761/Full Pairing for BW6_761
                        time:   [5.9286 ms 5.9474 ms 5.9669 ms]
```
full pairing is ~11% faster.

closes: #616 

---

write a little note why.

- [x] Wrote unit tests


weikengchen (commented):
Do you want to check if it is equivalent to the original one?

mmagician:
We can probably remove the check for whether 2nd Miller loop is negative. The parameter X^2 - X - 1 > 0, unless `P::X=1` I don't know enough about curve construction, but I guess this will never happen.
Also we could basically reuse `P::X`, and its sign, to determine the value and sign of the first loop count. Maybe to even find the 2-NAF of the 2nd count with a `const fn`. What's standing in the way is that X is currently defined as Bigint. Maybe in a next PR I can refactor the BW6 parameters, ideally removing these 4 parameters that are already known/can be computed at compile.

mmagician:
@weikengchen They're indeed equivalent. You can check by substituting the right branches into: https://github.com/mmagician/bw6-comparison.

Pratyush:
Hm the tests seem to be failing right now

mmagician:
@Pratyush The parameters for loop counts changed, so the tests will fail against the current master of arkworks/curves. I've updated both sides today though, and they are all green - I can add a temp patch to Cargo here to point to the compatible branch in curves.
However, one thing to reconsider (as pointed out to me by @swasilyev at some point) is that even though this algorithm is better for a single pairing, it seems that with a multipairing of 3 the methods are equivalent, and likely worsening the performance for 4+, see the last table in https://hackmd.io/@gnark/BW6-761-changes.
There seems to be one method (called 6' in that hackmd) which is **both** backwards compatible, faster for single and for multi-pairings. I could implement that, but won't be able to do it very soon. WDYT, how likely are we going to do multi-pairings of size 4+ vs. single pairings?

Pratyush:
I think multi-pairings often with some frequency, especially for batch-verification.

mmagician:
I ran some benchmarks, and it seems that the new version from this PR outperforms the current master, even for multi-pairings.

Some numbers for BW6-767:
```
optimized:

Pairing for BW6_767/Multi Pairing for BW6_767 with 1 pairs
                        time:   [3.9839 ms 4.0180 ms 4.0593 ms]
Pairing for BW6_767/Multi Pairing for BW6_767 with 5 pairs
                        time:   [8.7305 ms 8.7505 ms 8.7706 ms]
Pairing for BW6_767/Multi Pairing for BW6_767 with 10 pairs
                        time:   [14.508 ms 14.563 ms 14.617 ms]

master:

Pairing for BW6_767/Multi Pairing for BW6_767 with 1 pairs
                        time:   [4.4910 ms 4.5152 ms 4.5490 ms]
Pairing for BW6_767/Multi Pairing for BW6_767 with 5 pairs
                        time:   [11.000 ms 11.046 ms 11.092 ms]
Pairing for BW6_767/Multi Pairing for BW6_767 with 10 pairs
                        time:   [18.986 ms 19.064 ms 19.145 ms]
```

Where the improvement % actually **increases** with number of pairs!

Similar results for BW6-761:
```
optimized:

Pairing for BW6_761/Multi Pairing for BW6_761 with 1 pairs
                        time:   [3.5187 ms 3.5259 ms 3.5333 ms]
Pairing for BW6_761/Multi Pairing for BW6_761 with 5 pairs
                        time:   [8.1297 ms 8.2073 ms 8.3008 ms]
Pairing for BW6_761/Multi Pairing for BW6_761 with 10 pairs
                        time:   [13.810 ms 13.860 ms 13.911 ms]

master:

Pairing for BW6_761/Multi Pairing for BW6_761 with 1 pairs
                        time:   [3.9132 ms 3.9271 ms 3.9413 ms]
Pairing for BW6_761/Multi Pairing for BW6_761 with 5 pairs
                        time:   [9.9494 ms 9.9683 ms 9.9876 ms]
Pairing for BW6_761/Multi Pairing for BW6_761 with 10 pairs
                        time:   [17.237 ms 17.273 ms 17.309 ms]
```

I'm not quite sure what to make of this given the notes in https://hackmd.io/@gnark/BW6-761-changes. I suppose it comes down to the implementation differences across the gnark & arkworks at some different level in the stack.

In any case, I think the numbers suggest we can safely call this PR an improved version.

ec/src/models/bw6/mod.rs
```diff
@@ -114,23 +116,45 @@ pub trait BW6Config: 'static + Eq + Sized {
             })
             .product::<<BW6<Self> as Pairing>::TargetField>();
 
+        let f_u_inv;
+
+        // TODO: is it enough to get the inverse of f_1, or does f_u also need to get inverted?
```
Pratyush, line 121: Do we need to address these `TODO`s before merging?

ec/src/models/bw6/mod.rs
```diff
@@ -114,23 +116,45 @@ pub trait BW6Config: 'static + Eq + Sized {
...
 
                     let bit = Self::ATE_LOOP_COUNT_2[i - 1];
-                    if bit == 1 || bit == -1 {
+                    if bit == 1 {
+                        f *= &f_u;
+                        for &mut (p, ref mut coeffs) in pairs.iter_mut() {
+                            BW6::<Self>::ell(&mut f, &coeffs.next().unwrap(), &p.0);
+                        }
+                    } else if bit == -1 {
+                        f *= &f_u_inv;
```
Pratyush, line 157: The for loops in both cases are the same, so they can be extracted to outside the loop, no?

ec/src/models/bw6/mod.rs
```diff
@@ -114,23 +116,45 @@ pub trait BW6Config: 'static + Eq + Sized {
...
         }
+
+        // f_1(P) = f_(u+1)(P) = f_u(P) * l([u]q, q)(P)
+        let mut f_1 = cfg_chunks_mut!(pairs_1, 4)
+            .map(|pairs| {
+                let mut f = f_u;
+                for (p, coeffs) in pairs.iter_mut() {
+                    BW6::<Self>::ell(&mut f, &coeffs.next().unwrap(), &p.0);
+                }
+                f
```
Pratyush, line 136:
```suggestion
                pairs.iter_mut().fold(f_u, |f, (p, coeffs)|  {
                    BW6::<Self>::ell(&mut f, &coeffs.next().unwrap(), &p.0);
                })
```

ec/src/models/bw6/g2.rs
```diff
@@ -71,23 +81,39 @@ impl<P: BW6Config> From<G2Affine<P>> for G2Prepared<P> {
...
+        let neg_qu;
+
+        // Swap the signs of `qu`, `r` & `neg_qu` if the loop count is negative.
+        if P::ATE_LOOP_COUNT_1_IS_NEGATIVE {
+            qu = -r_affine;
+            neg_qu = r_affine;
+        } else {
+            qu = r_affine;
+            neg_qu = -qu;
+        }
```
Pratyush, line 99:
```suggestion
        // Swap the signs of `qu`, `r` & `neg_qu` if the loop count is negative.
        let (qu, neg_qu) = if P::ATE_LOOP_COUNT_1_IS_NEGATIVE {
            (-r_affine, r_affine)
        } else {
            (r_affine, -r_affine)
        };
```

ec/src/models/bw6/g2.rs
```diff
@@ -71,23 +81,34 @@ impl<P: BW6Config> From<G2Affine<P>> for G2Prepared<P> {
                 ell_coeffs_1.push(r.add_in_place(&q));
             }
         }
+        // TODO: this is probably the slowest part
```
mmagician, line 84: I think we can leave this TODO in, it's potential for even more future optimizations

Pratyush (approved):

PR #624, opened 2023-03-19
BW6 generic final exponentiation

mmagician:
## Description

bilinearity test fails.

---

Before we can merge this PR, please make sure that all the following items have been
checked off. If any of the checklist items are not applicable, please leave them but
write a little note why.

- [ ] Targeted PR against correct branch (master)
- [ ] Linked to GitHub issue with discussion and accepted design OR have an explanation in the PR that describes this work.
- [ ] Wrote unit tests
- [ ] Updated relevant documentation in the code
- [ ] Added a relevant changelog entry to the `Pending` section in `CHANGELOG.md`
- [ ] Re-reviewed `Files changed` in the GitHub PR explorer

PR #633, opened 2023-03-27
Generic pairings for BW6

swasilyev:
Part of the project of getting BW6-767 the outer curve for BLS12-381 into arkworks, supersedes (fixes) https://github.com/arkworks-rs/algebra/pull/624

Implements pairings for the BW6 family in a generic way, following https://yelhousni.github.io/phd.pdf, section 4.4.2

1. Introduces generic BW6 curve parameters: CM lifting cofactors and "trace modulo r"
2. Makes the existing miller loop implementation support both BW6-761 and BW6-767. It conflicts with https://github.com/arkworks-rs/algebra/pull/617. I suggest we adapt the latter later.
3. Implements the hard part of the final exponentiation in generic way ~for curves with "trace modulo r" = 0 (this is BW6-767)~
4. The existing hand-optimized hard part for BW6-761 moved to the curve specification.

---

write a little note why.

- [x] Wrote unit tests


swasilyev:
A fun observation. In the Miller loop computation we are currently free to get rid of 
```
        if Self::ATE_LOOP_COUNT_1_IS_NEGATIVE {
            f_1.cyclotomic_inverse_in_place();
        }
```
For BW6-761 nothing changes as it has `X > 0`. BW6-767 isn't used afaik, so we have freedom here. @yelhousni wdyt

ec/src/models/bw6/mod.rs
```diff
@@ -185,113 +211,97 @@ impl<P: BW6Config> BW6<P> {
...
+            let g = Self::exp_by_x_plus_1(&(f * &d))
+                .cyclotomic_inverse()
+                .unwrap()
+                * &c
+                * &b;
+            // d2 = (ht**2+3*hy**2)//4
+            let d2 = ((P::H_T * P::H_T + 3 * P::H_Y * P::H_Y) / 4) as u32;
+            let d2 = <P::Fp as PrimeField>::BigInt::from(d2);
+            // d1 = (ht-hy)//2
+            let d1 = ((P::H_T - P::H_Y) / 2) as u32;
```
mmagician, line 254: I think you're at a danger of missing out a negative value here.

ec/src/models/bw6/mod.rs
```diff
@@ -185,113 +211,97 @@ impl<P: BW6Config> BW6<P> {
...
+            let d1 = <P::Fp as PrimeField>::BigInt::from(d1);
+            // H = F**d1 * E
+            let h = f.cyclotomic_exp(d1) * &e;
```
mmagician, line 257:
```suggestion
            let d1 = ((P::H_T - P::H_Y) / 2);
            // H = F**d1 * E
            let mut h = f.cyclotomic_exp(&[d1 as u64]);
            if d1 < 0 {
                h.cyclotomic_inverse_in_place();
            }
            h *= *e;
```

ec/src/models/bw6/mod.rs
```diff
@@ -185,113 +211,97 @@ impl<P: BW6Config> BW6<P> {
...
+                * &d;
+            // G = ((F * D)**(u+1)).conjugate() * C * B
+            let g = Self::exp_by_x_plus_1(&(f * &d))
+                .cyclotomic_inverse()
+                .unwrap()
+                * &c
+                * &b;
+            // d2 = (ht**2+3*hy**2)//4
+            let d2 = ((P::H_T * P::H_T + 3 * P::H_Y * P::H_Y) / 4) as u32;
+            let d2 = <P::Fp as PrimeField>::BigInt::from(d2);
```
mmagician, line 252: This is fine as we're taking squares. But you can simply by avoiding the explicit conversion to `BigInt`

mmagician, line 252:
```suggestion
            let d2 = ((P::H_T * P::H_T + 3 * P::H_Y * P::H_Y) / 4) as u64;
```

ec/src/models/bw6/mod.rs
```diff
@@ -185,113 +211,97 @@ impl<P: BW6Config> BW6<P> {
...
+            // d1 = (ht-hy)//2
+            let d1 = ((P::H_T - P::H_Y) / 2) as u32;
+            let d1 = <P::Fp as PrimeField>::BigInt::from(d1);
+            // H = F**d1 * E
+            let h = f.cyclotomic_exp(d1) * &e;
+            // H = H**2 * H * B * G**d2
+            let h = h.square() * &h * &b * g.cyclotomic_exp(d2);
```
mmagician, line 259:
```suggestion
            let h = h.square() * &h * &b * g.cyclotomic_exp(&[d2]);
```

ec/src/models/bw6/mod.rs
```diff
@@ -185,113 +211,97 @@ impl<P: BW6Config> BW6<P> {
...
+            let g = Self::exp_by_x_plus_1(&e) * &fc;
+            // H = G * C
+            let h = g * &c;
+            // I = (G * D)**(u+1) * Fc.conjugate()
+            let i = Self::exp_by_x_plus_1(&(g * &d)) * fc.cyclotomic_inverse().unwrap();
+            // d2 = (ht**2+3*hy**2)//4
+            let d2 = ((P::H_T * P::H_T + 3 * P::H_Y * P::H_Y) / 4) as u32;
+            let d2 = <P::Fp as PrimeField>::BigInt::from(d2);
+            // d1 = (ht+hy)//2
+            let d1 = ((P::H_T + P::H_Y) / 2) as u32;
```
mmagician, line 296: Same here, can cast to `u64` directly and avoid `BigInt` conversion.

mmagician (changes requested):

mmagician (approved):
Looks good 👍🏼

mmagician:
@swasilyev On second thoughts this should go into breaking changes not features section in the CHANGELOG, WDYT?

ec/src/models/bw6/mod.rs
```diff
@@ -159,22 +173,34 @@ impl<P: BW6Config> BW6<P> {
...
+        P::exp_by_x(f) * &f.cyclotomic_inverse().unwrap()
+    }
+
+    fn exp_by_x_minus_1_div_3(f: &Fp6<P::Fp6Config>) -> Fp6<P::Fp6Config> {
+        let u = P::Fp::from(P::X);
+        let one = P::Fp::one();
+        let three = P::Fp::from(3u32);
         if P::X_IS_NEGATIVE {
-            f.cyclotomic_inverse_in_place();
+            let exp = (u + one) / three;
```
Pratyush, line 189: Is there no way to hardcode this? We can avoid a field inversion that way. same for the other branch.

ec/src/models/bw6/mod.rs
```diff
@@ -185,113 +211,101 @@ impl<P: BW6Config> BW6<P> {
...
+            let i = Self::exp_by_x_plus_1(&(g * &d)) * fc.cyclotomic_inverse().unwrap();
+            // d2 = (ht**2+3*hy**2)//4
+            let d2 = ((P::H_T * P::H_T + 3 * P::H_Y * P::H_Y) / 4) as u64;
+            // d1 = (ht+hy)//2
+            let d1 = (P::H_T + P::H_Y) / 2;
+            // J = H**d1 * E
+            let mut j = h.cyclotomic_exp(&[d1 as u64]);
+            if d1 < 0 {
+                j.cyclotomic_inverse_in_place();
+            }
```
Pratyush, line 302: If these are the only places where we use `P::H_T` and `P::H_Y`, it might be worth it to wrap the computations of `d2` and `d1` into helper methods where we can additionally enforce  

(a) that the cast to `u64` works correctly (i.e. via a `try_from`),
(b) that we're correctly handling the d1 < 0 case.

test-templates/src/pairing.rs
```diff
@@ -49,6 +49,19 @@ macro_rules! test_pairing {
...
+                for _ in 0..ITERATIONS {
+                    let rng = &mut test_rng();
+                    let fp_ext = <$Pairing as Pairing>::TargetField::rand(rng);
+                    let gt = <$Pairing>::final_exponentiation(MillerLoopOutput(fp_ext))
+                        .unwrap()
+                        .0;
+                    let r = <$Pairing as Pairing>::ScalarField::MODULUS;
+                    assert!(gt.cyclotomic_exp(r).is_one());
+                }
+            }
```
Pratyush, line 64: Great test!

ec/src/models/bw6/mod.rs
swasilyev: The best way to hardcode that would probably be some kind of an addition chain. So far added it as BW6Config::X_MINUS_1_DIV_3

ec/src/models/bw6/mod.rs
```diff
@@ -159,22 +173,34 @@ impl<P: BW6Config> BW6<P> {
...
     }
 
-    fn final_exponentiation_first_chunk(
-        elt: &Fp6<P::Fp6Config>,
-        elt_inv: &Fp6<P::Fp6Config>,
-    ) -> Fp6<P::Fp6Config> {
+    fn final_exponentiation_easy_part(elt: Fp6<P::Fp6Config>) -> Fp6<P::Fp6Config> {
         // (q^3-1)*(q+1)
-
+        let elt_inv = elt.inverse().unwrap();
```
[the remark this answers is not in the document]
swasilyev: Not really 'cyclotomic'. It's the easy part, so we are not in the cyclotomic subgroup yet, and the conjugation in the quadratic extension of Fp^3 gives us p^3-Frobenius (this is what the original author meant by calling `cyclotomic_inverse_in_place()` in a line below). I tried to clean up the code a bit for clarity. We could use `inverse_in_place()`, but it's currently implemented via `inverse()`.

ec/src/models/bw6/mod.rs
swasilyev: So far `H_T` and `H_Y`  are used only in the final exp, but might appear useful later. Otherwise we could just hardcode these `d1` and `d2` to the Config.

I addressed handling the `d1<0` case, by adding `cyclotomic_exp_signed()`. As for `try_from`, I think `as u64` does exactly what we want for negative `d1`s.

swasilyev:
Btw, i think that currently the change is not breaking, as i moved the existing curve-specific implementation to the curve definition in https://github.com/arkworks-rs/curves/pull/156. Though i would remove it later in another pr.

Pratyush:
I think it's still a breaking change because defining a BW6 curve now requires specifying more constants.

mmagician:
@swasilyev I think the last part is to move the CHANGELOG entry from features to breaking changes: I agree with Pratyush that is indeed a breaking change on the API.

Pratyush (approved):

PR #639, opened 2023-04-21
Make `short_weierstrass::Affine`'s `infinity` flag conditional

Pratyush:
## Description

This PR allows `short_weierstrass::Affine`'s layout to conditionally contain a `bool` flag indicating whether or not it is zero. This is helpful for compressing the representation of the point in memory. Note that due to the unfortunate lack of associated type defaults, we cannot set the type of the flag to be `bool` by default.

In the future, it would be nice to automatically derive this from certain conditions (e.g., `A.is_zero()`) 

---

write a little note why.

- [ ] Wrote unit tests

### DIFF
--- a/ec/src/models/short_weierstrass/mod.rs
+++ b/ec/src/models/short_weierstrass/mod.rs
@@ -2,7 +2,10 @@ use ark_serialize::{
     CanonicalDeserialize, CanonicalDeserializeWithFlags, CanonicalSerialize,
     CanonicalSerializeWithFlags, Compress, SerializationError, Valid, Validate,
 };
-use ark_std::io::{Read, Write};
+use ark_std::{
+    hash::Hash,
+    io::{Read, Write},
+};
 
 use ark_ff::fields::Field;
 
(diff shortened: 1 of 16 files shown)


Pratyush:
Note that I'm not sure this is necessarily a change that we want. We should benchmark and see if there's perf differences.

ec/src/models/short_weierstrass/mod.rs
```diff
@@ -30,6 +33,9 @@ pub trait SWCurveConfig: super::CurveConfig {
     /// Generator of the prime-order subgroup.
     const GENERATOR: Affine<Self>;
 
+    /// A type that is stored in `Affine<Self>` to indicate whether the point is at infinity.
+    type ZeroIndicator: ZeroInd<Self>;
```
Pratyush, line 37: Not really happy with this name; happy to take suggestions for better name.

ec/src/models/short_weierstrass/mod.rs
```diff
@@ -198,3 +204,30 @@ pub trait SWCurveConfig: super::CurveConfig {
         }
     }
 }
+
+pub trait ZeroInd<C: SWCurveConfig>:
```
Pratyush, line 208: Same for this.

weikengchen (approved):

Issue #629, opened 2023-03-22
Remove SWUMap and WBMap, fix MapToCurve semantics


burdges:
Afaik, there is no reason for WB hash-to-curve when SWU hash-to-curve works, right?  Also, There is only one suite listed for each curve in https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-hash-to-curve-16#name-suites-for-hashing  As an aside, they refer to both WB and SWU as SSWU.

There are reasons to expose the underlying map-to-curve as their hash-to-field favors [expand_message_xmd](https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-hash-to-curve-16#name-expand_message_xmd) for sha256 when other likely prefer [expand_message_xof](https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-hash-to-curve-16#name-expand_message_xof) for blake2x, shake, merlin, etc. and even PRNGs

It seemingly follows we should remove the `SWUMap` and `WBMap` types, replace their `MapToCurve` methods by free methods, like
```
pub fn map_to_curve<C>(point: P::BaseField) -> Result<C, HashToCurveError> 
where
    C: CurveGroup,
    <C as CurveGroup>::Config: SWUConfig,
{ ... }
```
and 
```
pub fn map_to_curve<C>(point: P::BaseField) -> Result<C, HashToCurveError> 
where
    C: CurveGroup,
    <C as CurveGroup>::Config: WBConfig,
{ ... }
```
We finally replace the `MapToCurve`  trait by 
```
pub trait MapToCurve: CurveGroup  {
    fn map_to_curve(<Self as CurveGroup>::BaseField) -> Self;
}
```
Individual curves then `impl MapToCurve` by invoking `hashing::swu::map_to_curve` or `hashing::wb::map_to_curve` as appropriate.

Afaik, we've no reason to output an affine point here like we currently do.  In fact, we pointlessly normalize the point both before and after a cofactor multiplication right now.

burdges:
In principle, we could merge `MapToCurve` into `CurveGroup` itself, but we should not define maps without some consensus, which maybe the sage script provides, or maybe not.  Interestingly, polymorphic free fns can be monomorphized as associated constants of traits like https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=8e02c2aea90434583aca18e0bae73459 so you could've an optional method like
```
pub trait CurveGroup {
    ...
[...]

Issue #630, opened 2023-03-22
Idiomatic HashToField


burdges:
Is `construct_dst_prime` from [`ExpanderXof`](https://github.com/arkworks-rs/algebra/blob/master/ff/src/fields/field_hashers/expander/mod.rs#L24) anywhere in the spec?  If not, then maybe we should remove it?  Anyways..

I'd think `ExpanderXof` should become a free `pub fn` like
```
fn update_dst_prime<H, const K: u16>(h &mut H, dst: &[u8]) 
where H: ExtendableOutput+Default+Sized,
{
    if self.dst.len() > MAX_DST_LENGTH {
        let mut long = H::default();
        long.update(&LONG_DST_PREFIX.clone());
        long.update(&self.dst);
        let mut dst = [0u8; MAX_DST_LENGTH];
        long.finalize_xof_into(dst);
        h.update( dst[0..(2 * K + 7) >> 3] );
    } else {
        h.update(dst);
    };
    // I2OSP(len,1) https://www.rfc-editor.org/rfc/rfc8017.txt
    h.update(dst.len() as u8);
}

pub fn expand_xof<H>(domain: &[u8], message: &[u8]) -> Result<<H as ExtendableOutput>::Reader, &'static str>
where H: ExtendableOutput+Default+Sized
{
    if message.len() > u16::MAX as usize {
        return Err("expand_message_xof cannot handle messages longer than 2^16 bytes.");
    }
    let len = message.len() as u16;
    let mut h = H:default();

    h.update(message)
    /// I2OSP(len,2) https://www.rfc-editor.org/rfc/rfc8017.txt
    h.update(len.to_be_bytes())
    update_dst_prime(&mut h,domain);
    h.finalize_xof()
}
```

We could seemingly  `impl<H: Digest+Clone> XofReader for ExpanderXmd<H>` too.  I think [`KeyInit`](https://docs.rs/digest/latest/digest/trait.KeyInit.html#) cannot work here, so we'd provide some free fn like
```
pub fn expand_xmd<H: Digest+Clone>(domain: &[u8], message: &[u8]) -> ExpanderXmd<H> { ... }
```
It's possible the rust crypto project would upstream this somehow even.

At this point, XoFs have become our common currency, so our `HashToField` trait also becomes a free fn like
```
pub fn hash_to_field<F: Field,H: XofReader>(h: &mut H) -> F { .. }
```
As field hashers are specified for each curve, individual curves might specify them via
```
[...]

PR #627, opened 2023-03-20
Pseduo-remove MapToCurve::new, renamed to check_parameters

burdges:
## Description

MapToCurve::new seemingly originates from a more runtime oriented elliptic curve crate: https://github.com/armfazh/redox-ecc/blob/master/src/ellipticcurve.rs#L36
    
Arguably test_parameters should be inherent methods, invoked by whoever defined the parameters, but maybe not?  I left the trait method for now.  I suspect the hash-to-curve code would benefit form a wider more idiomatic rewrite btw.

---

Before we can merge this PR, please make sure that all the following items have been
checked off. If any of the checklist items are not applicable, please leave them but
write a little note why.

- [ ] Targeted PR against correct branch (master)
- [ ] Linked to GitHub issue with discussion and accepted design OR have an explanation in the PR that describes this work.
- [ ] Wrote unit tests
- [ ] Updated relevant documentation in the code
- [ ] Added a relevant changelog entry to the `Pending` section in `CHANGELOG.md`
- [ ] Re-reviewed `Files changed` in the GitHub PR explorer

PR #643, opened 2023-04-25
Refactor hash-to-curve around digest::XofReader trait

burdges:
## Description

Addresses several concerns:

Hash-to-field and hash-to-curve are logically consumers of XoFs.  The hash-to-curve spec is written this way.  Also, ristretto exploits this, likely so do elligator deployments.  It follows polymorphic implementations should expose hashing directly from XoFs, but still push users towards compliant domain separation.  In fact, arkworks' hash-to-curve cannot currently use shake cipher suites, or other XoFs, only sha2 ones, which excludes some elligator curves, and all curves with security levels above 128 bits.

At the same time, the hash-to-curve spec creates messy dependencies across its internal XoF boundary.  In particular, a security parameter plays several roles but depends upon the curve, or even the specific hash-to-curve.  We hide this annoying security parameter in the MapToCurve trait.  It's plausible but uncertain if `digest::core_api::BlockSizeUser` might permit computing this security parameter from the hash selected.

Arkworks' `Field as UniformRand` cannot be made constant-time, so this exposes hashing-to-fields directly.  At present, this hashing-to-fields is not constant-time because arkworks' field arithmetic is not constant-time, but at least one could deploy a protocol built with hash-to-field and later do constant-time field code.

We do not expose hash-to-field in exactly the style of the hash-to-curve spec.  A hash-to-curve for bls12-381 has 128 bits of security, but one could build a curve with higher security with the same base field or whose group order matched the base field, and then require shake256.

Arkworks needs considerable other boilerplate for hash-to-curve.  We make a true default-per-curve hash-to-curve possible.

We preserve roughly the old MapToCurve interface, which already looked suitable for some snark friendly hash-to-curve schemes, although perhaps not within its security assumptions.  It's likely the new XoF interface supports some other  snark friendly hash-to-curve schemes. 

We should discuss this new interface further of course, so no curves PR for now, but doing one looks pretty trivial.

supersedes:  #627 

closes: #629 & #630

---

write a little note why.

- [ ] Wrote unit tests


burdges:
Do you guys want this PR to be from a non-organizational repo so that you can change it?

I could move somewhere else or @mmagician could move it to be internal or something else?

burdges:
I've been too distracted lately but I'll try to get to it next week.  It's really not that hard.

burdges:
Odd, why does https://github.com/arkworks-rs/algebra/pull/643/commits/15fc1a56ccfc8c911372f2411337a0ab04a24138 not appear above now?

PR #588, opened 2023-01-23
GLV parameters structure

simonmasson:
## Description

This PR provides the structure for implementing the GLV scalar multiplication.

- [x] Implementation for the SW curve model
- [ ] Implementation for the TE curve model (could be useful for ed25519, bandersnatch, etc.)
- [x] Tests for the scalar decomposition and the `glv_mul` vs `mul`.
- [ ] General benchmark code in order to compare with `mul`. I couldn't find the way to write them, but I write different benchmarks in [this PR](https://github.com/arkworks-rs/curves/pull/147) in the `curves` repository.
- [ ] A way to integrate `glv_mul` into `mul` once the tests for different curves pass (only if glv is indeed faster).
 
---

Before we can merge this PR, please make sure that all the following items have been
checked off. If any of the checklist items are not applicable, please leave them but
write a little note why.

- [ ] Targeted PR against correct branch (master)
- [ ] Linked to GitHub issue with discussion and accepted design OR have an explanation in the PR that describes this work.
- [ ] Wrote unit tests
- [ ] Updated relevant documentation in the code
- [ ] Added a relevant changelog entry to the `Pending` section in `CHANGELOG.md`
- [ ] Re-reviewed `Files changed` in the GitHub PR explorer

PR #644, opened 2023-04-25
GLV parameters structure, v2

mmagician:
## Description

Building on the work from #588 and introducing slight interface & implementation improvements:
1. Use `[(bool, Self::ScalarField); 4]` instead of separate structs for bools and scalars. (Best would be to use `[(bool, num_bigint::BigInt); 4]` but `const` initialization isn't supported for `BigInt`.
2. Simplify internal impl details by using `num_bigint::BigInt` instead of `BigUint` - this way we don't need to keep track of all the signs explicitly.

Benches:
- without GLV:
```


Arithmetic for Bls12_381::G1/Scalar Multiplication
                        time:   [148.64 µs 148.69 µs 148.73 µs]
```
- with GLV:
```
Arithmetic for Bls12_381::G1/Scalar Multiplication
			time:   [94.551 µs 94.621 µs 94.693 µs]
```

closes: #588 

@simonmasson 

---

write a little note why.

- [ ] Wrote unit tests


ec/src/scalar_mul/glv.rs
```diff
@@ -1,62 +1,163 @@
...
+    /// The coefficients of the endomorphism
+    const ENDO_COEFFS: &'static [Self::BaseField];
+
+    /// The eigenvalue corresponding to the endomorphism.
+    const LAMBDA: Self::ScalarField;
+
+    /// A 4-element vector representing a 2x2 matrix of coefficients the for scalar decomposition, s.t. k-th entry in the vector is at col i, row j in the matrix, with ij = BE binary decomposition of k.
+    /// The entries are the LLL-reduced bases.
+    /// The determinant of this matrix must equal `ScalarField::characteristic()`.
+    const SCALAR_DECOMP_COEFFS: [(bool, Self::ScalarField); 4];
```
Pratyush, line 24:
```suggestion
    const SCALAR_DECOMP_COEFFS: [(bool, <Self::ScalarField as PrimeField>::BigInt); 4];
```

ec/src/scalar_mul/glv.rs
```diff
@@ -1,62 +1,163 @@
...
+        <Self as CurveConfig>::ScalarField,
+        bool,
+    ) {
+        let scalar: BigInt = k.into_bigint().into().into();
+
+        let coeff_bigints: [BigInt; 4] = Self::SCALAR_DECOMP_COEFFS
+            .iter()
+            .map(|x| {
+                BigInt::from_biguint(
+                    x.0.then(|| Sign::Plus).unwrap_or_else(|| Sign::Minus),
```
Pratyush, line 41:
```suggestion
                    x.0.then_some(Sign::Plus).unwrap_or(Sign::Minus),
```

ec/src/scalar_mul/glv.rs
```diff
@@ -1,62 +1,163 @@
...
+        <Self as CurveConfig>::ScalarField,
+        bool,
+        <Self as CurveConfig>::ScalarField,
+        bool,
+    ) {
+        let scalar: BigInt = k.into_bigint().into().into();
+
+        let coeff_bigints: [BigInt; 4] = Self::SCALAR_DECOMP_COEFFS
+            .iter()
+            .map(|x| {
```
Pratyush, line 39: we can directly use `array::map`

test-curves/src/bls12_381/g1.rs
```diff
@@ -71,6 +77,34 @@ pub const G1_GENERATOR_X: Fq = MontFp!("3685416753713387016781088315183077757961
...
+
+    const LAMBDA: Self::ScalarField =
+        MontFp!("52435875175126190479447740508185965837461563690374988244538805122978187051009");
+
+    const SCALAR_DECOMP_COEFFS: [(bool, Self::ScalarField); 4] = [
+        (true, MontFp!("228988810152649578064853576960394133504")),
+        (true, MontFp!("1")),
+        (false, MontFp!("1")),
+        (true, MontFp!("228988810152649578064853576960394133503")),
+    ];
```
Pratyush, line 93: Can we add some references indicating where these constants come from?

test-templates/src/glv.rs
```diff
@@ -0,0 +1,65 @@
...
+        if is_k1_positive && !is_k2_positive {
+            assert_eq!(k1 - k2 * P::LAMBDA, k);
+        }
+        if !is_k1_positive && is_k2_positive {
+            assert_eq!(-k1 + k2 * P::LAMBDA, k);
+        }
+        if !is_k1_positive && !is_k2_positive {
+            assert_eq!(-k1 - k2 * P::LAMBDA, k);
+        }
+        // could be nice to check if k1 and k2 are indeed small.
```
Pratyush, line 29: I think we can do this via https://docs.rs/num-bigint/latest/num_bigint/struct.BigUint.html#method.bits

test-templates/src/glv.rs
```diff
@@ -0,0 +1,65 @@
...
+pub fn glv_projective<P: GLVConfig>() {
+    // check that glv_mul indeed computes the scalar multiplication
+    let mut rng = ark_std::test_rng();
+
+    let g = Projective::<P>::generator();
+    for _i in 0..100 {
+        let k = P::ScalarField::rand(&mut rng);
+
+        let k_g = <P as GLVConfig>::glv_mul_projective(g, k);
+        let k_g_2 = g.mul(k).into_affine();
```
Pratyush, line 48:
```suggestion
        let k_g_2 = g.mul(k);
```

ec/src/scalar_mul/glv.rs
```diff
@@ -1,62 +1,163 @@
...
+        let beta_2 = &scalar * &n12 / &r;
+
+        // b = vector([int(beta[0]), int(beta[1])]) * self.curve.N
+        // b = (β1N11 + β2N21, β1N12 + β2N22) with the signs!
+        //   = (b11   + b12  , b21   + b22)   with the signs!
+
+        // b1
+        let b11 = &beta_1 * &n11;
+        let b12 = &beta_2 * &n21;
+        let b1 = (b11 + b12) % &r;
```
Pratyush, line 67: We can compute this via standard `ScalarField` operations, right?

ec/src/scalar_mul/glv.rs
```diff
@@ -1,62 +1,163 @@
...
+        let [n11, n12, n21, n22] = coeff_bigints;
+
+        let r = BigInt::from(Self::ScalarField::MODULUS.into());
+
+        // beta = vector([k,0]) * self.curve.N_inv
+        // The inverse of N is 1/r * Matrix([[n22, -n12], [-n21, n11]]).
+        // so β = (k*n22, -k*n12)/r
+
+        let beta_1 = &scalar * &n22 / &r;
+        let beta_2 = &scalar * &n12 / &r;
```
Pratyush, line 58: Let's leave a TODO to add multiplication and division methods to `BigInteger` so that we can avoid this allocating code.

ec/src/scalar_mul/glv.rs
```diff
@@ -1,62 +1,163 @@
...
+    const SCALAR_DECOMP_COEFFS: [(bool, Self::ScalarField); 4];
 
     /// Decomposes a scalar s into k1, k2, s.t. s = k1 + lambda k2,
-    fn scalar_decomposition(k: &Self::ScalarField) -> (Self::ScalarField, Self::ScalarField);
+    fn scalar_decomposition(
+        k: Self::ScalarField,
+    ) -> (
+        <Self as CurveConfig>::ScalarField,
+        bool,
+        <Self as CurveConfig>::ScalarField,
```
Pratyush, line 32:
```suggestion
        Self::ScalarField,
        bool,
        Self::ScalarField,
```

ec/src/scalar_mul/glv.rs
```diff
@@ -1,62 +1,163 @@
...
+            k2.sign() == Sign::Plus,
+        )
+    }
+
+    fn endomorphism(p: &Projective<Self>) -> Projective<Self>;
+
+    fn endomorphism_affine(p: &Affine<Self>) -> Affine<Self>;
+
+    fn glv_mul_projective(p: Projective<Self>, k: Self::ScalarField) -> Projective<Self> {
+        let (k1, sgn_k1, k2, sgn_k2) = Self::scalar_decomposition(k);
```
Pratyush, line 93:
```suggestion
        let ((sgn_k1, k1), (sgn_k2, k2)) = Self::scalar_decomposition(k);
```

test-curves/src/bls12_381/g1.rs
```diff
@@ -71,6 +77,34 @@ pub const G1_GENERATOR_X: Fq = MontFp!("3685416753713387016781088315183077757961
...
+
+    fn endomorphism(p: &Projective<Self>) -> Projective<Self> {
+        let mut res = (*p).clone();
```
Pratyush, line 96:
```suggestion
        let mut res = *p;
```

test-curves/src/bls12_381/g1.rs
```diff
@@ -71,6 +77,34 @@ pub const G1_GENERATOR_X: Fq = MontFp!("3685416753713387016781088315183077757961
...
+        res.x *= Self::ENDO_COEFFS[0];
+        res
+    }
+
+    fn endomorphism_affine(p: &Affine<Self>) -> Affine<Self> {
+        let mut res = (*p).clone();
```
Pratyush, line 102:
```suggestion
        let mut res = *p;
```

test-templates/src/glv.rs
```diff
@@ -0,0 +1,65 @@
...
+    AffineRepr, CurveGroup, Group,
+};
+use ark_std::UniformRand;
+
+pub fn glv_scalar_decomposition<P: GLVConfig>() {
+    let mut rng = ark_std::test_rng();
+    for _i in 0..100 {
+        let k = P::ScalarField::rand(&mut rng);
+
+        let (k1, is_k1_positive, k2, is_k2_positive) = <P as GLVConfig>::scalar_decomposition(k);
```
Pratyush, line 15:
```suggestion
        let ((is_k1_positive, k1), (is_k2_positive, k2)) = <P as GLVConfig>::scalar_decomposition(k);
```

Pratyush:
@mmagician thank you for the PR! I have left some comments.

ec/src/scalar_mul/glv.rs
mmagician: yes good catch

ec/src/scalar_mul/glv.rs
mmagician: https://github.com/arkworks-rs/algebra/issues/645

ec/src/scalar_mul/glv.rs
mmagician: But then in definition in `test-curves`:
```
BigInt::new([228988810152649578064853576960394133504]),
```
doesn't work - I'd still need to pass in a full 4-limb BigInt, which is a bit cumbersome.

And unfortunately I can't do `from(...u64)` since `from` isn't constant.

Pratyush: There's a `BigInt!` macro: https://docs.rs/ark-ff/latest/ark_ff/macro.BigInt.html

test-templates/src/glv.rs
```diff
@@ -0,0 +1,65 @@
...
+pub fn glv_affine<P: GLVConfig>() {
+    // check that glv_mul indeed computes the scalar multiplication
+    let mut rng = ark_std::test_rng();
+
+    let g = Affine::<P>::generator();
+    for _i in 0..100 {
+        let k = P::ScalarField::rand(&mut rng);
+
+        let k_g = <P as GLVConfig>::glv_mul_affine(g, k);
+        let k_g_2 = g.mul(k).into_affine();
```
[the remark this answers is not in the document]
mmagician: I think in this case we just override `mul_projective`. But yeah that makes the above test case useless.
We could remove the override for test-curves so it's at least tested here. Given the significant improvement in benchmarks though, I'd be in favour of overriding the defaults in `curves` repo, where these tests would not add value.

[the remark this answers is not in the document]
Pratyush: Hm maybe a middle ground would be to extract the default double-and-add impl into a free `pub fn double_and_add`, call that in the default impl, and then later on we can test correctness vs the free method `pub fn double_and_add`.

test-curves/src/bls12_381/g1.rs
mmagician: Maybe @simonmasson has some scripts? Otherwise I can write some sage later

test-templates/src/glv.rs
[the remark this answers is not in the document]
mmagician: I've also changed `double_and_add`'s methods arguments to take `impl AsRef<[u64]>` instead.

ec/src/scalar_mul/glv.rs
```diff
@@ -1,62 +1,163 @@
...
+            Self::ScalarField::from(k1.to_biguint().unwrap()),
+            k1.sign() == Sign::Plus,
+            Self::ScalarField::from(k2_abs),
+            k2.sign() == Sign::Plus,
+        )
+    }
+
+    fn endomorphism(p: &Projective<Self>) -> Projective<Self>;
+
+    fn endomorphism_affine(p: &Affine<Self>) -> Affine<Self>;
```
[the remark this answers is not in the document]
mmagician: I think it needs at least one copy anyway: https://github.com/arkworks-rs/algebra/blob/83b4d6cd98b8af5b48f1161a56fda0df506f5f46/ec/src/scalar_mul/glv.rs#L82

test-curves/src/bls12_381/g1.rs
mmagician: @Pratyush  https://github.com/arkworks-rs/algebra/pull/644/commits/7d86856f6c6220caad5ad8eedda59057254d1143
It's using a different order than given for the bases than in the source, I might adapt the impl later to match the Guide also with parameter names, but first I'd like to get this merged so it could be tested against the curves implemented in https://github.com/arkworks-rs/curves/pull/147 without breaking anything.

Pratyush (approved):

PR #647, opened 2023-05-03
fix msm comment:the default msm implementation in fact checks length

zhiqiangxu:
The msm method checks that `bases` and `scalars` have the same length.


ValarDragon (approved):
Lol, ty for fix :pray:

mmagician (approved):

PR #659, opened 2023-06-29
Implement Elligator2 hash-to-curve for Twisted Edwards curves

drskalman:
## Description

The IETF standard requires/recommond Elligator 2 hash-to-curve for Twisted Edwards Curve. This PR implement the Elligator2 map-to-curve so such hashing is possible in Arkworks.

- [x] Wrote unit tests


burdges:
"The IETF standard requires/recommends Elligator 2 hash-to-curve for [for any curve with a ] twisted Edwards form", so ed_on_*, bls12-377, etc.

It won't be easily usable without https://github.com/arkworks-rs/algebra/pull/643 or similar of course, given the current interface is basically exclusive to bls12-381.

weikengchen:
Jeff, I would likely take a look at #643 separately, for the reason that that one is a considerably larger PR...

burdges:
Sure.  Also, there are a few small parts of #643 like https://github.com/arkworks-rs/algebra/pull/643/commits/15fc1a56ccfc8c911372f2411337a0ab04a24138 which permit all of this to be moved out of algebra.  That's not ideal long-term, but it'd simplify using hash-to-curve besides bls12-381 now.

drskalman:
>  given the current interface is basically exclusive to bls12-381.

Hmm I wonder why is that. I have implemented a complete example for a TE curves in the test example. no?

drskalman:
@mmagician you are familiar with the code, it is mostly borrowed from the SWU map. You should be able to review it quite easily.

burdges:
> Hmm I wonder why is that. I have implemented a complete example for a TE curves in the test example. no?

What you have here is fine.  It's not possible to make it the default for a curve.  It's also not possible to generate the base field elements using the correct field hasher either (shake).  #643 already solves those issues.  I've already slurped too much up into #643 though, so it's fine to do this part separately.

Pratyush:
What's the status of this?

mmagician:
I think it's ready, I will review it against the spec.

ec/src/hashing/curve_maps/elligator2/mod.rs
```diff
@@ -0,0 +1,268 @@
...
+}
+
+/// Represents the Elligator2 hash-to-curve map defined by `P`.
+pub struct Elligator2Map<P: TECurveConfig>(PhantomData<fn() -> P>);
+
+impl<P: Elligator2Config> MapToCurve<Projective<P>> for Elligator2Map<P> {
+    /// Constructs a new map if `P` represents a valid map.
+    fn new() -> Result<Self, HashToCurveError> {
+        // Verifying that U is a non-square
+        if P::Z.legendre().is_qr() {
```
mmagician, line 34: I would make these sanity checks into `debug_assert` instead.
Actually, I notice that the same can and should be done to `swu` and `wb` maps.

ec/src/hashing/curve_maps/elligator2/mod.rs
```diff
@@ -0,0 +1,268 @@
...
+        }
+
+        // We assume that the Montgomery curve is correct and  as such we do
+        // not verify the prerequisite for applicability of Elligator2 map
+        Ok(Elligator2Map(PhantomData))
+    }
+
+    /// Map an arbitrary base field element `element` to a curve point.
+    /// Based on
+    /// <https://github.com/zcash/pasta_curves/blob/main/src/hashtocurve.rs>.
```
mmagician, line 47:
```suggestion
```

mmagician: I think it's a copy-paste leftover from swu

ec/src/hashing/curve_maps/elligator2/mod.rs
```diff
@@ -0,0 +1,268 @@
...
+                1,
+            )
+        } else {
+            (
+                x2,
+                gx2.sqrt()
+                    .expect("gx2 is a quadratic residue because gx1 is not. Q.E.D"),
+                0,
+            )
+        };
```
mmagician, line 102:
```suggestion
        let (x, y): (P::BaseField, P::BaseField) = if gx1.legendre().is_qr() {
            (
                x1,
                gx1.sqrt()
                    .expect("We have checked that gx1 is a quadratic residue. Q.E.D"),
            )
        } else {
            (
                x2,
                gx2.sqrt()
                    .expect("gx2 is a quadratic residue because gx1 is not. Q.E.D"),
            )
        };
```

ec/src/hashing/curve_maps/elligator2/mod.rs
```diff
@@ -0,0 +1,268 @@
...
+    models::twisted_edwards::{Affine, Projective},
+};
+
+/// Trait defining the necessary parameters for the Elligator2 hash-to-curve method
+/// for twisted edwards curves form of:
+/// `b * y² = x³ + a * x² + x` from [\[WB2019\]]
+/// according to [\[HSSWW22\]]
+///
+/// - [\[BHKL13\]] <http://dx.doi.org/10.1145/2508859.2516734>
+/// - [\[HSSWW22\]] <https://tools.ietf.org/html/draft-irtf-cfrg-hash-to-curve-16>
```
mmagician, line 17:
```suggestion
/// according to [\[HSSWW23\]]
///
/// - [\[BHKL13\]] <http://dx.doi.org/10.1145/2508859.2516734>
/// - [\[HSSWW23\]] <https://datatracker.ietf.org/doc/html/rfc9380>
```

ec/src/hashing/curve_maps/elligator2/mod.rs
```diff
@@ -0,0 +1,268 @@
...
+
+pub trait Elligator2Config: TECurveConfig + MontCurveConfig {
+    /// An element of the base field that is not a square root see \[BHKL13, Section 5\].
+    /// When `BaseField` is a prime field, [\[HSSWW22\]] mandates that `Z` is the
```
mmagician, line 21:
```suggestion
    /// When `BaseField` is a prime field, [\[HSSWW23\]] mandates that `Z` is the
```

ec/src/hashing/curve_maps/elligator2/mod.rs
```diff
@@ -0,0 +1,268 @@
...
+            )
+        };
+
+        let s = x * k;
+        let t = y * k;
+
+        // this is an affine point on the Montgomery curve ideally TECurve
+        // should come with a mapping to its montgomery curve so we could
+        // just call that mapping here, but it seems not being the case, so
+        // we just implement the rational map here from [\[HSSWW22\]] Appendix D
```
mmagician, line 110:
```suggestion
        // we just implement the rational map here from [\[HSSWW23\]] Appendix D
```

mmagician (changes requested):

ec/src/hashing/curve_maps/elligator2/mod.rs
```diff
@@ -0,0 +1,268 @@
...
+    const Z: Self::BaseField;
+}
+
+/// Represents the Elligator2 hash-to-curve map defined by `P`.
+pub struct Elligator2Map<P: TECurveConfig>(PhantomData<fn() -> P>);
+
+impl<P: Elligator2Config> MapToCurve<Projective<P>> for Elligator2Map<P> {
+    /// Constructs a new map if `P` represents a valid map.
+    fn new() -> Result<Self, HashToCurveError> {
+        // Verifying that U is a non-square
```
mmagician, line 33:
```suggestion
        // Verifying that Z is a non-square
```

ec/src/hashing/curve_maps/elligator2/mod.rs
```diff
@@ -0,0 +1,268 @@
...
+
+        // We assume that the Montgomery curve is correct and  as such we do
+        // not verify the prerequisite for applicability of Elligator2 map
+        Ok(Elligator2Map(PhantomData))
+    }
+
+    /// Map an arbitrary base field element `element` to a curve point.
+    /// Based on
+    /// <https://github.com/zcash/pasta_curves/blob/main/src/hashtocurve.rs>.
+    fn map_to_curve(&self, element: P::BaseField) -> Result<Affine<P>, HashToCurveError> {
```
mmagician, line 48: could we be consistent and call it `point` like in the other mappings?

mmagician: Or at least `u` like in RFC 9380

ec/src/hashing/curve_maps/elligator2/mod.rs
mmagician: Actually that's only half-right. It seems that parts of steps 6,7 are missing:
```
        // 6. If is_square(gx1), set x = x1, y = sqrt(gx1) with sgn0(y) == 1.
        // 7. Else set x = x2, y = sqrt(gx2) with sgn0(y) == 0.
```
In fact, when I place an assert in the code:
`        assert_eq!(parity(&y), sgn0, "parity of y is not equal to sgn0(y)");`
both tests fail :(

I think that what we need is to return `sgn0`, then call `parity(&y)` and if they mismatch, negate `y`.

mmagician, line 102:
```suggestion
        let (x, mut y, sgn0): (P::BaseField, P::BaseField, bool) = if gx1.legendre().is_qr() {
            (
                x1,
                gx1.sqrt()
                    .expect("We have checked that gx1 is a quadratic residue. Q.E.D"),
                true,
            )
        } else {
            (
                x2,
                gx2.sqrt()
                    .expect("gx2 is a quadratic residue because gx1 is not. Q.E.D"),
                false,
            )
        };
        
        if parity(&y) != sgn0 {
            y = -y;
        }
```

mmagician (changes requested):

ec/src/hashing/curve_maps/elligator2/mod.rs
mmagician, line 110:
```suggestion
        // `(s, t)` is an affine point on the Montgomery curve.
        // Ideally, the TECurve would come with a mapping to its Montgomery curve, 
        // so we could just call that mapping here.
        // This is currently not supported in arkworks, so
        // we just implement the rational map here from [\[HSSWW22\]] Appendix D
```

mmagician (changes requested):

ec/src/hashing/curve_maps/elligator2/mod.rs
Pratyush: Do we need the types here?

drskalman: No apparently we don't I'll remove them commiting @mmagician suggestion.

ec/src/hashing/curve_maps/elligator2/mod.rs
drskalman: I used point in SWU because pasta implementation used point, but it is not a point per se. We do not impose any topology or geometry on the field, so it is more correct to call it an element (which is going to be mapped to a point on the curve which has geometrical properties). So I found it confusing to call the domain element a point when it is not on a curve. As such, I'm happier to change the name of the parameter in SWU and WB to element instead.

Having said that, if you prefer `point` over `element` I'm not going to dwell on it :-) your call.

drskalman: Done in https://github.com/arkworks-rs/algebra/pull/659/commits/354db69df8048c6ec6a0027c11a03b7f9d7bf15d

ec/src/hashing/curve_maps/elligator2/mod.rs
drskalman: Done in https://github.com/arkworks-rs/algebra/pull/659/commits/354db69df8048c6ec6a0027c11a03b7f9d7bf15d

drskalman:
> y mention this. Otherwise green light.

Done in https://github.com/arkworks-rs/algebra/pull/659/commits/d8667ca7aa2a9d5800128bea3c87c1e1349f694f

ec/src/hashing/curve_maps/elligator2.rs
```diff
@@ -0,0 +1,270 @@
...
+        // 7. Else set x = x2, y = sqrt(gx2) with sgn0(y) == 0.
+        // 8. s = x * K
+        // 9. t = y * K
+        // 10. return (s, t)
+
+        // ark a is irtf J
+        // ark b is irtf k
+        let j = <P as MontCurveConfig>::COEFF_A;
+        let k = <P as MontCurveConfig>::COEFF_B;
+        let j_on_k = j / k;
```
Pratyush, line 60: Can this be precomputed in some way?

ec/src/hashing/curve_maps/elligator2.rs
```diff
@@ -0,0 +1,270 @@
...
+
+        let ksq_inv = k
```
Pratyush, line 62: We are computing k.inverse() twice: once right above and once here.

ec/src/hashing/curve_maps/elligator2.rs
drskalman: @Pratyush I see you have already brought up issue 1. yourself: https://github.com/arkworks-rs/algebra/pull/627#issuecomment-1520918942

Pratyush: I'd be happy to revert the `new` removal; I assumed y'all had come to some consensus on that front internally at Web3.

weikengchen: So I think the consensus is to resume the `new` so that we have opportunities for precomputation? I agree that this is a good approach because otherwise we need to hardcode the parameters, but these parameters are not super easy to compute: https://github.com/FindoraNetwork/noah/blob/develop/crypto/src/hashing_to_the_curve/ed25519/sswu.rs#L27

drskalman: No we don't have consensus on Web3 side as we have hard-to-convince anti-OOP worriers here. That is why I asked to see what have been on Arkworks reviewers mind to sign off on removing `new` and the ability of having longer-than-one-hash data for MapToCurve.

The counter argument is:

  "This is NOT how substrate+frame works. Frame pretends the lifetime of your code is each tx."

I doubt the validity of the statement (Frame pallets could have longer lived objects) but for the sake of argument suppose it is true.

1. Should Arkworks remove the ability of ever storing any runtime customization/pre-computation by any MapToCurve in future because a specific part of a specific implementation of a specific protocol is not able to take advantage of that optimization. 

2. Even in cases like Elligator where the pre-computation is rather simple and perhaps it makes sense to be outsourced to the curve implementor, it is even more important to enforce verification of the pre-computation in `new`, rather than to leave at the mercy of the same curve implementor to call `check_parameters`. We could feature-decorate the check in `new` so it only happens in debug runs. In this way it won't take an efficiency toll where the code is unable to keep the map alive for more one hash.

burdges: Why is there even a discussion?  `const fn` can do this just fine.  Or do we have some required methods which cannot be made cost?

Pratyush: Trait methods are not currently callable in a `const` context.

burdges: Yeah I just looked that up.  oops.

I'd say make `check_parameters` do this
```
if cfg!(debug_assertions) {
    let correct = whatever_computation
    if Self::thing != correct {
        println!("Incorrect parameter should be {}",correct);
    }
}
```

drskalman: Maybe we should first run some benchmarks before we fighting imaginary perfomance foot guns and presumed bottlenecks. This is one-hash (so no keep-alive benefit) before and after merging @burdge 's changes.
```
     Running benches/my_benchmark.rs (~/code/arkworks-algebra/target/release/deps/my_benchmark-1efd55fa18a21b6d)
elligator - with new    time:   [3.1494 µs 3.1551 µs 3.1618 µs]
elligator - no new      time:   [3.9473 µs 3.9711 µs 4.0002 µs]
```
It seems that we have lost 20+% in performance. I'm not sure why and I don't know exactly what they have changed. 

My point is that we had `new` and the option of having a stateful map, it was worknig. There are plenty of other useful stuff we could spend our time on, rather than implementing future-foot-gun prevention measures resulting in 20% loss of our hash perfomance.

Pratyush: Merged.

drskalman:
This is the benchmarking code:
```
    fn hash_arbitary_string_to_curve_elligator2_keep_alive() {
        let test_elligator2_to_curve_hasher = MapToCurveBasedHasher::<
            Projective<TestElligator2MapToCurveConfig>,
            DefaultFieldHasher<Sha256, 128>,
            Elligator2Map<TestElligator2MapToCurveConfig>,
        >::new(&[1])
        .unwrap();

        for i in 0..100000 {
            let to_be_hashed = format!("{}",i);
            let hash_result = test_elligator2_to_curve_hasher.hash(to_be_hashed.as_bytes()).expect("hash should work for all values");
              assert!(
                 hash_result.is_on_curve(),
                 "hash results into a point off the curve"
             );

        }
    }

    fn hash_arbitary_string_to_curve_elligator2_drop_hasher_after_hash() {

        for i in 0..100000 {
            let test_elligator2_to_curve_hasher = MapToCurveBasedHasher::<
                    Projective<TestElligator2MapToCurveConfig>,
                DefaultFieldHasher<Sha256, 128>,
                Elligator2Map<TestElligator2MapToCurveConfig>,
                >::new(&[1])
                .unwrap();
            let to_be_hashed = format!("{}",i);
            let hash_result = test_elligator2_to_curve_hasher.hash(to_be_hashed.as_bytes()).expect("hash should work for all values");
              assert!(
                 hash_result.is_on_curve(),
                 "hash results into a point off the curve"
             );

        }
    }
```
`cargo bench` automatically compiles in `release` so no debug assert. it could be that computing in stack is faster than computing with `const` somehow.

Sure go ahead and revert the removal of the new and I would revert to 74e07bc.

mmagician:
@drskalman I've re-reverted the necessary [commit](https://github.com/arkworks-rs/algebra/pull/696/), so that we end up without `new` parameters. Once that's merged, let's go ahead and adapt this PR to use `const`, we can then merge this.

mmagician:
@drskalman, the revert PR was just merged. We can now go ahead with the changes here, after adapting to the new API (removed `new`)

drskalman:
@mmagician I removed new and changed the interface. It should be ready now.

mmagician (approved):
Thanks @drskalman for quickly handling the changes, green light from my side!

ec/src/hashing/curve_maps/elligator2.rs
```diff
@@ -0,0 +1,297 @@
...
+};
+
+/// Trait defining the necessary parameters for the Elligator2 hash-to-curve method
+/// for twisted edwards curves form of:
+/// `b * y² = x³ + a * x² + x`
+/// from [\[BHKL13\]], according to [\[HSSWW23\]]
+///
+/// - [\[BHKL13\]] <http://dx.doi.org/10.1145/2508859.2516734>
+/// - [\[HSSWW23\]] <https://datatracker.ietf.org/doc/html/rfc9380>
+
```
Pratyush, line 17:
```suggestion
```

ec/src/hashing/curve_maps/elligator2.rs
```diff
@@ -0,0 +1,297 @@
...
+    /// Elliptic Curve defined by y^2 = x^3 + 11*x + 5 over Finite Field of size 101
+    /// sage: EW = EllipticCurve(F101,[11,5])
+    /// sage: EW.order().factor()
+    /// 2^3 * 11
+    /// sage: EW = EdwardsCurve(F101,-1,12)
+    /// sage: EW.gens()[0] * 8
+    /// (5 : 36 : 1)
+    /// Point_WeierstrassToEdwards(F101, 11, 5, F101(5), F101(36), a_given=-1, d_given=12)
+    /// (23, 24)
+
```
Pratyush, line 204:
```suggestion
```

ec/src/hashing/curve_maps/swu.rs
```diff
@@ -24,38 +23,26 @@ pub trait SWUConfig: SWCurveConfig {
...
+            !P::ZETA.legendre().is_qr(),
+            "ZETA should be a quadratic non-residue for the SWU map"
+        );
 
         // Verifying the prerequisite for applicability  of SWU map
-        if P::COEFF_A.is_zero() || P::COEFF_B.is_zero() {
-            return Err(HashToCurveError::MapToCurveError("Simplified SWU requires a * b != 0 in the short Weierstrass form of y^2 = x^3 + a*x + b ".to_string()));
-        }
+        debug_assert!(!P::COEFF_A.is_zero() && !P::COEFF_B.is_zero(),
+		      "Simplified SWU requires a * b != 0 in the short Weierstrass form of y^2 = x^3 + a*x + b ");
```
Pratyush, line 37: Should these checks be reverted to `if`s, or do we want to keep them as `debug_assert`s?

ec/src/hashing/curve_maps/swu.rs
```diff
@@ -99,7 +86,7 @@ impl<P: SWUConfig> MapToCurve<Projective<P>> for SWUMap<P> {
         let gx1_square;
         let gx1;
 
-        assert!(
+        debug_assert!(
```
Pratyush, line 89: What was the motivation of changing this to `debug_assert`?

ec/src/hashing/curve_maps/wb.rs
```diff
@@ -88,9 +88,8 @@ impl<P: WBConfig> MapToCurve<Projective<P>> for WBMap<P> {
     fn check_parameters() -> Result<(), HashToCurveError> {
         match P::ISOGENY_MAP.apply(P::IsogenousCurve::GENERATOR) {
             Ok(point_on_curve) => {
-                if !point_on_curve.is_on_curve() {
-                    return Err(HashToCurveError::MapToCurveError(format!("the isogeny maps the generator of its domain: {} into {} which does not belong to its codomain.",P::IsogenousCurve::GENERATOR, point_on_curve)));
-                }
+                debug_assert!(point_on_curve.is_on_curve(),
```
Pratyush, line 91: Ditto

Pratyush (commented):
Looks mostly good to me, just a few questions about errors and assertions.

ec/src/hashing/curve_maps/swu.rs
mmagician: This check only concerns the pre-computed parameters, there's nothing runtime dependant.

ec/src/hashing/curve_maps/wb.rs
mmagician: This one also is only dependent on parameters derived at compile time

ec/src/hashing/curve_maps/swu.rs
mmagician: This one is a little more tricky, @drskalman maybe can confirm, but it should never happen at this point that the assert fails right?

ec/src/hashing/curve_maps/swu.rs
```diff
@@ -134,14 +121,18 @@ impl<P: SWUConfig> MapToCurve<Projective<P>> for SWUMap<P> {
...
 
         let x_affine = num_x / div;
-        let y_affine = if parity(&y) != parity(&point) { -y } else { y };
+        let y_affine = if parity(&y) != parity(&element) {
+            -y
+        } else {
+            y
+        };
         let point_on_curve = Affine::<P>::new_unchecked(x_affine, y_affine);
-        assert!(
```
mmagician, line 144: Here I'm also not sure

ec/src/hashing/curve_maps/swu.rs
mmagician: So, technically under the current code, it could happen that `div = 0`. [Here](https://github.com/arkworks-rs/algebra/pull/659/files#diff-1334780631de1d3814e5a555c6064b485165e7e7b63620c58214ff9209894b7dR73) we apparently get a non-zero `div`, but that's working off the assumption that `P::ZETA` is non-zero.
The current check  on `ZETA` [here](https://github.com/arkworks-rs/algebra/blob/0c945fa0210d70c17f93a42af3e0a25e7e5b757f/ec/src/hashing/curve_maps/swu.rs#L31) should actually be more strict - currently it allows a zero, whereas according to [RFC9380 Sec. 6.6.2](https://www.rfc-editor.org/rfc/rfc9380.pdf): "Z is non-square in F".
With the stricter condition, this `debug_assert` here is fine instead of `assert`.

ec/src/hashing/curve_maps/swu.rs
mmagician: well ok that just follows from the algorithm

ec/src/hashing/curve_maps/swu.rs
```diff
@@ -24,38 +23,26 @@ pub trait SWUConfig: SWCurveConfig {
...
     /// Checks if `P` represents a valid map.
     fn check_parameters() -> Result<(), HashToCurveError> {
         // Verifying that ZETA is a non-square
-        if P::ZETA.legendre().is_qr() {
-            return Err(HashToCurveError::MapToCurveError(
-                "ZETA should be a quadratic non-residue for the SWU map".to_string(),
-            ));
-        }
+        debug_assert!(
+            !P::ZETA.legendre().is_qr(),
```
mmagician, line 31:
```suggestion
            P::ZETA.legendre().is_qnr(),
```

mmagician: See comment below

ec/src/hashing/curve_maps/swu.rs
Pratyush: I guess my higher-level question was that earlier these checks would happen even in release mode; is it safe for them to happen only in debug mode now?

drskalman: So it should not fail if the map is correct. So, it is safe if the map is correct. However if an implementor choose a wrong constant, I thought it is a good idea that they get the error that they are mapping off the curve.

However, once they got the paramaters right, then we could forgo the check in the interest of performance assuming that they have at least run it once in the debug mode.
Given that it never fails if the parameters are right, we could remove the check completely if you don't like `debug_assert` rather than slowing done every hashing with `assert`. But I thought we can give a hand to the curve developers in debug time so they can fix their maps.

Your call.

Pratyush: Makes sense to me, thanks for the explanation!

Pratyush (approved):

PR #664, opened 2023-07-24
Add "filter" polynomial

Pratyush:
## Description

Adds a selector polynomial `s` parameterized by a group G and (a coset of) a subgroup H, such that, for all `h in H`, `s(h) = 1`, but for all `h in G \ H`, `s(h) = 0`.

Happy to bikeshed on the name.

---

write a little note why.

- [ ] Wrote unit tests


weikengchen:
I would propose an alternative name "subdomainizer" for the reason that "selector" has been used in many contexts like in Plonk, but its value is not necessarily one, and its selection is more complex.

Pratyush:
Good point; maybe we can rename the polynomial to something like "filter polynomial"? I like filter because we're filtering out the elements of the group that are not in the chosen coset. So we can rename the methods to `filter_poly_for`, and they can be used as follows:

```rust
let poly = domain.filter_poly_for(subdomain);
let eval = domain.evaluate_filter_poly_for(subdomain, point);
assert_eq!(poly.evaluate(&point), eval);
```

Pratyush:
Sounds good, I'll make the change later today.

weikengchen (approved):

PR #691, opened 2023-10-23
Implement `Polynomial` for `MultilinearExtension`

mmagician:
## Description

This allows us to implement [`PolynomialCommitment`](https://github.com/arkworks-rs/poly-commit/blob/master/src/lib.rs#L146) directly for MLEs.

@Antonio95 

---

write a little note why.

- [ ] Wrote unit tests

### DIFF
--- a/poly/src/evaluations/multivariate/multilinear/dense.rs
+++ b/poly/src/evaluations/multivariate/multilinear/dense.rs
@@ -1,6 +1,9 @@
 //! Multilinear polynomial represented in dense evaluation form.
 
-use crate::evaluations::multivariate::multilinear::{swap_bits, MultilinearExtension};
+use crate::{
+    evaluations::multivariate::multilinear::{swap_bits, MultilinearExtension},
+    Polynomial,
+};
 use ark_ff::{Field, Zero};
 use ark_serialize::{CanonicalDeserialize, CanonicalSerialize};
 use ark_std::{
@@ -304,6 +307,18 @@ impl<F: Field> Zero for DenseMultilinearExtension<F> {
 
     fn is_zero(&self) -> bool {
         self.num_vars == 0 && self.evaluations[0].is_zero()
+    }
+}
+
+impl<F: Field> Polynomial<F> for DenseMultilinearExtension<F> {
+    type Point = Vec<F>;
+
+    fn degree(&self) -> usize {
+        1
+    }
+
+    fn evaluate(&self, point: &Self::Point) -> F {
+        MultilinearExtension::<F>::evaluate(self, point).unwrap()
     }
 }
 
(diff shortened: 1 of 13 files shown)


Pratyush (approved):

mmagician:
One unfortunate consequence of this change is that now an MLE type has two implementations of `evaluate`, from `MultilinearExtension` and from `Polynomial` and we need to disambiguate the call, as seen e.g. here: https://github.com/arkworks-rs/algebra/pull/691/commits/f8e0378e2ee6dff9fd138ab23975f791c351c724.

I think instead we can remove the `evaluate` method from `MultilinearExtension`, since both `Dense` and `Sparse` implement `Polynomial`, WDYT @Pratyush? This is done in https://github.com/arkworks-rs/algebra/pull/691/commits/14eb52abda22c4aad5fb965e888a458c8495f738

poly/src/evaluations/multivariate/multilinear/sparse.rs
```diff
@@ -227,6 +219,19 @@ impl<F: Field> Index<usize> for SparseMultilinearExtension<F> {
     }
 }
 
+impl<F: Field> Polynomial<F> for SparseMultilinearExtension<F> {
+    type Point = Vec<F>;
+
+    fn degree(&self) -> usize {
```
Pratyush, line 225: It seems that elsewhere in the crate, we assume that `degree` here represents *total* degree, and not individual degree. This is in agreement with [mathematical convention](https://en.wikipedia.org/wiki/Degree_of_a_polynomial#Extension_to_polynomials_with_two_or_more_variables).

So this should be changed to return `self.num_vars`, and it might be worth adding an `individual_degree` method to `Polynomial` that would return the degree for univariate polynomials, `1` for multilinear polynomials, and the actual individual degree for general multivariate polynomials.

poly/src/evaluations/multivariate/multilinear/dense.rs
```diff
@@ -307,9 +284,40 @@ impl<F: Field> Zero for DenseMultilinearExtension<F> {
     }
 }
 
+impl<F: Field> Polynomial<F> for DenseMultilinearExtension<F> {
+    type Point = Vec<F>;
+
+    fn degree(&self) -> usize {
+        1
+    }
```
Pratyush, line 292: Ditto

poly/src/evaluations/multivariate/multilinear/sparse.rs
mmagician: I adapted this to represent the total degree. I've also created a separate issue to discuss the exact design and the need for `individual_degree`: https://github.com/arkworks-rs/algebra/issues/692.

Antonio95: Is this method meant to represent a total degree _bound_, or the actual total degree of the particular instance this is being called on? For instance, the polynomial `f(x, y, z) = xy + yz` has `num_vars` equal to 3, but actual total degree 2.

Returning the actual total degree would be simple in the sparse case, but might require some work in the dense case if the polynomial is stored in evaluation form.

Incidentally, this caveat would also have applied to the code as it was before @Pratyush 's comment: we were always returning 1, but in the fringe case of constant polynomials, that did not match the total degree.

PR #713, opened 2023-12-13
Implementation of bitwise operations between `BigInteger`

hdvanegasm:
## Description

I implemented bitwise operations between BigInteger: AND, XOR, NOT, and OR. In the modifications, the trait BigInteger now implements the necessary traits to support the mentioned bitwise operations, and I included the implementations for those operations for the BigInt struct. Also, I wrote some tests for this functionality.

This PR doesn't close any issue. It's just a feature that I found missing in the library.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 - [\#689](https://github.com/arkworks-rs/algebra/pull/689) (`ark-serialize`) Add `CanonicalSerialize` and `CanonicalDeserialize` impls for `VecDeque` and `LinkedList`.
 - [\#693](https://github.com/arkworks-rs/algebra/pull/693) (`ark-serialize`) Add `serialize_to_vec!` convenience macro.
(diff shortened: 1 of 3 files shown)


mmaker:
in the unittests, any chance you could add some checks on the higher limbs (right now you're testing only on the least significant) and de morgan's law ?

mmaker (approved):
Looks good to me now! A few notes copy-pasted from my commit message:

- It's not clear to me that BigInteger should represent integers mod 2^(N*64), so I'm not ready to take responsability for supporting bitwise  NOT in the API of BigInteger (it's however in BigInt)

- I'm also not sure bitwise operations like >> should be treated differently. I think we should add support for Shr in the API and deprecate div2() and divn() perhaps in a different PR?


perhaps another maintainer can have a look at this?

z-tech (approved):
These changes look safe to me. Shr and such would be nice (related) features and cleaning up any duplicated capability seems reasonable.

z-tech:
Thanks btw @hdvanegasm!

Issue #714, opened 2023-12-14
Implement `core::ops::Shr for BigInt` in place of `BigInt::divn()`


mmaker:
https://github.com/arkworks-rs/algebra/blob/master/ff/src/biginteger/mod.rs#L434 
The BigInt::divn function should instead be done with overloading using the trait Shr from the standard library. 
(Cf. #713)

mmaker:
@hdvanegasm want to give it a shot? :)

hdvanegasm:
I would like to implement not only `core::ops::Shr` but also `core::ops::Shl`. If I will implement `core::ops::Shl` for `BigInt<N>` by moving it `M` bits, should I return a `BigInt<N + M>` or should I just return a BigInt<N> and chopping the overflow?

mmaker:
This issue is just for `Shr` :)

PR #736, opened 2023-12-19
Deprecated `divn` and implemented `core::ops::Shr`

hdvanegasm:
## Description

This PR closes https://github.com/arkworks-rs/algebra/issues/714. I choose to add `#deprecate` annotation to `divn()` instead of removing it to keep the backward compatibility. Let me know if it's best to remove such old function.


hdvanegasm:
Can someone help me with the "CI / Test against curves (bls12_377)" test? AFAIK, I modified the files that the check is complaining about as you can see in the "Files Changed" tab. Why is it failing?

mmaker:
can you please make two separate pull requests for `Shr` and `Shl`?

hdvanegasm:
Done! you can consider this as the PR for `core::ops::Shr` in place of `divn()`. Unfortunatelly, the check that was failing before still fails. Do you know why is this?

z-tech:
> Done! you can consider this as the PR for `core::ops::Shr` in place of `divn()`. Unfortunatelly, the check that was failing before still fails. Do you know why is this?

Hi @hdvanegasm, we saw that this check fails because you've deprecated the functions described by the error message. You must change these to the new replacements and the check will pass

hdvanegasm:
Hi @z-tech. Thanks for your reply. Notice that the check is complaining about lines that where changed in this PR. An example of this is presented here: https://github.com/arkworks-rs/algebra/pull/736/files#diff-7bfbb2ccb1c569a529a64dfba848ae83498a02f8f3012fbd7b2c87f0c11c904eR830. You may take a look at the changed files and you will find that all the lines that the check is showing are changed already for the operator `>>`. Let me know if I'm wrong in this 😄

hdvanegasm:
Is it possible that this check is failling because the CI file is pointing to an old version of the repo? Take a look at this: https://github.com/arkworks-rs/algebra/blob/d9527c832888ffbd9231999b30a700ce3e572c16/.github/workflows/ci.yml#L420

z-tech:
> Is it possible that this check is failling because the CI file is pointing to an old version of the repo? Take a look at this:
> 
> https://github.com/arkworks-rs/algebra/blob/d9527c832888ffbd9231999b30a700ce3e572c16/.github/workflows/ci.yml#L420

I see. If you remove this does it work? If yes, it is safe to remove we have archived this repo.

ff/src/biginteger/mod.rs
```diff
@@ -729,6 +729,38 @@ impl<B: Borrow<Self>, const N: usize> BitOr<B> for BigInt<N> {
     }
 }
 
+impl<const N: usize> Shr<u32> for BigInt<N> {
+    type Output = Self;
+
+    fn shr(self, mut rhs: u32) -> Self::Output {
```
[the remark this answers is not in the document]
hdvanegasm: I thikn that it should be "(...) will NOT return an underflow error if the value being shifted is larger than N * 64", given that the limb size 64 for the case of BigInt.

I didn't document the trait definition because we are forcing the implementor to specify the `shr()` function. So, I didn't find an appropriate place to put this message. Let me know if it's fine like it is or if you have a proper place to put this message in the trait definition.

hdvanegasm:
> > Is it possible that this check is failling because the CI file is pointing to an old version of the repo? Take a look at this:
> > https://github.com/arkworks-rs/algebra/blob/d9527c832888ffbd9231999b30a700ce3e572c16/.github/workflows/ci.yml#L420
> 
> I see. If you remove this does it work? If yes, it is safe to remove we have archived this repo.

I removed the check completely. Let me know if this is what you were looking for.

ff/src/biginteger/mod.rs
```diff
@@ -729,6 +729,44 @@ impl<B: Borrow<Self>, const N: usize> BitOr<B> for BigInt<N> {
...
+                let a = &mut data[N - i - 1];
+                let t2 = *a << (64 - rhs);
+                *a >>= rhs;
+                *a |= t;
+                t = t2;
+            }
+        }
+        Self::new(data)
+    }
+}
```
Pratyush, line 768: Let's add an impl of `ShrAssign`

```suggestion
impl<const N: usize> ShrAssign<u32> for BigInt<N> {
    /// Computes the bitwise shift right operation in place.
    ///
    /// Differently from the built-in numeric types (u8, u32, u64, etc.) this
    /// operation does *not* return an underflow error if the number of bits
    /// shifted is larger than N * 64. Instead the result will be saturated to
    /// zero.
    fn shr_assign(&mut self, mut rhs: u32) {
        if rhs >= (64 * N) as u32 {
            return Self::from(0u64);
        }

        while rhs >= 64 {
            let mut t = 0;
            for i in 0..N {
                core::mem::swap(&mut t, &mut self[N - i - 1]);
            }
            rhs -= 64;
        }

        if rhs > 0 {
            let mut t = 0;
            for i in 0..N {
                let a = &mut self[N - i - 1];
                let t2 = *a << (64 - rhs);
                *a >>= rhs;
                *a |= t;
                t = t2;
            }
        }
	}
}

impl<const N: usize> Shr<u32> for BigInt<N> {
    type Output = Self;

    /// Computes bitwise shift right operation.
    ///
    /// Differently from the built-in numeric types (u8, u32, u64, etc.) this
    /// operation does *not* return an underflow error if the number of bits
    /// shifted is larger than N * 64. Instead the result will be saturated to
    /// zero.
    fn shr(mut self, mut rhs: u32) -> Self::Output {
        self >>= rhs;
        self
    }
}
```

ff/src/biginteger/mod.rs
```diff
@@ -809,6 +847,7 @@ pub trait BigInteger:
     + for<'a> BitOrAssign<&'a Self>
     + BitOr<Self, Output = Self>
     + for<'a> BitOr<&'a Self, Output = Self>
+    + Shr<u32, Output = Self>
```
Pratyush, line 850:
```suggestion
    + Shr<u32, Output = Self>
    + ShrAssign<u32>
```

curves/bls12_381/src/fields/tests.rs
```diff
@@ -875,7 +875,7 @@ fn test_fq_repr_divn() {
             0x0,
         ])
     );
-    a.divn(64);
+    a = a >> 64;
```
Pratyush, line 878:
```suggestion
    a >>= 64;
```

curves/bls12_381/src/fields/tests.rs
```diff
@@ -863,7 +863,7 @@ fn test_fq_repr_divn() {
             0x244,
         ])
     );
-    a.divn(130);
+    a = a >> 130;
```
Pratyush, line 866:
```suggestion
    a >>= 130;
```

curves/bls12_381/src/fields/tests.rs
```diff
@@ -851,7 +851,7 @@ fn test_fq_repr_divn() {
             0x912f9480d7528a7,
         ])
     );
-    a.divn(50);
+    a = a >> 50;
```
Pratyush, line 854:
```suggestion
    a >>= 50;
```

curves/bls12_381/src/fields/tests.rs
```diff
@@ -839,7 +839,7 @@ fn test_fq_repr_divn() {
             0x1225f2901aea514e,
         ])
     );
-    a.divn(1);
+    a = a >> 1;
```
Pratyush, line 842:
```suggestion
    a >>= 1;
```

curves/bls12_381/src/fields/tests.rs
```diff
@@ -827,7 +827,7 @@ fn test_fq_repr_divn() {
         0x16080cf4071e0b05,
         0x1225f2901aea514e,
     ]);
-    a.divn(0);
+    a = a >> 0;
```
Pratyush, line 830:
```suggestion
    a >>= 0;
```

ec/src/scalar_mul/variable_base/mod.rs
```diff
@@ -197,7 +197,7 @@ fn msm_bigint<V: VariableBaseMSM>(
 
                     // We right-shift by w_start, thus getting rid of the
                     // lower bits.
-                    scalar.divn(w_start as u32);
+                    scalar = scalar >> w_start as u32;
```
Pratyush, line 200:
```suggestion
                    scalar >>= w_start as u32;
```

ff/src/biginteger/mod.rs
```diff
@@ -1,4 +1,4 @@
-use core::ops::{BitAnd, BitAndAssign, BitOr, BitOrAssign, BitXor, BitXorAssign, Not};
+use core::ops::{BitAnd, BitAndAssign, BitOr, BitOrAssign, BitXor, BitXorAssign, Not, Shr};
```
Pratyush, line 1:
```suggestion
use core::ops::{BitAnd, BitAndAssign, BitOr, BitOrAssign, BitXor, BitXorAssign, Not, Shr, ShrAssign};
```

Pratyush (commented):
Small fixes

ff/src/biginteger/mod.rs
```diff
@@ -729,6 +729,54 @@ impl<B: Borrow<Self>, const N: usize> BitOr<B> for BigInt<N> {
...
+            for i in 0..N {
+                let a = &mut self[N - i - 1];
+                let t2 = *a << (64 - rhs);
+                *a >>= rhs;
+                *a |= t;
+                t = t2;
+            }
+        }
+	}
+}
```
Pratyush, line 763:
```suggestion
        if rhs >= (64 * N) as u32 {
            *self = Self::from(0u64);
        }

        while rhs >= 64 {
            let mut t = 0;
            for i in 0..N {
                core::mem::swap(&mut t, &mut self.0[N - i - 1]);
            }
            rhs -= 64;
        }

        if rhs > 0 {
            let mut t = 0;
            for i in 0..N {
                let a = &mut self.0[N - i - 1];
                let t2 = *a << (64 - rhs);
                *a >>= rhs;
                *a |= t;
                t = t2;
            }
        }
	}
}
```

Pratyush (approved):

Issue #432, opened 2022-06-30
Refactor FixedBase 


Pratyush:
New API for FixedBase scalar multiplication (from meeting)

```rust
pub trait AffineCurve {
    fn precompute_table(&self, scalar_size: usize, window_size: usize) -> Preprocessing<Self>; // name TBD.
}

// elsewhere
impl<G: AffineCurve> Preprocessing<G> {
	// TODO: resolve how to handle parallel vs non-parallel cases.
    #[cfg(feature = "parallel")
    fn mul(&self, scalars: impl IntoParallelIterator<Item = T::ScalarField>) -> Vec<G>;
    
    #[cfg(not(feature = "parallel"))]
    fn mul(&self, scalars: impl IntoIterator<Item = T::ScalarField>) -> Vec<G>;
}
```

PR #746, opened 2024-01-10
`FixedBase` refactor: rename `msm` -> `mul`, declutter the interface and hide aux methods

mmagician:
## Description

I don't think it's worth doing a bigger refactor proposed in https://github.com/arkworks-rs/algebra/issues/432 - yet still the two MSMs we had were confusing people. This PR is sort of bridging this gap of having a documented, simple interface that is clearly distinct from `VariableBaseMSM`.

I renamed `FixedBase::msm` into `FixedBase::mul` and introduces the following refactors:
- Make all methods except for `mul` private. (Edit: `get_window_table` also remains public)
- Simplify the `mul` interface: takes `g` and list of scalars as the parameters. The rest of the parameters can be inferred, and the internal methods called from `mul` directly.
- Add a description and add a doctest

@msmirnov18

closes: #432

---

write a little note why.

- [ ] Wrote unit tests


mmagician:
Actually, I think `get_window_table` should remain public e.g. for [Groth16](https://github.com/arkworks-rs/groth16/blob/42b38f1675fd45aa4429a2d335653e37507ec95c/src/generator.rs#L145).

Pratyush:
What do you think about just moving this to the `ScalarMul` trait? I think since we're doing a breaking change, a larger refactor would be nice too. I can handle it if you'd like.

mmagician:
Sure, we can move it. I'm open to other refactors too, the question is how useful this method is.

mmagician:
What other refactors are you thinking of, actually?

ec/src/scalar_mul/fixed_base.rs
```diff
@@ -9,19 +9,16 @@ use super::ScalarMul;
...
             super::ln_without_floats(num_scalars)
         }
     }
 
-    pub fn get_window_table<T: ScalarMul>(
-        scalar_size: usize,
-        window: usize,
-        g: T,
-    ) -> Vec<Vec<T::MulBase>> {
+    pub fn get_window_table<T: ScalarMul>(window: usize, g: T) -> Vec<Vec<T::MulBase>> {
```
Pratyush, line 20: I don't think we should get rid of `scalar_size`, because there might be some cases where this is smaller than `T::ScalarField::MODULUS_BIT_SIZE`, and constructing the larger table is wasted effort.

Pratyush:
> Sure, we can move it. I'm open to other refactors too, the question is how useful this method is.

In general this method is useful whenever you want to multiply the same element by different scalars. This is useful primarily in setup methods for, e.g., Groth16 or KZG, but only in testing.

So I'd like to keep it, and just make it easier to use. 

> 
What other refactors are you thinking of, actually, aside from moving?

I'll take a stab at these changes, and then we can compare and see if they're worth it.

Pratyush:
@mmagician I've added my changes; let me know if there's anything further you'd like to change.

mmagician:
Looks good, I'd approve if I weren't the PR author 👍🏼 Thanks for the changes!

Pratyush (approved):
Thanks!

PR #750, opened 2024-01-12
Expanding description for SparsePolynomial::from_coefficients_vec

msmirnov18:
## Description

closes: #XXXX

---

write a little note why.

- [ ] Wrote unit tests

### DIFF
--- a/poly/src/polynomial/univariate/sparse.rs
+++ b/poly/src/polynomial/univariate/sparse.rs
@@ -233,6 +233,26 @@ impl<F: Field> SparsePolynomial<F> {
     }
 
     /// Constructs a new polynomial from a list of coefficients.
+    /// Since the function does not automatically combine like terms, the input *has to be* in the
+    /// correct form: no multiple monomials of the same degree.
+    /// # Example
+    /// ```
+    /// use ark_ff::fields::{Fp64, MontBackend, MontConfig};
+    /// use ark_poly::polynomial::univariate::SparsePolynomial;
+    ///
+    /// #[derive(MontConfig)]
+    /// #[modulus = "101"]
+    /// #[generator = "2"]
+    /// pub struct FieldConfig;
+    /// pub type FF = Fp64<MontBackend<FieldConfig, 1>>;
+    ///
+    /// // Creating the polynomial 2*x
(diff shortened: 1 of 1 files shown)


poly/src/polynomial/univariate/sparse.rs
```diff
@@ -233,6 +233,26 @@ impl<F: Field> SparsePolynomial<F> {
...
+    /// pub struct FieldConfig;
+    /// pub type FF = Fp64<MontBackend<FieldConfig, 1>>;
+    ///
+    /// // Creating the polynomial 2*x
+    /// let p1 = SparsePolynomial::from_coefficients_vec(vec![(1, FF::from(2))]);
+    /// // Creating the polynomial x + x
+    /// let p2 = SparsePolynomial::from_coefficients_vec(vec![(1, FF::from(1)), (1, FF::from(1))]);
+    /// // Checking if x + x is equal to 2*x leads to panick.
+    /// assert_eq!(p1, p2);
+    /// ```
```
mmagician, line 255: It feels a bit weird to define a negative example like this one. The expanded description should suffice

poly/src/polynomial/univariate/sparse.rs
```diff
@@ -233,6 +233,26 @@ impl<F: Field> SparsePolynomial<F> {
     }
 
     /// Constructs a new polynomial from a list of coefficients.
+    /// Since the function does not automatically combine like terms, the input *has to be* in the
+    /// correct form: no multiple monomials of the same degree.
```
mmagician, line 237:
```suggestion
    /// The function does not combine like terms and so multiple monomials
    /// of the same degree are ignored.
```

poly/src/polynomial/univariate/sparse.rs
mmagician: (side note for the future: the doctests actually have to pass and no panics are allowed. The alternative would be `assert_ne`)

poly/src/polynomial/univariate/sparse.rs
msmirnov18: I have updated the description according to your suggestions.

msmirnov18: I agree that having a negative example in the docs does feel weird.

mmagician (approved):

PR #759, opened 2024-01-16
Remove assertion and add test

Pratyush:
## Description

Removed an assertion and added some testing for an edge case.

---

write a little note why.

- [ ] Wrote unit tests

### DIFF
--- a/test-templates/src/groups.rs
+++ b/test-templates/src/groups.rs
@@ -112,7 +112,7 @@ macro_rules! __test_group {
                     }
                 }
 
-                // The table size is only determined by the number of non-zero entries.
+                // num_scalars != scalars.len()
                 let mut scalars = vec![ScalarField::rand(&mut rng); 100];
                 scalars[0] = ScalarField::zero();
                 let table = BatchMulPreprocessing::new(a, scalars.len() - 1);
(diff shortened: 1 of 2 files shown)


mmagician:
This should solve the fail in https://github.com/arkworks-rs/poly-commit/pull/142, right? Or is there something else there broken?

Pratyush:
Yup it should.

test-templates/src/groups.rs
```diff
@@ -111,6 +111,18 @@ macro_rules! __test_group {
                         assert_eq!(context.mul_with_table(&bad_table, &b), None);
                     }
                 }
+
+                // The table size is only determined by the number of non-zero entries.
```
mmagician, line 115: I think the table size could be anything the user sets.
E.g. [here](https://github.com/arkworks-rs/poly-commit/blob/f6a0c13bf44d7d1776f238bc9afabeab3c5a9511/poly-commit/src/multilinear_pc/mod.rs#L58) the table size is `num_vars`, while we have `2^(num_vars+1) - 1` scalars, if I'm reading the code correctly.

Pratyush: In that case I'll rename `max_num_scalars` to just `num_scalars`.

mmagician (approved):
That is better yes

Issue #717, opened 2023-12-16
Implement modulus (%) operator between `BigInteger`


hdvanegasm:
I found that the BigInteger trait is missing the modulus operation, i.e., the possibility of computing `a % b` at least for two positive BigIntegers. To implement this, we can refer to "The Art Of Computer Programming" by Donald Knuth, Section 4.3, Algorithm D. Such algorithm implement division with remainder, and it can be directly used to compute the modulus.

PR #761, opened 2024-01-17
Implementation of modulus operator for `BigInteger`

hdvanegasm:
Closes https://github.com/arkworks-rs/algebra/issues/717.

In this PR, I implemented the modulus operator for `BigInteger`. The implementation "The Art Of Computer Programming" by Donald Knuth, Section 4.3, Algorithm D, with the help and guidance from https://github.com/rust-num/num-bigint/blob/master/src/biguint/division.rs.

It's possible that this implementation needs some improvements but we can consider it as a starting point. The tests for the implementation include two parts: one for which I use random generated numbers and the other for which I used fixed constant numbers to test the modulus.

### DIFF
--- a/ff/src/biginteger/mod.rs
+++ b/ff/src/biginteger/mod.rs
@@ -1,6 +1,9 @@
-use core::ops::{
-    BitAnd, BitAndAssign, BitOr, BitOrAssign, BitXor, BitXorAssign, Not, Shl, ShlAssign, Shr,
-    ShrAssign,
+use core::{
+    cmp::Ordering,
+    ops::{
+        BitAnd, BitAndAssign, BitOr, BitOrAssign, BitXor, BitXorAssign, Not, Rem, RemAssign, Shl,
+        ShlAssign, Shr, ShrAssign,
+    },
 };
 
 use crate::{
@@ -294,6 +297,7 @@ impl<const N: usize> BigInt<N> {
 
 impl<const N: usize> BigInteger for BigInt<N> {
     const NUM_LIMBS: usize = N;
+    const RADIX: u128 = 1 << 64;
 
     #[inline]
     fn add_with_carry(&mut self, other: &Self) -> bool {
@@ -474,6 +478,11 @@ impl<const N: usize> BigInteger for BigInt<N> {
     #[inline]
     fn is_zero(&self) -> bool {
         self.0.iter().all(|&e| e == 0)
+    }
+
+    #[inline]
+    fn is_one(&self) -> bool {
+        self.0.iter().skip(1).all(|&e| e == 0) && self.0[0] == 1
     }
 
     #[inline]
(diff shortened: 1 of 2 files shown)


ff/src/biginteger/mod.rs
```diff
@@ -1118,6 +1453,17 @@ pub trait BigInteger:
...
+    /// Returns true iff this number is one.
+    /// # Example
+    ///
+    /// ```
+    /// use ark_ff::{biginteger::BigInteger64 as B, BigInteger as _};
+    ///
+    /// let mut one = B::from(1u64);
+    /// assert!(one.is_one());
+    /// ```
+    fn is_one(&self) -> bool;
```
mmaker, line 1465: I think we can make a default implementation for this: 
```
self == Self::from(1)
```

CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## Pending
 
+- [\#761](https://github.com/arkworks-rs/algebra/pull/761) (`ark-ff`) Implementation of modulus operator for `BigInteger`.
```
Pratyush, line 5: Could you move this to the `Features` section?

ff/src/biginteger/mod.rs
```diff
@@ -870,6 +872,326 @@ impl<const N: usize> Not for BigInt<N> {
...
+    // Covers the case of N = 1, for which the quotient and remainder are
+    // computed as usual.
+    if N == 1 {
+        let q: u64 = a.0[0] / d.borrow().0[0];
+        let r: u64 = a.0[0] % d.borrow().0[0];
+        return (BigInt::<N>::from(q), BigInt::<N>::from(r));
+    }
+
+    // Covers immediate cases where a <= d.
+    match a.cmp(d.borrow()) {
```
Pratyush, line 904: I think for simplicity we can add a line that borrows `d` just once:

```suggestion
	let d = d.borrow();

    // Covers immediate cases where a <= d.
    match a.cmp(d) {
```

Pratyush (changes requested):
Thanks for the PR!

I have a couple of questions. Could you comment on why this implementation seems much longer than, the impl in [`crypto-bigint`](https://github.com/RustCrypto/crypto-bigint/blob/e925f370995da5ca519db482906aff4679656c4a/src/uint/div.rs#L38)?

Is the algorithm you implement much faster? Or is it handling too many cases (e.g., because it was adapted from `num-bigint` which handles variable length?)

hdvanegasm:
Hi, thanks for the comments and sorry for the delay.

Regarding your question. Unfortunately, I didn't know about the `crypto-bigint` until now. Comparing the implementations, the one presented here is longer given that I had to add some conditionals to setup the inputs to the algorithm to the correct format. For example, the algorithm implemented here requires that the most significant limb of the divisor is non-zero, so I had to remove the leading zero limbs. After the computation, I had to add the missing limbs back to obtain a `BigInt<N>`. Also, this algorithm requires right-shifting of the dividend, which need additional conditionals in case that there is an overflow, because those overflow bits cannot be discarded.

Comparing the performance of both algorithms, in `crypto-bigint`, they claim to have a function that is constant time with respect to the length of the inputs. However, looking at "The art of computer programming", Vol 2, Ed. 3, Section 4.3, under certain reasonable assumptions, the author claims that the running time is $O(M \cdot N)$, where $M$ is the number of limbs of quotient and $N$ is the number of limbs of the divisor, assuming that all the limbs of the `BigInt<N>` divisor are non-zero. Then, the number of limbs of the quotient is the difference between the number of limbs of the dividend and the divisor plus one. Hence, the running time would be bunded by $O(NA -N^2+N)$, where $A$ is the number of limbs of the dividend. But $A \leq N$, so the running time of dividing two `BigInt<N>` would be $O(N)$.

Certainly, the previous analysis shows that `crypto-bigint` is more efficient according to their claims.

PR #767, opened 2024-01-31
Change generator of ark-curve25519

mmaker:
In this test: https://github.com/arkworks-rs/nimue/blob/main/src/plugins/tests.rs#L43

I try to make compatible transcripts between arkworks and dalek. The reason it passes right now is that, at [line 54](https://github.com/arkworks-rs/nimue/blob/main/src/plugins/tests.rs#L54), I negate the generator. It'd be nice if they both were compatible so that people can use transcripts interchangeably between the two libraries


mmagician:
Could you add a CHANGELOG entry please, this is a breaking change

mmaker:
done 🫡

Pratyush (changes requested):

Issue #645, opened 2023-04-25
Add multiplication and division methods for `BigInteger`


mmagician:
This will simplify some code, e.g. see [here](https://github.com/arkworks-rs/algebra/pull/644#discussion_r1176874187).

PR #772, opened 2024-02-02
add `mul` method for `BigInteger`

tcoratger:
## Description

This pull request introduces a new `mul` method for the `BigInteger` trait. The `mul` method allows efficient in-place multiplication of two `BigInteger` values, providing a convenient and performant way to compute the product of two instances. 

This enhancement complements the existing set of arithmetic operations supported by the `BigInteger` trait, contributing to the versatility and usability of the library.

Should be part of #645.

---

- [X] Wrote unit tests

### DIFF
--- a/ff/src/biginteger/mod.rs
+++ b/ff/src/biginteger/mod.rs
@@ -422,7 +422,29 @@ impl<const N: usize> BigInteger for BigInt<N> {
     }
 
     #[inline]
-    fn mul(&mut self, other: &Self) {
+    fn mul(&self, other: &Self) -> (Self, Self) {
+        if self.is_zero() || other.is_zero() {
+            let zero = Self::zero();
+            return (zero, zero);
+        }
+
+        let mut r = crate::const_helpers::MulBuffer::<N>::zeroed();
+
+        let mut carry = 0;
+
+        for i in 0..N {
+            for j in 0..N {
+                r[i + j] = mac_with_carry!(r[i + j], self.0[i], other.0[j], &mut carry);
+            }
+            r.b1[i] = carry;
+            carry = 0;
+        }
+
+        return (Self(r.b0), Self(r.b1));
+    }
+
+    #[inline]
+    fn mul_low(&mut self, other: &Self) {
         if self.is_zero() || other.is_zero() {
             *self = Self::zero();
             return;
@@ -439,6 +461,11 @@ impl<const N: usize> BigInteger for BigInt<N> {
         }
 
         *self = res
+    }
+
+    #[inline]
+    fn mul_high(&self, other: &Self) -> Self {
+        self.mul(other).1
     }
 
     #[inline]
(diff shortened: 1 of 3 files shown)


ff/src/biginteger/mod.rs
```diff
@@ -421,6 +421,26 @@ impl<const N: usize> BigInteger for BigInt<N> {
...
+        for i in 0..N {
+            for j in 0..(N - i) {
+                res.0[i + j] = mac_with_carry!(res.0[i + j], self.0[i], other.0[j], &mut carry);
+            }
+            carry = 0;
+        }
+
+        *self = res
+    }
+
```
Pratyush, line 443: Let's to add a `mul_high` method that returns the high part of the result (i.e. the carry), and let's rename this method to `mul_low`, adjusting the documentation appropriately

Pratyush (changes requested):
Thanks for the PR! Left a couple of small comments

ff/src/biginteger/mod.rs
Pratyush: Oh and the `mul` method can then return both a tuple containing both the high and low parts. This is useful for some applications.

ff/src/biginteger/tests.rs
```diff
@@ -1,5 +1,6 @@
 use crate::{biginteger::BigInteger, UniformRand};
 use num_bigint::BigUint;
+// use std::println;
```
Pratyush, line 3:
```suggestion
```

ff/src/biginteger/tests.rs
```diff
@@ -49,6 +50,33 @@ fn biginteger_arithmetic_test<B: BigInteger>(a: B, b: B, zero: B) {
     let mut a_plus_a = a;
     a_plus_a.add_with_carry(&a); // Won't assert anything about carry bit.
     assert_eq!(a_mul2, a_plus_a);
+
+    // a * 1 = a
```
Pratyush, line 54: Let's also add some tests to `test-templates`, if appropriate?

ff/src/biginteger/mod.rs
tcoratger: Should be fixed now

ff/src/biginteger/tests.rs
tcoratger: For these implementations around `BigInt`, all unit tests can be found in this file:
- So I would say that for the moment there is not necessarily a need to outsource the tests to `test-templates` no?
- I added a `max` parameter to try to test with a slightly wider coverage of all the multiplication features
- If you want I can also add another function here that tests something specifically for multiplication? But it seems to me that with the tests here we are pretty well covered.
- To avoid mixing the PRs, if you want we can add a new issue to migrate all the tests that are here to `test-templates` and try to put an even broader conversion on all the functions of the implementation, I don't know if you see any interest in it?

Pratyush (approved):
Thanks, this looks great!

ff/src/biginteger/mod.rs
```diff
@@ -421,6 +421,53 @@ impl<const N: usize> BigInteger for BigInt<N> {
...
+        for i in 0..N {
+            for j in 0..(N - i) {
+                res.0[i + j] = mac_with_carry!(res.0[i + j], self.0[i], other.0[j], &mut carry);
+            }
+            carry = 0;
+        }
+
+        *self = res
+    }
```
Pratyush, line 464:
```suggestion
    #[inline]
    fn mul_low(&self, other: &Self) -> Self {
        if self.is_zero() || other.is_zero() {
            *self = Self::zero();
            return;
        }

        let mut res = Self::zero();
        let mut carry = 0;

        for i in 0..N {
            for j in 0..(N - i) {
                res.0[i + j] = mac_with_carry!(res.0[i + j], self.0[i], other.0[j], &mut carry);
            }
            carry = 0;
        }

        res
    }
```

Pratyush: (For consistency, the operations should all return values. If we want in place operations, we should mark those explicitly)

tcoratger: Sure no problem, I've just followed the convention of other operations here like `mul2` or `muln`.

Just did a follow up PR here: https://github.com/arkworks-rs/algebra/pull/775

PR #775, opened 2024-02-07
return value for `mul_low` instead of in place

tcoratger:
Related https://github.com/arkworks-rs/algebra/pull/772#discussion_r1480855367


Pratyush (approved):

Pratyush:
Thank you for the PR!

PR #779, opened 2024-02-13
Allow multiplication of extension field elements directly by the `BasePrimeField` element

mmagician:
## Description

Note, there was already a method to multiply by the extension field's `BaseField`: but for a field tower, a `BaseField` might not necessarily be the `BasePrimeField`, but rather an extension field itself.

---

write a little note why.

- [x] Wrote unit tests

### DIFF
--- a/ff/src/fields/mod.rs
+++ b/ff/src/fields/mod.rs
@@ -349,7 +349,7 @@ pub trait Field:
         Some(res)
     }
 
-    fn mul_by_base_field_elem(&self, elem: &Self::BasePrimeField) -> Self;
+    fn mul_by_base_prime_field(&self, elem: &Self::BasePrimeField) -> Self;
 }
 
 // Given a vector of field elements {v_i}, compute the vector {v_i^(-1)}
--- a/ff/src/fields/models/cubic_extension.rs
+++ b/ff/src/fields/models/cubic_extension.rs
@@ -341,11 +341,11 @@ impl<P: CubicExtConfig> Field for CubicExtField<P> {
         P::mul_base_field_by_frob_coeff(&mut self.c1, &mut self.c2, power);
     }
 
-    fn mul_by_base_field_elem(&self, elem: &Self::BasePrimeField) -> Self {
+    fn mul_by_base_prime_field(&self, elem: &Self::BasePrimeField) -> Self {
         let mut result = *self;
-        result.c0 = result.c0.mul_by_base_field_elem(elem);
-        result.c1 = result.c1.mul_by_base_field_elem(elem);
-        result.c2 = result.c2.mul_by_base_field_elem(elem);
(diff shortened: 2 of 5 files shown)


mmagician:
@suyash67

ff/src/fields/mod.rs
```diff
@@ -350,6 +350,8 @@ pub trait Field:
         }
         Some(res)
     }
+
+    fn mul_by_base_field_elem(&self, elem: &Self::BasePrimeField) -> Self;
```
Pratyush, line 354:
```suggestion
    fn mul_by_base_prime_field(&self, elem: &Self::BasePrimeField) -> Self;
```

Also, should we move this into a `Mul<Self::BasePrimeField, Output = Self>` bound? (And maybe add `for<'a> Mul<&'a Self::BasePrimeField, Output = Self>`, `MulAssign<Self::BasePrimeField>`, and `for<'a> MulAssign<&'a Self::BasePrimeField>` bounds too?

Pratyush:
Where is the existing multiplication by base field?

ff/src/fields/mod.rs
mmagician: I tried this initially, but we already have `Mul<Self>` coming from a [macro](https://github.com/arkworks-rs/algebra/blob/3835af548d929d08f3ff1771c4349ed49c698efd/ff/src/fields/arithmetic.rs#L171), and adding a trait bound as you suggest, and then implementing:
```

impl<P: QuadExtConfig> Mul<<Self as Field>::BasePrimeField> for QuadExtField<P> {
    type Output = Self;
    ...
}
```
causes a `conflicting implementation` error.
I believe this is what you meant, right?

mmagician:
@Pratyush We have a method directly on `QuadExtField`: https://github.com/arkworks-rs/algebra/blob/9cb911f02077ab3701352b1f6a6807b19bd4ff8d/ff/src/fields/models/quadratic_extension.rs#L170

mmagician:
Actually, we also have the counterpart for `CubicExtField`, although the names weren't consistent, notice the underscore in `base_field`: https://github.com/arkworks-rs/algebra/blob/23693477d44e674ba0e386e2359da887633bca78/ff/src/fields/models/cubic_extension.rs#L122

I would propose to unify these two as part of this PR, at least when it comes to the name. We could pull that method out into a separate trait, but I'm not sure it's worth it (it is not used anywhere in the arkworks org).

ff/src/fields/mod.rs
Pratyush: Ouch that's annoying...

I guess Rust doesn't know that for `QuadExtField`, `BasePrimeField` is always different from `Self`...

But it should be able to figure that out: `BasePrimeField: PrimeField`, while `Self` doesn't implement that.

Pratyush (approved):

Issue #781, opened 2024-02-15
Use `in_place` suffix for `BigInteger` operations


tcoratger:
In the whole codebase, there is a standard which is to include the suffix `in_place` on operations which modify the value passed as a mutable parameter. For example here:

https://github.com/arkworks-rs/algebra/blob/cc2ad8c368c323ee9d9aacaabf8e353de95682e6/ec/src/models/bn/g2.rs#L53-L77

We know from the function name that the `self` mutable will be modified.

However similar operations are performed in the `BigInteger` trait like:

https://github.com/arkworks-rs/algebra/blob/cc2ad8c368c323ee9d9aacaabf8e353de95682e6/ff/src/biginteger/mod.rs#L1062

or 

https://github.com/arkworks-rs/algebra/blob/cc2ad8c368c323ee9d9aacaabf8e353de95682e6/ff/src/biginteger/mod.rs#L1037

without including the `in_place` suffix. For readability and consistency, it could be nice to include this suffix for all the `BigInteger` operations concerned by this type of logic.

Pratyush:
Yes, this would be great!

PR #782, opened 2024-02-16
Add `in_place` suffix for `BigInteger` methods

tcoratger:
## Description

This pull request introduces a standardized naming convention for methods in the `BigInteger` module, specifically targeting functions where the `self` mutable value is modified in-place. The convention adds the `in_place` suffix to method names to provide clarity and consistency in the codebase.

closes: #781 

---

write a little note why.

- [ ] Wrote unit tests

### DIFF
--- a/ec/src/hashing/curve_maps/swu.rs
+++ b/ec/src/hashing/curve_maps/swu.rs
@@ -73,9 +73,9 @@ impl<P: SWUConfig> MapToCurve<Projective<P>> for SWUMap<P> {
         let div = a * if ta.is_zero() { P::ZETA } else { -ta };
 
         let num2_x1 = num_x1.square();
-        let div2 = div.square();
-        let div3 = div2 * div;
-        let num_gx1 = (num2_x1 + a * div2) * num_x1 + b * div3;
+        let div2_in_place = div.square();
+        let div3 = div2_in_place * div;
+        let num_gx1 = (num2_x1 + a * div2_in_place) * num_x1 + b * div3;
 
         // 5. x2 = Z * u^2 * x1
         let num_x2 = zeta_u2 * num_x1; // same div
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,9 +32,9 @@
 
 ### Improvements
 
-- [\#736](https://github.com/arkworks-rs/algebra/pull/736) (`ark-ff`) Deprecate `divn()`, and use `core::ops::{Shr, ShrAssign}` instead.
-- [\#739](https://github.com/arkworks-rs/algebra/pull/739) (`ark-ff`) Deprecate `muln()`, and use `core::ops::{Shl, ShlAssign}` instead.
-- [\#771](https://github.com/arkworks-rs/algebra/pull/771) (`ark-ec`) Omit expensive  scalar multiplication in `is_in_correct_subgroup_assuming_on_curve()` for short Weierstrass curves of cofactor one.  
+- [\#736](https://github.com/arkworks-rs/algebra/pull/736) (`ark-ff`) Deprecate `divn_in_place()`, and use `core::ops::{Shr, ShrAssign}` instead.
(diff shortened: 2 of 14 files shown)


ec/src/hashing/curve_maps/swu.rs
```diff
@@ -73,9 +73,9 @@ impl<P: SWUConfig> MapToCurve<Projective<P>> for SWUMap<P> {
         let div = a * if ta.is_zero() { P::ZETA } else { -ta };
 
         let num2_x1 = num_x1.square();
-        let div2 = div.square();
-        let div3 = div2 * div;
-        let num_gx1 = (num2_x1 + a * div2) * num_x1 + b * div3;
+        let div2_in_place = div.square();
+        let div3 = div2_in_place * div;
+        let num_gx1 = (num2_x1 + a * div2_in_place) * num_x1 + b * div3;
```
Pratyush, line 78:
```suggestion
        let div2 = div.square();
        let div3 = div2 * div;
        let num_gx1 = (num2_x1 + a * div2) * num_x1 + b * div3;
```

Pratyush (approved):

tcoratger:
@Pratyush I thought about this PR and because the `MulAssign` or `AddAssign` traits of the Rust library use the suffix `Assign`, we will necessarily have `mul_assign` or `add_assign` which will be present somewhere in the code. 

To unify everything, wouldn't it be better to rename everything throughout the code with an `assign` suffix rather than `in_place`. And for example the instances of `square_in_place` would also become `square_assign`, what do you think?

mmaker:
I'd much rather have a `_assign` indeed. What do people think?

Pratyush:
Wouldn’t it be a massive breaking change?

mmaker:
Not more than adding the`in_place` suffix?  If it helps with naming consistency down the line before 1.0 I think that's a valid effort (and we should deprecate the previous names, not just remove them)

Pratyush:
Ah I assumed it was to do it for all of the other traits too, like Field and GroupOn Feb 28, 2024, at 8:51 AM, Michele Orrù ***@***.***> wrote:﻿
Not more than adding thein_place suffix?  If it helps with naming consistency down the line before 1.0 I think that's a valid effort (and we should deprecate the previous names, not just remove them)

mmaker:
[EDIT]To phrase better my comment: 

In `Field` we have
```
double_in_place (from AdditiveGroup)
neg_in_place (from AdditiveGroup)
sqrt_in_place
square_in_place
inverse_in_place
frobenius_map_in_place
```

I think it's nice to have either (1) `in_place` everywhere except trait implementations of `std::ops::*`, OR (2) `_assign` everywhere. I'd be consistent with whatever we pick, and propagate this change to ark-ff and ark-ec, deprecating old function calls that will be removed for 1.0.
I fear having intermediate solutions will make the naming conventions of the library confusing. 

@Pratyush In your comment, are you implying that having `_assign` only for `BigInt` is acceptable?
@tcoratger What do you think?

tcoratger:
@mmaker Agreed, I found it confusing at the beginning like in Montgomery backend for exemple we have:
- `add_assign`
- `sub_assign`
- `double_in_place`
- `neg_in_place`

The `std` library always uses the `_assign` suffix for all operations so for me it would be worth changing everything to have consistent naming everywhere (`_assign` in a `std` like style) in the codebase. But would agree that this is quite a massive change of a big part of the function names of the whole library.

burdges:
"Assignment operators" is the C term, so `_assign` made sense for operators, and hence maybe sensible for methods resembling operator methods, ala `mul2_assign`.

Afaik `_in_place` was always more traditional overall, and [std uses in_place for pointers](https://doc.rust-lang.org/std/ops/struct.Range.html?search=in_place), so nothing wrong with `_in_place` either, but maybe resembling operators should be the priority. 

Also, big integers are bigish for high performace code, so if a non-in-place variant feels particularly rediculous for operation `foo`, then `foo` could be the in-place variant, and `_cloned` could be the non-in-place variant, but again maybe resembling operators should be the priority.

Pratyush:
So my rough rule-of-thumb has been to use `_assign` for the binary operators, and `_in_place` for the unary ones (`negate_in_place`, `square_in_place`, etc). Happy to uniformize the notation to avoid confusion though.

PR #783, opened 2024-02-17
Small fix in `add_assign` for Projective

tcoratger:
Some fix in the logic of `add_assign` for projective coordinates to replace
```rust
if u1 == u2 && s1 == s2 {
            if s1 == s2 {
```
by
```rust
if u1 == u2 {
            if s1 == s2 {
```


Pratyush:
Thanks For the PR! We would need benchmarks here because this is pretty perf sensitive code

tcoratger:
> ```rust
> if u1 == u2 {
>             if s1 == s2 {
> ```

@Pratyush Yep, agreed, maybe let us push this change:

```rust
if u1 == u2 {
            if s1 == s2 {
 ```

now because this is a fix, and can keep the other for another PR with benchmark, what do you think?

tcoratger:
@Pratyush Done, just revert all except the fix, will open a PR when having time with some bench for the rest.

Pratyush (approved):

mmagician (commented):
Hmm doesn't this actually change the logic?
Specifically, the case `u1 == u2 && s1 != s2` enters the outer `else` in the current version, while with this PR it would enter the inner `else`.

tcoratger:
> Hmm doesn't this actually change the logic? Specifically, the case `u1 == u2 && s1 != s2` enters the outer `else` in the current version, while with this PR it would enter the inner `else`.

@mmagician Yes this is the goal because we changed the logic here: https://github.com/arkworks-rs/algebra/pull/780

The principle behind this is that if `u1 == u2` then:
- If `s1 == s2` we must make a doubling point.
- Otherwise `u1 == u2` and `s1 != s2` it is the case `a + (-a) = 0` having to return the identity without needing to carry out the rest of the logic in the else global.

However in the previous PR I forgot to remove the `&& s1 == s2` in the upper `if`. This PR corrects that.

mmagician (approved):
Ah yes I missed that. Thanks for pointing it out

Pratyush (approved):

PR #788, opened 2024-02-23
Add a few simplifications around Mont inverse

tcoratger:
## Description

- When using `div2` method of `BigInt`, we traverse the members in the opposite direction (big endian order) with a loop which until now was carried out in the little endian order. Therefore, in each iteration of the loop, the calculation of `N - i - 1` was necessary. If we go through the limbs in the opposite direction with a while as I propose here, the calculation is no longer carried out on each pass, because we take the member `i` directly.
- In the `inverse` method of the Montgomery back end, when we exit the following piece of code:
```rust
if a.is_zero() {
  None
}
```
there is no need for the else, because the rest will necessarily take place if `a` is not zero. Therefore, I removed the else here to avoid unnecessary indentations.

### DIFF
--- a/ff/src/biginteger/mod.rs
+++ b/ff/src/biginteger/mod.rs
@@ -470,8 +470,10 @@ impl<const N: usize> BigInteger for BigInt<N> {
     #[inline]
     fn div2(&mut self) {
         let mut t = 0;
-        for i in 0..N {
-            let a = &mut self.0[N - i - 1];
+        let mut i = N;
+        while i > 0 {
+            i -= 1;
+            let a = &mut self.0[i];
             let t2 = *a << 63;
             *a >>= 1;
             *a |= t;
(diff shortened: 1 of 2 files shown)


Pratyush (approved):

ff/src/biginteger/mod.rs
```diff
@@ -470,8 +470,10 @@ impl<const N: usize> BigInteger for BigInt<N> {
     #[inline]
     fn div2(&mut self) {
         let mut t = 0;
-        for i in 0..N {
-            let a = &mut self.0[N - i - 1];
+        let mut i = N;
+        while i > 0 {
+            i -= 1;
+            let a = &mut self.0[i];
```
Pratyush, line 476:
```suggestion
        for a in self.0.iter_mut().rev() {
```

PR #790, opened 2024-02-24
Clean up all imports

tcoratger:
Clean imports related to last Rust nightly version.

### DIFF
--- a/ff-macros/src/montgomery/mod.rs
+++ b/ff-macros/src/montgomery/mod.rs
@@ -1,4 +1,3 @@
-use quote::format_ident;
 use std::str::FromStr;
 
 use num_bigint::BigUint;


Pratyush:
I feel like this is some weird regression in rust nightly... it's not at all clear to me why this should error

curves/mnt4_753/src/fields/tests.rs
```diff
@@ -1,7 +1,7 @@
...
 use ark_algebra_test_templates::*;
 
-test_field!(100; fr; Fr; mont_prime_field);
-test_field!(100; fq; Fq; mont_prime_field);
-test_field!(100; fq2; Fq2);
-test_field!(100; fq4; Fq4);
+test_field!(fr; Fr; mont_prime_field);
+test_field!(fq; Fq; mont_prime_field);
+test_field!(fq2; Fq2);
+test_field!(fq4; Fq4);
```
Pratyush, line 7: Ditto for only 100

curves/mnt6_753/src/curves/tests.rs
```diff
@@ -1,7 +1,7 @@
 use crate::*;
 use ark_algebra_test_templates::*;
 
-test_group!(50; g1; G1Projective; sw);
-test_group!(50; g2; G2Projective; sw);
-test_group!(50; pairing_output; ark_ec::pairing::PairingOutput<MNT6_753>; msm);
+test_group!(g1; G1Projective; sw);
+test_group!(g2; G2Projective; sw);
+test_group!(pairing_output; ark_ec::pairing::PairingOutput<MNT6_753>; msm);
```
Pratyush, line 6: Same.

curves/mnt6_753/src/fields/tests.rs
```diff
@@ -3,10 +3,10 @@ use ark_algebra_test_templates::*;
...
 use ark_std::{rand::Rng, test_rng};
 
-test_field!(100; fr; Fr; mont_prime_field);
-test_field!(100; fq; Fq; mont_prime_field);
-test_field!(100; fq3; Fq3);
-test_field!(100; fq6; Fq6);
+test_field!( fr; Fr; mont_prime_field);
+test_field!( fq; Fq; mont_prime_field);
+test_field!( fq3; Fq3);
+test_field!( fq6; Fq6);
```
Pratyush, line 9: Same

curves/mnt6_753/src/curves/tests.rs
tcoratger: Fixed

curves/mnt4_753/src/fields/tests.rs
tcoratger: Fixed

curves/mnt6_753/src/fields/tests.rs
tcoratger: Fixed

curves/mnt4_753/src/curves/tests.rs
```diff
@@ -1,7 +1,7 @@
 use crate::*;
 use ark_algebra_test_templates::*;
 
-test_group!(50; g1; G1Projective; sw);
-test_group!(50; g2; G2Projective; sw);
-test_group!(50; pairing_output; ark_ec::pairing::PairingOutput<MNT4_753>; msm);
+test_group!(g1; G1Projective; sw);
+test_group!(g2; G2Projective; sw);
+test_group!(pairing_output; ark_ec::pairing::PairingOutput<MNT4_753>; msm);
```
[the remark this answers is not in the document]
tcoratger: Fixed

tcoratger:
> I feel like this is some weird regression in rust nightly... it's not at all clear to me why this should error

@Pratyush These are imports. There were two specific cases that Rust nightly now detects:
- Cases of overlaying of imports which seemed not to be detected before and which are now: for example if an element was already "under-imported" in another import and it is imported again manually.
- There is a second case which corresponds to the std prelude to which new components have just been added (https://doc.rust-lang.org/std/prelude/index.html) such as `Vec` for example . As arkworks is mostly `no_std`, when we explicitly imported `ark_std::vec::Vec`, then Rust saw it as an import overlay since `Vec` exists in the std prelude. But if we do not do this import with the `no_std` flag, then Rust does not recognize `Vec` because it is unable to use the std prelude. So the only way in my opinion is to import `ark_std::vec::*` to import `Vec` without really saying it.

Pratyush:
Merged #798 into this to allow it to merge.

PR #792, opened 2024-02-26
A few adjustments in big integer `shr_assign`

tcoratger:
- If the number of bits shifted is larger than `N * 64`, the function can be exited at the beginning without having to do the rest of the calculations.
- As the limbs are traversed from the end to the beginning, it is generally less expensive to do a while starting from the end to avoid looking for an index at the opposite end of the table at each iteration.

### DIFF
--- a/ff/src/biginteger/mod.rs
+++ b/ff/src/biginteger/mod.rs
@@ -821,10 +821,8 @@ impl<const N: usize> ShrAssign<u32> for BigInt<N> {
 
         while rhs >= 64 {
             let mut t = 0;
-            let mut limb = N;
-            while limb > 0 {
-                limb -= 1;
-                core::mem::swap(&mut t, &mut self.0[limb]);
+            for limb in self.0.iter_mut().rev() {
+                core::mem::swap(&mut t, limb);
             }
             rhs -= 64;
         }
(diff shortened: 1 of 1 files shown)


ff/src/biginteger/mod.rs
```diff
@@ -816,20 +816,25 @@ impl<const N: usize> ShrAssign<u32> for BigInt<N> {
...
 
         while rhs >= 64 {
             let mut t = 0;
-            for i in 0..N {
-                core::mem::swap(&mut t, &mut self.0[N - i - 1]);
+            let mut limb = N;
+            while limb > 0 {
+                limb -= 1;
+                core::mem::swap(&mut t, &mut self.0[limb]);
             }
```
Pratyush, line 828:
```suggestion
            for limb in self.limbs.iter_mut().rev() {
                core::mem::swap(&mut t, limb);
            }
```

ff/src/biginteger/mod.rs
```diff
@@ -816,20 +816,25 @@ impl<const N: usize> ShrAssign<u32> for BigInt<N> {
...
         }
 
         if rhs > 0 {
             let mut t = 0;
-            for i in 0..N {
-                let a = &mut self.0[N - i - 1];
+            let mut limb = N;
+            while limb > 0 {
+                limb -= 1;
+                let a = &mut self.0[limb];
```
Pratyush, line 837:
```suggestion
            for a in self.0.iter_mut().rev() {
```

PR #793, opened 2024-02-27
Remove useless variable decl in `into_bigint` Mont

tcoratger:
In this pull request, it is proposed to streamline the code by eliminating the unnecessary `tmp` variable. 

Currently, the `tmp` variable serves the sole purpose of holding the `[u64; N]` array temporarily to store the final `BigInt` returned at the end of the function. 

However, this approach can be made more readable and concise. Therefore, I suggest removing the `tmp` variable and replacing it with a direct declaration of `BigInt` at the function's conclusion. This modification enhances code clarity and removes redundant assignments.

### DIFF
--- a/ff/src/fields/models/fp/montgomery_backend.rs
+++ b/ff/src/fields/models/fp/montgomery_backend.rs
@@ -371,7 +371,7 @@ pub trait MontConfig<const N: usize>: 'static + Sync + Send + Sized {
     #[unroll_for_loops(12)]
     #[allow(clippy::modulo_one)]
     fn into_bigint(a: Fp<MontBackend<Self, N>, N>) -> BigInt<N> {
-        let mut r = a.0 .0;
+        let mut r = (a.0).0;
         // Montgomery Reduction
         for i in 0..N {
             let k = r[i].wrapping_mul(Self::INV);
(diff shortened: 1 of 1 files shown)


ff/src/fields/models/fp/montgomery_backend.rs
```diff
@@ -371,8 +371,7 @@ pub trait MontConfig<const N: usize>: 'static + Sync + Send + Sized {
     #[unroll_for_loops(12)]
     #[allow(clippy::modulo_one)]
     fn into_bigint(a: Fp<MontBackend<Self, N>, N>) -> BigInt<N> {
-        let mut tmp = a.0;
-        let mut r = tmp.0;
+        let mut r = a.0 .0;
```
mmaker, line 374: why the whitespace?

Pratyush: the whitespace is there for the same reason we had the variable: rust thinks `a.0.0` is actually `a.(0.0)`, and parse the latter part as a float

Pratyush (approved):

ff/src/fields/models/fp/montgomery_backend.rs
mmaker: gosh

tcoratger: I think it's rust fmt stuff. I found it strange at the beginning but I found a lots of related issues in [rustfmt](https://github.com/rust-lang/rustfmt) repo. Like this one: https://github.com/rust-lang/rustfmt/issues/5745

mmaker: this seems to work on play.rustlang tho
```

struct B (pub u8, pub u8);
struct A (pub B, pub B);

fn main() { 
    let a = A(B(0, 1), B(2, 3));
    println!("{}", a.0.0)
}
```

PR #796, opened 2024-02-27
Simplify `from_bits_be`in big int by reusing `from_bits_le` inside

tcoratger:
### Description
The current implementation of the `from_bits_be` function in the BigInt module involves reversing the input `bits` array and then doing some operations. This can be simplified by directly using `from_bits_le` with a reversed copy of the `bits` array, achieving the same functionality in a more concise and readable manner.

### DIFF
--- a/ff/src/biginteger/mod.rs
+++ b/ff/src/biginteger/mod.rs
@@ -549,20 +549,9 @@ impl<const N: usize> BigInteger for BigInt<N> {
 
     #[inline]
     fn from_bits_be(bits: &[bool]) -> Self {
-        let mut res = Self::default();
-        let mut acc: u64 = 0;
-
         let mut bits = bits.to_vec();
         bits.reverse();
-        for (i, bits64) in bits.chunks(64).enumerate() {
-            for bit in bits64.iter().rev() {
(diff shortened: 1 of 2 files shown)


mmaker (approved):

ff/src/biginteger/mod.rs
```diff
@@ -549,20 +549,9 @@ impl<const N: usize> BigInteger for BigInt<N> {
...
-        for (i, bits64) in bits.chunks(64).enumerate() {
-            for bit in bits64.iter().rev() {
-                acc <<= 1;
-                acc += *bit as u64;
-            }
-            res.0[i] = acc;
-            acc = 0;
-        }
-        res
+        Self::from_bits_le(&bits)
```
Pratyush, line 554: Could you please add a test to confirm that the logic is the same?

tcoratger: Done, just added two unit tests in the suite in order to confirm the logic in both directions.

Pratyush: Great, thanks!

Pratyush (approved):

PR #804, opened 2024-03-08
SW  <-> TE map for Bandersnatch

drskalman:
## Description

Implement the birational map between the Short Weierstrass and Twisted Edwards forms of the Bandersnatch curve.

It is important because it is significantly cheaper to verifiy SNARK using the incomplete addition formula of SW form while Elligator hash-to-curve is significantly cheaper than WB or SWU hash-to-curve.

Specifically for Bandersnatch serializing in TE form saves a byte from 33 to 32 which is more machine friendly.
---

- [X] Wrote unit tests
- [X] Added a relevant changelog entry to the `Pending` section in `CHANGELOG.md`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 - [\#772](https://github.com/arkworks-rs/algebra/pull/772) (`ark-ff`) Implementation of `mul` method for `BigInteger`.
 - [\#794](https://github.com/arkworks-rs/algebra/pull/794) (`ark-ff`) Fix `wasm` compilation.
+- [\#XXX](https://github.com/arkworks-rs/algebra/pull/XXX) (`ark-curves`) Implement TE <-> SW map for  Bandersnatch.
 
 ### Breaking changes
 
--- a/curves/ed_on_bls12_381_bandersnatch/src/curves/mod.rs
+++ b/curves/ed_on_bls12_381_bandersnatch/src/curves/mod.rs
@@ -1,12 +1,11 @@
+use crate::{Fq, Fr};
 use ark_ec::{
     hashing::curve_maps::elligator2::Elligator2Config,
     models::CurveConfig,
     short_weierstrass::{self, SWCurveConfig},
     twisted_edwards::{Affine, MontCurveConfig, Projective, TECurveConfig},
 };
-use ark_ff::{AdditiveGroup, MontFp};
-
-use crate::{Fq, Fr};
+use ark_ff::{AdditiveGroup, Field, MontFp, One};
 
 #[cfg(test)]
 mod tests;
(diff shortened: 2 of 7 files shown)


CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 - [\#772](https://github.com/arkworks-rs/algebra/pull/772) (`ark-ff`) Implementation of `mul` method for `BigInteger`.
 - [\#794](https://github.com/arkworks-rs/algebra/pull/794) (`ark-ff`) Fix `wasm` compilation.
+- [\#XXX](https://github.com/arkworks-rs/algebra/pull/XXX) (`ark-curves`) Implement TE <-> SW map for  Bandersnatch.
```
weikengchen, line 7: (since the PR is from an organization, I cannot edit directly)

```suggestion
- [\#XXX](https://github.com/arkworks-rs/algebra/pull/XXX) (`ark-curves`) Implement TE <-> SW map for  Bandersnatch.
+ [\#XXX](https://github.com/arkworks-rs/algebra/pull/XXX) (`ark-curves`) Implement TE <-> SW map for Bandersnatch.
```

burdges:
I'd think one or more flavors of this map should typically exist whenever the both the TE and SW forms exist, although some exceptions exist.  Among other reasons, serialization should be in Edwards, even if the circuts use SW, if only because the serialized form is one byte shorter, but often because advanced compression techniques are only researched for TE.

In particular, Bandersnatch should be serialized as ganderwagon, which reuqires passing through TE.  I suppose decaf377 would similarlly be the preferable serialization for BLS12-377.

As a rule, you'll have worse interfaces if you rush into trait design than if you let the requirement emerge organically, with the hash-to-curve shit show being one example.  We do not know if multiple isogenies were alredy deployed for bandersnatch, bls12-377, or ed25519.

mmagician:
I agree with @weikengchen here.
Also, contrary to what @burdges says, having an extra trait (say `SW_TE_Mapping`) doesn't mean the interface will necessarily get bloated. Instead, people can then choose to implement the trait for whatever curves they want.

burdges:
It's quite general if you do roughly:
```
pub trait SW2TE {
    type SW: SWCurveConfig;
    type TE: TECurveConfig;
    ...
}
```
You can even still do `impl SW2TE for Config { .. }` in bls12_377 there, but doing this does not forbid other mappings since the curves are given by associated types.

I'm warning against assuming that particular impl exists, aka ignoring the associated types.  I'm especially warning against doing this:
```
pub trait SW2TE: SWCurveConfig + TECurveConfig {
    ...
}
```

curves/ed_on_bls12_381_bandersnatch/src/curves/mod.rs
```diff
@@ -163,6 +162,77 @@ impl Elligator2Config for BandersnatchConfig {
...
+        debug_assert!(
+            point_on_sw_curve.is_on_curve(),
+            "TE point mapped off the SW curve"
+        );
+
+        return Some(point_on_sw_curve);
+    }
+
+    pub fn map_sw_to_te(point: SWAffine) -> Option<EdwardsAffine> {
+        //first  we are mappyng the sw point to  montgamory point
```
dvdplm, line 212:
```suggestion
        //first we are mappng the sw point to its montgomery point
```

curves/ed_on_bls12_381_bandersnatch/src/curves/mod.rs
```diff
@@ -163,6 +162,77 @@ impl Elligator2Config for BandersnatchConfig {
...
+        let w_denom = x_p_1.inverse()?;
+
+        let v = mx * v_denom;
+        let w = (mx - <<BandersnatchConfig as CurveConfig>::BaseField as One>::one()) * w_denom;
+
+        let point_on_te_curve = EdwardsAffine::new_unchecked(v, w);
+        debug_assert!(
+            point_on_te_curve.is_on_curve(),
+            "TE point mapped off the SW curve"
+        );
```
dvdplm, line 230: Same comment as above: consider using `EdwartdsAffine::new` instead and ditch the debug assert.

curves/ed_on_bls12_381_bandersnatch/src/curves/mod.rs
```diff
@@ -163,6 +162,77 @@ impl Elligator2Config for BandersnatchConfig {
...
+// 41180284393978236561320365279764246793818536543197771097409483252169927600582
+const MONT_A_OVER_THREE: Fq =
+    MontFp!("9992940898322946442093665462003920523391277922024982836398934612730118446984");
+const MONT_B_INV: Fq =
+    MontFp!("41180284393978236561320365279764246793818536543197771097409483252169927600582");
+
+impl BandersnatchConfig {
+    /// Map an point in TE form to its corresponding  point on SW curve
+    pub fn map_te_to_sw(point: EdwardsAffine) -> Option<SWAffine> {
+        //First we map the point from the TE to Montgamory
```
davxy, line 183:
```suggestion
        if point.is_zero() {
            return SWAffine::identity()
        }
        //First we map the point from the TE to Montgamory
```

curves/ed_on_bls12_381_bandersnatch/src/curves/mod.rs
davxy, line 212:
```suggestion
        if point.is_zero() {
            return EdwardsAffine::zero()
        }
        // First  we are mappyng the sw point to  montgamory point
```

moven0831:
Hi, I'm working on the SW <-> TE map for BN254 and came across this PR. Nice work! One question: is it possible to apply the same techniquq "Short Weierstrass <-> Montgomery <-> Twisted Edwards" for BN254 curve? Thank you.

curves/ed_on_bls12_381_bandersnatch/src/curves/mod.rs
```diff
@@ -169,6 +168,83 @@ impl Elligator2Config for BandersnatchConfig {
...
+        );
+
+        return Some(point_on_sw_curve);
+    }
+
+    pub fn map_sw_to_te(point: SWAffine) -> Option<EdwardsAffine> {
+        if point.is_zero() {
+            return EdwardsAffine::zero()
+        }
+        // First  we are mappyng the sw point to  montgamory point
```
Pratyush, line 224:
```suggestion
        // First we map the SW point to the Montgomery point
```

Pratyush (commented):
Sorry for the delay here, but what is the status of this PR?

PR #808, opened 2024-03-22
Add faster subgroup checks for BN254

jonas-lj:
## Description

Add faster subgroup checks for BN254. For G1, the check is trivial because G1 is the entire curve, and for G2 we add an implementation based on an algorithm from [Co-factor clearing and subgroup membership
testing on pairing-friendly curves](https://eprint.iacr.org/2022/352.pdf) by Housni et al. For G2, the check is ~2x faster than a naive check (multiplication by 6x^2 ~ 127 bits instead of the 254 bit characteristics). See also the explanation in https://hackmd.io/@jpw/bn254#Subgroup-check-for-mathbb-G_2.

---

write a little note why.

- [X] Wrote unit tests


Pratyush (approved):
Looks good to me! Thank you for the PR!

PR #824, opened 2024-05-01
Move Hashbrown to be optional

### LINKED ISSUE #812: Ark-ec pulls in full hashbrown without feature gate
That eventually pulls things like `once_cell` that require atomics on some platforms (compare-exchange), and in general HashMap is quite heavy primitive that is still not in `alloc` of Rust itself

Full hashbrown is pulled in
- https://github.com/arkworks-rs/algebra/blob/bb663bc27e776af81e66c94d9016bde2dbba9968/ec/src/scalar_mul/variable_base/stream_pippenger.rs#L5 - for peppinger - can be solved by feature gate. Independently, do you think unordered map wins here performance-wise vs ordered (a-la BTree)?
- https://github.com/arkworks-rs/algebra/blob/bb663bc27e776af81e66c94d9016bde2dbba9968/ec/src/hashing/curve_maps/swu.rs#L154 - in test, dev-dependencies will just work
- https://github.com/arkworks-rs/algebra/blob/bb663bc27e776af81e66c94d9016bde2dbba9968/ec/src/hashing/curve_maps/elligator2.rs#L161 - same, tests only

shamatar:
## Description

Moves Hashbrown as optional dependency to allow compiling on systems without atomics. Long story short - unfortunately just making Hashbrown's `ahash` optional on the workspace will not solve everything as it removes `HashMap::new()` too, that is used in few places. Here is a variant that moves multivariate commits under feature flag, that were the only part that depends on the Hashbrown. At the moment benchmarks in `poly` part are broken as they depend on multivariate part. I believe it's possible to fix one way or another, just need to know which one.

Another option would be to try to make hash use explicit (and not use `ahash`), will be posted in alternative PR

Fixes #812 

- [ ] Wrote unit tests

### DIFF
--- a/ec/Cargo.toml
+++ b/ec/Cargo.toml
@@ -26,8 +26,8 @@ num-traits.workspace = true
 num-integer.workspace = true
 rayon = { workspace = true, optional = true }
 zeroize = { workspace = true, features = ["zeroize_derive"] }
-hashbrown.workspace = true
 itertools.workspace = true
+hashbrown = { workspace = true, optional = true }
 
 [dev-dependencies]
 ark-test-curves = { workspace = true, features = ["bls12_381_curve"] }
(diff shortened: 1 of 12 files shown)


mmagician:
closed by the alternative solution in https://github.com/arkworks-rs/algebra/pull/825

PR #825, opened 2024-05-01
Replace Ahash with Fnv for Hashbrown

### LINKED ISSUE #812: Ark-ec pulls in full hashbrown without feature gate

shamatar:
Alternative to #824 , will fix #812 too for systems without atomics. Fnv provides no key randomization, but it shouldn't be required for commitment scheme.

- [ ] Wrote unit tests

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -55,8 +55,7 @@ num-integer = { version = "0.1", default-features = false }
 criterion = "0.5.0"
 derivative = "2"
 digest = { version = "0.10", default-features = false }
-hashbrown = { version = "0.14", default-features = false }
-fnv = { version = "1.0", default-features = false }
+hashbrown = { version = "0.14", default-features = false, features = ["inline-more", "allocator-api2"] }
 hex = "0.4"
 itertools = { version = "0.12", default-features = false }
 libtest-mimic = "0.7.0"
@@ -69,7 +68,6 @@ sha2 = { version = "0.10", default-features = false }
 sha3 = { version = "0.10", default-features = false }
 blake2 = { version = "0.10", default-features = false }
 zeroize = { version = "1", default-features = false }
-
 
 proc-macro2 = "1.0"
 quote = "1.0"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -55,7 +55,8 @@ num-integer = { version = "0.1", default-features = false }
 criterion = "0.5.0"
 derivative = "2"
 digest = { version = "0.10", default-features = false }
-hashbrown = "0.14"
+hashbrown = { version = "0.14", default-features = false }
+fnv = { version = "1.0", default-features = false }
 hex = "0.4"
 itertools = { version = "0.12", default-features = false }
 libtest-mimic = "0.7.0"
--- a/ec/Cargo.toml
+++ b/ec/Cargo.toml
@@ -27,8 +27,13 @@ num-integer.workspace = true
 rayon = { workspace = true, optional = true }
 zeroize = { workspace = true, features = ["zeroize_derive"] }
 hashbrown.workspace = true
-fnv.workspace = true
 itertools.workspace = true
+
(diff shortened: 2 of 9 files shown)


Pratyush:
How about we use SipHash for non-atomic platforms, and ahash elsewhere?

shamatar:
To be honest I'm not sure from top of my head if I can make cargo dependency optional based on whether a platform has atomics. Will try

shamatar:
Another question is whether or not try to include aHash with or without randomization if it's done as external dependency instead of default hash in hashbrown. Hashbrown itself doesn't have and option to continue aHash's features, but maybe it can be imported independently and without randomization is may have no atomics by itself

Pratyush:
> To be honest I'm not sure from top of my head if I can make cargo dependency optional based on whether a platform has atomics. Will try

I think it's possible like so: https://github.com/PyO3/pyo3/blob/fe79f548174eb8108813f202bb0df9428ddfd806/Cargo.toml#L48

shamatar:
restored features (inlines) on hashbrown + ahash is used on platforms with atomics

Pratyush (approved):

Pratyush (approved):

autquis:
Hi! Can anyone give me some pointers on why the `default-hasher` of `hashbrown` is not suitable for us? They have changed it to FoldHash orlp/fold hash.

shamatar:
Mainly because it was using atomics and would not compile on some platforms

shamatar:
I didn’t check FoldHash. The main problem was that aHash was seeded, and seeing mechanism itself used atomics. So it may apply to FoldHash too as seeding is standard mechanism to avoid timing attacks on hash maps if queries could be externally provided/uncontrolled. For arkworks it’s not needed, but default configuration was to use such mechanism

autquis:
Thanks a lot for the pointers! I just checked the seeding mechanism in FoldHash https://github.com/orlp/foldhash/blob/c55c471921ec68e07fdf1623a5d1fa6cc3f09809/src/seed.rs#L310 and it seems they provide a fixed global state for the platforms that do not support atomics.
I would appreciate it very much if you could have a look and confirm if I understood correctly.

In that case, I can go ahead and enable `default-hasher` from `hashbrown` and would simplify things.

autquis:
Hi again! I would be very grateful if you could look into this.
I believe FoldHash is similar to FNV in the sense of seeding on non-atomic systems. A thumbs up from you would suffice :)

shamatar:
What’s the interest for the default hasher? It’s just a feature flag and only provides few extra constructor methods. Maybe Default::default() starts to work, but not sure. At the moment I do not have capacity to check building on atomic-free platforms. Seeding code will most likely work, but it also depends if hashbrown doesn’t do some manual seeding itself

autquis:
Thanks for the answer.
Regarding the interest in the default hasher: I think it simplifies the code. Currently, we have this conditional dependency on aHash and FNV, while having this feature enabled (assuming that FoldHash works for both atomic/non-atomic) we can get rid of the conditional dependency + we will have more concise code for instantiating a hashmap.

I will need this in some other repos of arkworks as well.

PR #811, opened 2024-03-25
Relax `Arc` cfg; remove `Send` bound from `Valid`; implement `Valid` & `CanonicalDeserialize` for `Rc`

mmagician:
## Description

closes: https://github.com/arkworks-rs/algebra/issues/649

depends on https://github.com/arkworks-rs/std/pull/48

---

Before we can merge this PR, please make sure that all the following items have been
checked off. If any of the checklist items are not applicable, please leave them but
write a little note why.

- [x] Targeted PR against correct branch (master)
- [x] Linked to GitHub issue with discussion and accepted design OR have an explanation in the PR that describes this work.
- [x] Wrote unit tests
- [x] Updated relevant documentation in the code
- [x] Added a relevant changelog entry to the `Pending` section in `CHANGELOG.md`
- [x] Re-reviewed `Files changed` in the GitHub PR explorer

PR #833, opened 2024-05-29
Fix the missing `parallel` feature flag for ark-serialize

weikengchen:
## Description

This appears to be the blocker for a number of PRs since the latest test no longer allows a feature that is not declared in the Cargo.toml.

---

write a little note why.

N/A:
- [ ] Wrote unit tests

### DIFF
--- a/serialize/src/lib.rs
+++ b/serialize/src/lib.rs
@@ -69,16 +69,8 @@ pub trait Valid: Sized {
     where
         Self: 'a,
     {
-        #[cfg(feature = "parallel")]
-        {
-            use rayon::{iter::ParallelBridge, prelude::ParallelIterator};
-            batch.par_bridge().try_for_each(|e| e.check())?;
-        }
-        #[cfg(not(feature = "parallel"))]
-        {
-            for item in batch {
-                item.check()?;
-            }
+        for item in batch {
+            item.check()?;
         }
         Ok(())
     }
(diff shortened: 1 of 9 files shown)


weikengchen:
Actually this is not as easy as I think: the problem is that we also implement Valid for things like Rc<T>, but such structure is not "sendable".

weikengchen:
Should we comment out the current parallel implementation then? The problem seems to be as follows: parallel check always requires the elements to be sendable to other threads, and therefore this places a limitation

Pratyush:
This is a problem introduced by https://github.com/arkworks-rs/algebra/pull/811. I would rather that we revert the part there with removes `Valid: Send` instead of giving up on parallel check

weikengchen:
I made an edit---now it has a different implementation.

Since all the batch_check requests would be reduced to the base primitives, and many trivial data structures may not benefit from parallel check:

- now, the parallel check is not going to be applied to `Valid` in general
- the base primitive needs to opt-in to accept parallel check
- now, only EC points and pairing element are opted in

weikengchen:
limitation of the current implementation: it copies.

serialize/src/lib.rs
```diff
@@ -241,3 +233,28 @@ pub fn buffer_bit_byte_size(modulus_bits: usize) -> (usize, usize) {
...
+    #[cfg(feature = "parallel")]
+    {
+        use ark_std::vec::Vec;
+        use rayon::prelude::{IntoParallelIterator, ParallelIterator};
+        batch
+            .map(|x| (*x).clone())
+            .collect::<Vec<T>>()
+            .into_par_iter()
+            .try_for_each(|e| e.check())?;
+    }
```
Pratyush, line 252: Why does the `par_bridge` solution not work here?

mmagician: `ParallelBridge` is implemented for `Iterator + Send`. We no longer have `Send` bound on `batch` here. So we'd need to put an explicit bound: `batch: impl Iterator<Item = &'a T> + Send`, but this propagates all the way up to `Valid`, and would in turn lead to two definitions of `batch_check`:

```
    #[cfg(feature = "parallel")]
    fn batch_check<'a>(batch: impl Iterator<Item = &'a Self> + ParallelBridge)...

    fn batch_check<'a>(batch: impl Iterator<Item = &'a Self>) ...
```

Issue #834, opened 2024-06-16
Deserialization of array can panic


ggwpez:
The deserialisation code of a static sized array uses `unwrap` instead of propagating the error:  

https://github.com/arkworks-rs/algebra/blob/065cd24fc5ae17e024c892cee126ad3bd885f01c/serialize/src/impls.rs#L481-L483

This was possibly done since the `try_from_fn` function is experimental. However, it can cause a panic instead of an error return.

I would propose to use a vector with capacity. There is probably a way to make it a faster with `unsafe`, but thats not my speciality:

```rust
let mut result = Vec::<T>::with_capacity(N);
for i in 0 .. N {
    result.push(T::deserialize_with_mode(&mut reader, compress, validate)?);
}

let array: [T; N] = result.try_into().map_err(|_| SerializationError::InvalidData)?;
Ok(array)
```

burdges:
Yeah, this is a serious bug.  nice catch.

There are crates that do this, like [arrayvec](https://docs.rs/arrayvec/latest/arrayvec/struct.ArrayVec.html), but the unsafe solution looks like this:
```
impl<T: CanonicalDeserialize, const N: usize> CanonicalDeserialize for [T; N] {
    #[inline]
    fn deserialize_with_mode<R: Read>(
        mut reader: R,
        compress: Compress,
        validate: Validate,
    ) -> Result<Self, SerializationError> {
        let mut array = MaybeUninit::uninit_array::<N>();
        for i in 0..N {
            array[i].write(T::deserialize_with_mode(&mut reader, compress, Validate::No) ?);
        }
        if let Validate::Yes = validate {
            T::batch_check(result.iter())?
        }
        Ok(unsafe { MaybeUninit::array_assume_init(array) })
    }
}
```
What this unsafe is solving is the possibility that no sensible default initilization exists for `T`.

Pratyush:
We want to avoid allocation when possible, but the `MaybeUninit` solution looks reasonable to me. Would folks be interested in opening a PR?

burdges:
Appears arrayvec already gets used in ff, so maybe that isolated the unsafety better.
```
impl<T: CanonicalDeserialize, const N: usize> CanonicalDeserialize for [T; N] {
    #[inline]
    fn deserialize_with_mode<R: Read>(
        mut reader: R,
        compress: Compress,
        validate: Validate,
    ) -> Result<Self, SerializationError> {
        let mut array = ArrayVec::<T, N>::new();
[...]

PR #837, opened 2024-06-19
Fix Array Deserialization Panic

ggwpez:
## Description

Use `ArrayVec` - as suggested [here](https://github.com/arkworks-rs/algebra/issues/834#issuecomment-2178950139) - to remove an `unwrap` in the decoding of arrays.

closes: #834

---

write a little note why.

- [x] Wrote unit tests

### DIFF
--- a/serialize/src/impls.rs
+++ b/serialize/src/impls.rs
@@ -470,7 +470,7 @@ impl<T: CanonicalDeserialize, const N: usize> CanonicalDeserialize for [T; N] {
         if let Validate::Yes = validate {
             T::batch_check(array.iter())?
         }
-        Ok(array.into_inner().unwrap_or_else(|_| unreachable!()))
+        Ok(array.into_inner().ok().unwrap())
     }
 }
 
--- a/serialize/src/impls.rs
+++ b/serialize/src/impls.rs
@@ -461,19 +461,16 @@ impl<T: CanonicalDeserialize, const N: usize> CanonicalDeserialize for [T; N] {
     ) -> Result<Self, SerializationError> {
         let mut array = ArrayVec::<T, N>::new();
         for _ in 0..N {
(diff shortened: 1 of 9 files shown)


burdges:
Just use arrayvec here, my code stupidly had a memopry leak if `T` allocates.  https://github.com/arkworks-rs/algebra/issues/834#issuecomment-2178950139

serialize/src/impls.rs
```diff
@@ -458,13 +459,21 @@ impl<T: CanonicalDeserialize, const N: usize> CanonicalDeserialize for [T; N] {
...
+        }
         if let Validate::Yes = validate {
-            T::batch_check(result.iter())?
+            T::batch_check(array.iter())?
         }
-        Ok(result)
+        Ok(array
+            .into_inner()
+            // Defensive error return:
+            .map_err(|_| SerializationError::NotEnoughSpace)?)
```
mmagician, line 476: can this line ever throw an error? It looks like `NotEnoughSpace` should have been caught in the loop above

ggwpez: It should not, which is why i commented `Defensive`.  
I can also unwrap it if you prefer, but then `T: Debug` is required.

ggwpez: It [does](https://doc.rust-lang.org/std/result/enum.Result.html#method.expect)... but i can use  an `.unwrap_or_else(|_| unreachable!())`.

Pratyush: Oh I was looking at `Option::unwrap()` lol

burdges: `.map_err(|_| ()).unwrap()` works too.

Pratyush (approved):

mmagician (approved):

PR #846, opened 2024-07-02
affine versioned pairing supporting for BitVM

PayneJoe:
## Description

This PR propose is to give caller an another more option to choose when do pairing. In current implementation, the coefficients of divisor line function is computed through projective coordinates as it's more efficient (than affine coordinates) when computing pairings, while it's not when verifying pairings (recursive snark). As Andrija and Liam stated in 5.2 section of [On Proving Pairings](https://eprint.iacr.org/2024/640.pdf). 

So for supporting verifying pairings in [BitVM](https://github.com/BitVM/BitVM), I add one more option method for pairing. 

closes: #XXXX

---

write a little note why.

- [ ] Wrote unit tests

### DIFF
--- a/ec/src/models/bn/g2.rs
+++ b/ec/src/models/bn/g2.rs
@@ -40,6 +40,88 @@ pub struct G2HomProjective<P: BnConfig> {
     z: Fp2<P::Fp2Config>,
 }
 
+impl<P: BnConfig> G2Prepared<P> {
+    fn affine_double_in_place(t: &mut G2Affine<P>, three_div_two: &P::Fp) -> EllCoeff<P> {
+        //  for affine coordinates
+        //  slope: alpha = 3 * x^2 / 2 * y
+        let mut alpha = t.x.square();
+        alpha /= t.y;
+        alpha.mul_assign_by_fp(&three_div_two);
+        let bias = t.y - alpha * t.x;
+
+        // update T
+        // T.x = alpha^2 - 2 * t.x
+        // T.y = -bias - alpha * T.x
+        let tx = alpha.square() - t.x.double();
+        t.y = -bias - alpha * tx;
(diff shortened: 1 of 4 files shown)


swasilyev:
So this pairing computation results in another target group element?

PayneJoe:
> So this pairing computation results in another target group element?

the test result proves that it's equal with current pairing result (same element of Gt), even so for the sake of consistency with script implementation in BitVM we need to a benchmark, this pairing in arkworks is my benchmark

swasilyev:
Eh. Why then we need code that computes the same result in a supposedly slower way?

PayneJoe:
> Eh. Why then we need code that computes the same result in a supposedly slower way?

Okay, let me introduce a naive and toy case to explain why we need that, what the benchmark actually mean.

If I want to implement a division algorithm for two numbers (whatever numbers is okay, even big integer) with javascript language, say `a / b`. I might need to benchmark to test if the algorithm is right. By now we do not have the same algorithm (a / b) implementation in C++, we only have the equivalence algorithm, say `a * 1/b`. 

If use `javascript(a / b) == c++(a * 1/b)` in my unit test, someone else may argue that the algorithms are not consistent with each other, `c++(a / b)` is accurate benchmark, and this is what I want.

PayneJoe:
> Eh. Why then we need code that computes the same result in a supposedly slower way?

PS: Most of the time, you are do computing pairings (in snark verifier), so projective is your option, if you are do verifying pairings, affine is your option. It's the best option for recursive snark.

weikengchen:
I think what will happen with this PR is that it would be held off until the R1CS gadget for verifying affine pairing is done. If the value is the same, we do not need this in algebra, and the R1CS witness generation would be in R1CS.

PR #850, opened 2024-08-11
Infallible div by vanishing poly

davxy:
As it can't fail maybe is not worth returning an `Option`. Saves some "dummy" unwraps in downstream

---

This trivially changes the interface, I'm not sure if is really worth mentioning in Changelog


weikengchen:
I agree with the change in that, throughout the entire arkworks org, we do always unwrap it. The logic seems to hold even if `domain_size = 0`.

Pratyush (approved):

PR #852, opened 2024-08-23
perf: ATE_LOOP_COUNT hamming weight reduced from 26 to 22

FatihSolak:
## Description

ATE_LOOP_COUNT in the BN254 curve has been updated, so the new version has a 22 hamming weight, while previous one had 26. 

---


Pratyush:
Thank you for the PR! Could you add a test showing that the two miller loops are equivalent, or some other justification of how you were able to reduce the hamming weight? Thank you!

Hakkush-07:
Hi, we noticed the current ATE_LOOP_COUNT in BN254 curve is not in NAF form since it has consecutive non-zero elements at a few places. It has 26 non-zero elements and its length is 65.
`[0, 0, 0, 1, 0, 1, 0, -1, 0, 0, 1, -1, 0, 0, 1, 0, 0, 1, 1, 0, -1, 0, 0, 1, 0, -1, 0, 0, 0, 0, 1, 1, 1, 0, 0, -1, 0, 0, 1, 0, 0, 0, 0, 0, -1, 0, 0, 1, 1, 0, 0, -1, 0, 0, 0, 1, 1, 0, -1, 0, 0, 1, 0, 1, 1]`

This is however, is still a signed-digit representation of the correct value $6x+2=29793968203157093288$ but not the most efficient form which is NAF. It being in this form does not affect the correctness of calculations but less Hamming weight is better. Applying the [algorithm](https://en.wikipedia.org/wiki/Non-adjacent_form#Converting_to_NAF) on Wikipedia gives the following result:
`[0, 0, 0, 1, 0, 1, 0, -1, 0, 0, -1, 0, 0, 0, 1, 0, 0, -1, 0, -1, 0, 0, 0, 1, 0, -1, 0, 0, 0, 0, -1, 0, 0, 1, 0, -1, 0, 0, 1, 0, 0, 0, 0, 0, -1, 0, 0, -1, 0, 1, 0, -1, 0, 0, 0, -1, 0, -1, 0, 0, 0, 1, 0, -1, 0, 1]`

This version has Hamming weight 22 but has length 66. However, the last 3 bits in this representation (..., -1, 0, 1) is equivalent to the last 2 bits of the current version (..., 1, 1). That is because 
$(-1)\cdot 2^k+(0)\cdot 2^{k+1}+(1)\cdot 2^{k+2}=(1)\cdot 2^k+(1)\cdot 2^{k+1}$
Utilizing that equivalence, we arrive at the following representation with 22 Hamming weight and length of 65:
`[0, 0, 0, 1, 0, 1, 0, -1, 0, 0, -1, 0, 0, 0, 1, 0, 0, -1, 0, -1, 0, 0, 0, 1, 0, -1, 0, 0, 0, 0, -1, 0, 0, 1, 0, -1, 0, 0, 1, 0, 0, 0, 0, 0, -1, 0, 0, -1, 0, 1, 0, -1, 0, 0, 0, -1, 0, -1, 0, 0, 0, 1, 0, 1, 1]`

Pratyush:
Great, thank you!

Pratyush (approved):

PR #864, opened 2024-11-05
Implemented add, mult, inv, sub, div, pow and neg for binary fields and wrote tests for each function

martinsander00:
Only file to review is bf.rs inside /algebra/ff/src/fields

Description:
Implemented add, mult, inv, sub, div, pow and neg for binary fields and wrote tests for each function. 

## Description

closes: #XXXX

---

write a little note why.

- [ ] Wrote unit tests

### DIFF
--- a/ff/src/fields/bf.rs
+++ b/ff/src/fields/bf.rs
@@ -1,0 +1,153 @@
+use num_bigint::BigUint;
+use num_traits::{Zero, One};
+
+#[derive( Clone, PartialEq, Eq)]
+pub struct BinaryFieldElement {
+    value: BigUint,
+}
+
+impl BinaryFieldElement {
+    // Constructor to create a new BinaryFieldElement
+    pub fn new(value: BigUint) -> Self {
+        BinaryFieldElement { value }
+    }
+
+    // Method for addition in GF(2^k)
+    pub fn add(&self, other: &Self) -> Self {
+        // Addition in binary fields is done with XOR
+        BinaryFieldElement::new(&self.value ^ &other.value)
+    }
+    
+    pub fn sub(&self, other: &Self) -> Self {
+        self.add(other)
+    }
+
+    // Method for multiplication in GF(2^k)
+    pub fn mul(&self, other: &Self) -> Self {
+        BinaryFieldElement::new(binmul(self.value.clone(), other.value.clone(), None))
+    }
+
+    // Method for division (multiplication by the inverse)
+    pub fn div(&self, other: &Self) -> Self {
+        self.mul(&other.inv()) // Division as multiplication by inverse
+    }
+
+    pub fn neg(&self) -> Self {
+        self.clone()
+    }
+    
+    pub fn pow(&self, exponent: u64) -> Self {
+        if exponent == 0 {
+            BinaryFieldElement::new(BigUint::one())
+        } else if exponent == 1 {
+            self.clone()
+        } else if exponent % 2 == 0 {
+            let half_pow = self.pow(exponent / 2);
+            half_pow.mul(&half_pow)
+        } else {
+            self.mul(&self.pow(exponent - 1))
+        }
+    }
+
+    // In a binary field GF(2^w) the inverse is a^{-1} = a^{2m-2}
+    pub fn inv(&self) -> Self {
+        let bit_len = self.value.bits();
+        let l = 1 << (bit_len - 1).ilog2() + 1;
+        return self.pow(2_u64.pow(l) - 2);
+    }
+}
+
+// Binary field multiplication with reduction
+fn binmul(v1: BigUint, v2: BigUint, length: Option<usize>) -> BigUint {
+    if v1.is_zero() || v2.is_zero() {
+        return BigUint::zero();
+    }
+    if v1.is_one() || v2.is_one() {
+        return v1 * v2;
+    }
+
+    let length = length.unwrap_or_else(|| 1 << (v1.bits().max(v2.bits()) - 1).next_power_of_two());
+    let halflen = length / 2;
+    let quarterlen = halflen / 2;
+    let halfmask = (BigUint::one() << halflen) - BigUint::one();
+
+    let l1 = &v1 & &halfmask;
+    let r1 = &v1 >> halflen;
+    let l2 = &v2 & &halfmask;
+    let r2 = &v2 >> halflen;
+
+    // Optimized case for (L1, R1) == (0, 1)
+    if l1.is_zero() && r1 == BigUint::one() {
+        let out_r = binmul(BigUint::one() << quarterlen, r2.clone(), Some(halflen)) ^ l2.clone();
+        return r2 ^ (out_r << halflen);
+    }
+
+    // Perform Karatsuba multiplication with three main sub-multiplications
+    let l1l2 = binmul(l1.clone(), l2.clone(), Some(halflen));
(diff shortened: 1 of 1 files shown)


ff/src/fields/bf.rs
```diff
@@ -0,0 +1,153 @@
+use num_bigint::BigUint;
+use num_traits::{Zero, One};
+
+#[derive( Clone, PartialEq, Eq)]
+pub struct BinaryFieldElement {
+    value: BigUint,
+}
+
+impl BinaryFieldElement {
+    // Constructor to create a new BinaryFieldElement
```
tcoratger, line 10:
```suggestion
    // Constructor to create a new [`BinaryFieldElement`]
```

ff/src/fields/bf.rs
```diff
@@ -0,0 +1,153 @@
...
+    pub fn new(value: BigUint) -> Self {
+        BinaryFieldElement { value }
```
tcoratger, line 12:
```suggestion
        Self { value }
```

ff/src/fields/bf.rs
```diff
@@ -0,0 +1,153 @@
...
+    }
+
+    // Method for addition in GF(2^k)
+    pub fn add(&self, other: &Self) -> Self {
+        // Addition in binary fields is done with XOR
+        BinaryFieldElement::new(&self.value ^ &other.value)
```
tcoratger, line 18:
```suggestion
        Self::new(&self.value ^ &other.value)
```

ff/src/fields/bf.rs
```diff
@@ -0,0 +1,153 @@
...
+    }
+    
+    pub fn sub(&self, other: &Self) -> Self {
+        self.add(other)
+    }
+
+    // Method for multiplication in GF(2^k)
+    pub fn mul(&self, other: &Self) -> Self {
+        BinaryFieldElement::new(binmul(self.value.clone(), other.value.clone(), None))
```
tcoratger, line 27:
```suggestion
        Self::new(binmul(self.value.clone(), other.value.clone(), None))
```

ff/src/fields/bf.rs
```diff
@@ -0,0 +1,153 @@
...
+}
+
+impl BinaryFieldElement {
+    // Constructor to create a new BinaryFieldElement
+    pub fn new(value: BigUint) -> Self {
+        BinaryFieldElement { value }
+    }
+
+    // Method for addition in GF(2^k)
+    pub fn add(&self, other: &Self) -> Self {
```
tcoratger, line 16: Here why not implementing directly the `Add` trait like this is done here for example? I think this is also applicable to sub, mul, neg... so that we can use directly `+`, `-`, `*` operations

https://github.com/arkworks-rs/algebra/blob/9ce33e6ef1368a0f5b01b91e6df5bc5877129f30/ec/src/pairing.rs#L192-L200

ff/src/fields/bf.rs
```diff
@@ -0,0 +1,153 @@
...
+    // In a binary field GF(2^w) the inverse is a^{-1} = a^{2m-2}
+    pub fn inv(&self) -> Self {
+        let bit_len = self.value.bits();
+        let l = 1 << (bit_len - 1).ilog2() + 1;
+        return self.pow(2_u64.pow(l) - 2);
+    }
+}
+
+// Binary field multiplication with reduction
+fn binmul(v1: BigUint, v2: BigUint, length: Option<usize>) -> BigUint {
```
tcoratger, line 61: I think we can integrate this method directly inside the `BinaryFieldElement` impl no? using:

```rust
fn binmul(&self, rhs: Self, length: Option<usize>) -> Self {
```

ff/src/fields/bf.rs
```diff
@@ -0,0 +1,153 @@
...
+    let r1r2_high = binmul(BigUint::one() << quarterlen, r1r2.clone(), Some(halflen));
+    let z3 = binmul(l1 ^ r1.clone(), l2 ^ r2.clone(), Some(halflen));
+
+    l1l2.clone() ^ r1r2.clone() ^ ((z3 ^ l1l2 ^ r1r2 ^ r1r2_high) << halflen)
+}
+
+
+
+#[cfg(test)]
+mod tests {
```
tcoratger, line 97: can we add some randomized tests here as this is done here for example? https://github.com/arkworks-rs/algebra/blob/master/ff/src/biginteger/tests.rs

ff/src/fields/bf.rs
```diff
@@ -0,0 +1,153 @@
+use num_bigint::BigUint;
+use num_traits::{Zero, One};
+
+#[derive( Clone, PartialEq, Eq)]
+pub struct BinaryFieldElement {
+    value: BigUint,
+}
```
tcoratger, line 7: Here I'm not sure this way of doing things is the more efficient approach because no matter the size of the field concerned, you store in your `value` as `BigUint` a vector of `BigDigit` (u64 or 32 I think) but for field extensions before 32, we don't need as much space.

So I wonder if it is not more optimized to do something custom made like this:

https://gitlab.com/IrreducibleOSS/binius/-/blob/main/crates/field/src/binary_field.rs?ref_type=heads#L770-L777

so that we can have a very efficient implementation for small extensions.

ff/src/fields/bf.rs
martinsander00: Will do this! Thanks for the suggestion

ff/src/fields/bf.rs
martinsander00: SO we could, but why would someone use binmul instead of mul? I think mul is more intuitive right?

ff/src/fields/bf.rs
martinsander00: This makes a lot of sense, but it would mean implementing a separate Binary Field implementation for each field size. Should I submit these changes as part of my next PR, or should I first push the other changes we discussed (operators and moving binmul) and then implement the different field sizes in subsequent PRs?

ff/src/fields/bf.rs
martinsander00: done! Thanks for the suggestion

ff/src/fields/bf.rs
martinsander00: @tcoratger Hi! Hope you are doing well. 

Could you check the changes that I made? I added every change you proposed except for the BigUint efficient implementations. I talked with Pratyush and he told me that I could work on that as soon as everything else looks well. What do you think?

z-tech:
Hi, thank you for opening this PR.

> https://gitlab.com/IrreducibleOSS/binius/-/blob/main/crates/field/src/binary_field.rs?ref_type=heads#L721

This pattern you linked from irreducible might be cool.

In general, every time you half the size of the primitive that holds the field element starting from `u128` you expect ~2x throughput on lots of foundational operations using vectorization. Even without vectorization there is a significant difference. I can suggest that you try running some benches to compare the performance of arithmetic between system primitives with arithmetic between  `BigUInt`s. 

Along those lines, I have a [project](https://github.com/compsec-epfl/space-efficient-sumcheck/pull/65/files) running with students to implement `M31` and `BabyBear` including vectorization. I'm only getting up to speed here but potentially it might make more sense for this effort to happen off of that branch (if at all at this point in time)?

In any case, could you please provide a description of the motivation of the proposed implementation called `BinaryFieldElement`? I think I see where the inspiration came from but I'm not sure what the plan is about how to use it.

Thanks!

PR #869, opened 2024-11-12
poly: add specific unit tests and simplify `DensePolynomial` addition

tcoratger:
## Description

add specific unit tests and simplify `DensePolynomial` addition

- [ ] Wrote unit tests

### DIFF
--- a/poly/src/polynomial/univariate/dense.rs
+++ b/poly/src/polynomial/univariate/dense.rs
@@ -289,7 +289,9 @@ impl<'a, 'b, F: Field> Add<&'a DensePolynomial<F>> for &'b DensePolynomial<F> {
     type Output = DensePolynomial<F>;
 
     fn add(self, other: &'a DensePolynomial<F>) -> DensePolynomial<F> {
+        println!("iciiiiiii");
         let mut result = if self.is_zero() {
+            println!("111111111");
             other.clone()
         } else if other.is_zero() {
             self.clone()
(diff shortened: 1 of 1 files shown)


poly/src/polynomial/univariate/dense.rs
```diff
@@ -378,27 +378,32 @@ impl<'a, F: Field> AddAssign<&'a DensePolynomial<F>> for DensePolynomial<F> {
...
-            self.coeffs.iter_mut().for_each(|c| *c *= &f);
+        // No need to modify self if other is zero
+        if other.is_zero() {
             return;
-        } else if other.is_zero() {
+        }
+
+        // If the first polynomial is zero, just copy the second one and scale by f.
+        if self.is_zero() {
+            self.coeffs = other.coeffs.iter().map(|&c| f * c).collect();
```
Pratyush, line 388: I prefer the older method because it preserves the allocation of `self`, whereas this approach creates a new allocation.

Actually I realized that the old method doesn't preserve allocations either, and we should be using `self.coeffs.clear()` instead of `self.coeffs.truncate(0)`.

Pratyush (commented):
Thank you for the PR! only 1 minor comment.

poly/src/polynomial/univariate/dense.rs
tcoratger: yep agreed, just fixed

Pratyush (approved):
Thank you!

PR #894, opened 2024-11-26
chore: small adjustments

tcoratger:
## Description

@Pratyush Here I propose a slightly different formalism to manage lints compared to what we had before on the repo. This formalism inherits from what is already done on reth https://github.com/paradigmxyz/reth/blob/83af493179393c81dc2c48374306f2b88cc075db/Cargo.toml#L160-L254 with a curated list of lints that have already proven themselves.

In this methodology, lints are directly introduced into the cargo.toml file and then dispatched to all the crates inside the repo via

```
[lints]
workspace = true
```

So it's easier to manage and more uniform. For now I have only applied it to the poly crate but we can adopt this for the other crates in follow up PRs if you agree.

---

write a little note why.

- [ ] Wrote unit tests

### DIFF
--- a/poly/src/evaluations/multivariate/multilinear/mod.rs
+++ b/poly/src/evaluations/multivariate/multilinear/mod.rs
@@ -45,7 +45,21 @@ type DefaultHasher = fnv::FnvHasher;
 /// Index represents a point, which is a vector in {0,1}^`num_vars` in little
 /// endian form. For example, `0b1011` represents `P(1,1,0,1)`
 pub trait MultilinearExtension<F: Field>:
-    Sized + Clone + Debug + Hash + PartialEq + Eq + Add + Neg + Zero + CanonicalSerialize + CanonicalDeserialize + for<'a> AddAssign<&'a Self> + for<'a> SubAssign<&'a Self> + Index<usize> + Polynomial<F, Point = Vec<F>>
+    Sized
+    + Clone
+    + Debug
+    + Hash
+    + PartialEq
+    + Eq
+    + Add
+    + Neg
+    + Zero
+    + CanonicalSerialize
+    + CanonicalDeserialize
+    + for<'a> AddAssign<&'a Self>
+    + for<'a> SubAssign<&'a Self>
+    + Index<usize>
+    + Polynomial<F, Point = Vec<F>>
 {
     /// Returns the number of variables in `self`
     fn num_vars(&self) -> usize;
--- a/poly/src/evaluations/multivariate/multilinear/mod.rs
+++ b/poly/src/evaluations/multivariate/multilinear/mod.rs
@@ -44,6 +44,7 @@ type DefaultHasher = fnv::FnvHasher;
 ///
 /// Index represents a point, which is a vector in {0,1}^`num_vars` in little
 /// endian form. For example, `0b1011` represents `P(1,1,0,1)`
+#[allow(clippy::trait_duplication_in_bounds)]
 pub trait MultilinearExtension<F: Field>:
     Sized
     + Clone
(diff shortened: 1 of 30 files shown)


poly/src/polynomial/univariate/mod.rs
```diff
@@ -15,7 +15,7 @@ pub use sparse::SparsePolynomial;
 use rayon::prelude::*;
 
 /// Represents either a sparse polynomial or a dense one.
-#[derive(Clone)]
+#[derive(Debug, Clone)]
```
tcoratger, line 18:
```suggestion
#[derive(Clone)]
```

poly/src/evaluations/multivariate/multilinear/mod.rs
```diff
@@ -57,7 +57,6 @@ pub trait MultilinearExtension<F: Field>:
     + CanonicalSerialize
     + CanonicalDeserialize
     + for<'a> AddAssign<&'a Self>
-    + for<'a> AddAssign<(F, &'a Self)>
```
Pratyush, line 60: Why was this bound removed?

poly/src/evaluations/multivariate/multilinear/sparse.rs
```diff
@@ -77,12 +77,6 @@ impl<F: Field> SparseMultilinearExtension<F> {
             }
             map.entry(index).or_insert(F::rand(rng));
         }
-        let mut buf = Vec::new();
-        for (arg, v) in map.iter() {
-            if *v != F::zero() {
-                buf.push((*arg, *v));
-            }
-        }
```
Pratyush, line 85: Why was this loop removed?

tcoratger: because buf is never used after this so right now this loop is kind of useless

poly/src/polynomial/mod.rs
```diff
@@ -26,7 +26,6 @@ pub trait Polynomial<F: Field>:
     + CanonicalSerialize
     + CanonicalDeserialize
     + for<'a> AddAssign<&'a Self>
-    + for<'a> AddAssign<(F, &'a Self)>
```
Pratyush, line 29: ditto

poly/src/evaluations/multivariate/multilinear/sparse.rs
Pratyush: I see...

poly/src/evaluations/multivariate/multilinear/mod.rs
tcoratger: I think this was seen as a false positive with `for<'a> AddAssign<&'a Self>`, just fixed

poly/src/polynomial/mod.rs
tcoratger: I think this was seen as a false positive with `for<'a> AddAssign<&'a Self>`, just fixed

Pratyush (approved):
Thanks for the PR, this is great! If you could propagate this to the other crates in the repo as well that would be lovely =). We should also add it to the template https://github.com/arkworks-rs/template so that future repos will inherit it

tcoratger:
> Thanks for the PR, this is great! If you could propagate this to the other crates in the repo as well that would be lovely =). We should also add it to the template https://github.com/arkworks-rs/template so that future repos will inherit it

@Pratyush Yes no problem, will propagate this in follow up PRs, one for each crate so that we have an easier review each time :)

tcoratger:
@Pratyush Here do you want me to do everything in this PR (for all the crates) or in multiple PRs (I think this is easier for review with 1 PR/crate because there are a lot of changes each time)?

Pratyush:
1 PR per crate sounds good!

tcoratger:
> 1 PR per crate sounds good!

@Pratyush perfect now I think everything is ready to merge this one :)

Pratyush (approved):

tcoratger:
Hey @Pratyush is there anything to add here or is it ready to merge, in order to unlock the other PRs about the same topic :) ?

PR #896, opened 2024-11-28
ff: refactor `find_naf` and add unit tests

tcoratger:
## Description

closes: #XXXX

---

write a little note why.

- [ ] Wrote unit tests

### DIFF
--- a/ff/src/biginteger/arithmetic.rs
+++ b/ff/src/biginteger/arithmetic.rs
@@ -435,7 +435,7 @@ mod tests {
     fn test_find_naf_zero() {
         // Test for zero input
         let naf = find_naf(&[0]);
-        assert_eq!(naf, vec![]);
+        assert!(naf.is_empty());
     }
 
     #[test]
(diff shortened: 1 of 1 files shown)


ff/src/biginteger/arithmetic.rs
```diff
@@ -153,53 +152,30 @@ pub fn mac_with_carry(a: u64, b: u64, c: u64, carry: &mut u64) -> u64 {
...
+
         res.push(z);
-        div2(&mut num);
+
+        // Perform an in-place division of `num` by 2, using bit-shifting.
+        num.iter_mut().rev().fold(0, |carry, x| {
+            let next_carry = *x << 63;
+            *x = (*x >> 1) | carry;
+            next_carry
+        });
```
Pratyush, line 178: Thanks for the PR! I think I like the old code structure slightly better since it separates out the arithmetic logic from the algorithm logic. I do like the new implementations of the arithmetic better, so maybe we can go back to the old code structure but with the new arithmetic impls?

tcoratger: Fixed, I've also added a more complete documentation inside the function for the reader to understand the various operations that are performed.

Pratyush (approved):

PR #906, opened 2024-12-15
refactor `sub_assign` `DensePolynomial`

tcoratger:
## Description

closes: #XXXX

---

write a little note why.

- [ ] Wrote unit tests

### DIFF
--- a/poly/src/polynomial/univariate/dense.rs
+++ b/poly/src/polynomial/univariate/dense.rs
@@ -509,8 +509,7 @@ impl<'a, F: Field> Sub<&'a SparsePolynomial<F>> for &DensePolynomial<F> {
     #[inline]
     fn sub(self, other: &'a SparsePolynomial<F>) -> DensePolynomial<F> {
         if self.is_zero() {
-            let result = other.clone();
-            (-result).into()
+            (-other.clone()).into()
         } else if other.is_zero() {
             self.clone()
         } else {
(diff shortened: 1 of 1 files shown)


poly/src/polynomial/univariate/dense.rs
```diff
@@ -509,8 +509,7 @@ impl<'a, F: Field> Sub<&'a SparsePolynomial<F>> for &DensePolynomial<F> {
     #[inline]
     fn sub(self, other: &'a SparsePolynomial<F>) -> DensePolynomial<F> {
         if self.is_zero() {
-            let result = other.clone();
-            (-result).into()
+            (-other.clone()).into()
```
Pratyush, line 512: I think to emphasize the order of operations, this should be 
```suggestion
            (-(other.clone())).into()
```
Since negation takes its argument by value.

poly/src/polynomial/univariate/dense.rs
```diff
@@ -538,24 +537,21 @@ impl<'a, F: Field> SubAssign<&'a DensePolynomial<F>> for DensePolynomial<F> {
...
     fn sub_assign(&mut self, other: &'a DensePolynomial<F>) {
         if self.is_zero() {
-            self.coeffs.resize(other.coeffs.len(), F::zero());
-        } else if other.is_zero() {
-            return;
-        } else if self.degree() >= other.degree() {
-        } else {
-            // Add the necessary number of zero coefficients.
-            self.coeffs.resize(other.coeffs.len(), F::zero());
+            self.coeffs = other.coeffs.iter().map(|&x| -x).collect();
```
Pratyush, line 540: The hope with the old way was that if self.coeffs has capacity large enough (e.g., it was the result of a previous operation that resulted in a zero), then we can avoid allocation.

tcoratger: Oh yes you are right didn't notice that. Then there is no point to refactor this, this is already optimized, let us close :)

PR #915, opened 2025-01-04
Update CHANGELOG: Fix handling of duplicate terms in SparsePolynomial

woojucy:
## Description

In the current implementation of `univariate::SparsePolynomial::from_coefficients_vec`, duplicate terms with the same degree are not explicitly handled. As a result, if multiple terms with the same degree are provided, they remain as separate entries in the resulting sparse polynomial. This can lead to inconsistencies in polynomial operations and unnecessary redundancy.

**Current Behavior:**  
Given an input like:  
```rust
vec![
    (0, F::from(2)), 
    (0, F::from(1)), 
    (0, F::from(1)), 
    (3, F::from(5)),
    (3, F::from(3)),
    (4, F::from(0)),
]
```
The resulting polynomial contains duplicate terms without merging them.

**Proposed Improvement:**  
1. **Sort Terms by Degree:** Ensure terms are sorted by degree.  
2. **Merge Duplicate Terms:** For terms with the same degree, sum their coefficients into a single term.  
3. **Remove Zero Coefficient Terms:** Eliminate terms where the coefficient becomes zero after merging.  

**Proposed Implementation Example:**  
```rust
fn from_coefficients_vec(mut coeffs: Vec<(usize, F)>) -> Self {
    // Sort terms by degree
    coeffs.sort_by(|(c1, _), (c2, _)| c1.cmp(c2));

    // Merge duplicate terms
    let mut merged_coeffs: Vec<(usize, F)> = Vec::new();
    for (degree, coeff) in coeffs {
        if let Some((last_degree, last_coeff)) = merged_coeffs.last_mut() {
            if *last_degree == degree {
                *last_coeff += coeff; // Combine coefficients for duplicate degrees
                continue;
            }
        }
        merged_coeffs.push((degree, coeff));
    }

    // Remove zero coefficient terms
    merged_coeffs.retain(|(_, coeff)| !coeff.is_zero());

    // Create the SparsePolynomial
    Self { coeffs: merged_coeffs }
}
```

**Expected Behavior:**  
For the same input:  
```rust
[(0, F::from(4)), (3, F::from(8))]
```
The resulting polynomial will now be consistent and free from redundant terms.

**Why is this change beneficial?**  
- Ensures clarity and consistency in sparse polynomial representations.  
- Prevents redundant terms, which may cause unexpected results in polynomial operations.  
- Aligns with expected mathematical representations of polynomials.

---

write a little note why.

- [ ] Wrote unit tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@
 
 ### Bugfixes
 
-- (`ark-poly`) Fix handling of duplicate terms in `SparsePolynomial::from_coefficients_vec`. Terms with the same degree are now correctly merged, and zero coefficient terms are removed.
+- [\#915](https://github.com/arkworks-rs/algebra/pull/915) (`ark-poly`) Added `is_valid_coefficients_vec` to `SparsePolynomial::from_coefficients_vec` to remove duplicates and zero coefficients before polynomial creation.
 
 ## v0.5.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@
 
 ### Bugfixes
 
+- (`ark-poly`) Fix handling of duplicate terms in `SparsePolynomial::from_coefficients_vec`. Terms with the same degree are now correctly merged, and zero coefficient terms are removed.
+
 ## v0.5.0
 
 - [\#772](https://github.com/arkworks-rs/algebra/pull/772) (`ark-ff`) Implementation of `mul` method for `BigInteger`.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@
 
 ### Bugfixes
 
-- [\#915](https://github.com/arkworks-rs/algebra/pull/915) (`ark-poly`) Added `is_valid_coefficients_vec` to `SparsePolynomial::from_coefficients_vec` to remove duplicates and zero coefficients before polynomial creation.
+- [\#915](https://github.com/arkworks-rs/algebra/pull/915) (`ark-poly`) Added `is_valid_coefficients_vec` to `SparsePolynomial` to remove duplicates and zero coefficients before polynomial creation.
 
 ## v0.5.0
 
--- a/poly/src/polynomial/univariate/sparse.rs
+++ b/poly/src/polynomial/univariate/sparse.rs
@@ -230,33 +230,44 @@ impl<F: Field> Zero for SparsePolynomial<F> {
 }
 
 impl<F: Field> SparsePolynomial<F> {
+
+    // Function to validate that the input coefficient vector is simplified
+    pub fn is_valid_coefficients_vec(coeffs: &[(usize, F)]) -> bool {
+        let mut last_degree = None;
+        for &(degree, ref coeff) in coeffs {
+            if coeff.is_zero() {
+                return false; // Zero coefficients are not allowed
+            }
+            if let Some(last) = last_degree {
+                if last == degree {
+                    return false; // Duplicate degrees are not allowed
+                }
+            }
+            last_degree = Some(degree);
+        }
+        true
+    }
+    
     /// Constructs a new polynomial from a list of coefficients.
     pub fn from_coefficients_slice(coeffs: &[(usize, F)]) -> Self {
         Self::from_coefficients_vec(coeffs.to_vec())
     }
 
     /// Constructs a new polynomial from a list of coefficients.
+    /// The function does not combine like terms and so multiple monomials
+    /// of the same degree are ignored.
     pub fn from_coefficients_vec(mut coeffs: Vec<(usize, F)>) -> Self {
-        // Sort terms by degree
-        coeffs.sort_by(|(c1, _), (c2, _)| c1.cmp(c2));
-    
-        // Merge duplicate terms
-        let mut merged_coeffs: Vec<(usize, F)> = Vec::new();
-        for (degree, coeff) in coeffs {
-            if let Some((last_degree, last_coeff)) = merged_coeffs.last_mut() {
-                if *last_degree == degree {
-                    *last_coeff += coeff; // Combine coefficients for duplicate degrees
-                    continue;
-                }
-            }
-            merged_coeffs.push((degree, coeff));
+        // While there are zeros at the end of the coefficient vector, pop them off.
+        while coeffs.last().map_or(false, |(_, c)| c.is_zero()) {
+            coeffs.pop();
         }
-    
-        // Remove zero coefficient terms
-        merged_coeffs.retain(|(_, coeff)| !coeff.is_zero());
-    
-        // Create the SparsePolynomial
-        Self { coeffs: merged_coeffs }
+        // Ensure that coeffs are in ascending order.
+        coeffs.sort_by(|(c1, _), (c2, _)| c1.cmp(c2));
+        // Check that either the coefficients vec is empty or that the last coeff is
+        // non-zero.
+        assert!(coeffs.last().map_or(true, |(_, c)| !c.is_zero()));
+
+        Self { coeffs }
     }
 
     /// Perform a naive n^2 multiplication of `self` by `other`.
--- a/poly/src/polynomial/univariate/sparse.rs
+++ b/poly/src/polynomial/univariate/sparse.rs
@@ -241,6 +241,12 @@ pub enum CoeffVecError {
     DuplicateDegree,
 }
 
+#[derive(Debug)]
+pub enum CoeffVecError {
+    ZeroCoefficient,
+    DuplicateDegree,
+}
+
 impl<F: Field> SparsePolynomial<F> {
 
     // Function to validate that the input coefficient vector is simplified
--- a/poly/src/polynomial/univariate/sparse.rs
+++ b/poly/src/polynomial/univariate/sparse.rs
@@ -236,20 +236,27 @@ impl<F: Field> SparsePolynomial<F> {
     }
 
     /// Constructs a new polynomial from a list of coefficients.
-    /// The function does not combine like terms and so multiple monomials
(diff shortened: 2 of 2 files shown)


z-tech:
Hi, thank you for raising this PR

1) Do we think this case of inputs with duplicate terms is occurring organically?
2) Have we considered instead that the function should reject when the vector is not already simplified?

At the least, approach 2 will force callers to become aware of situation 1, which seems to be the focus of the PR.

To do this safely, I think it would be more prudent to open a function call with a strongly defined "valid input" and then later deprecate this one so that callers can at their own pace address the potential unexpected results in polynomial operations mentioned.

What are thoughts?

woojucy:
That's a good point. We need to examine how frequently duplicate terms naturally occur.  

Additionally, we should also consider the issue of "consistency" in the function's behavior. For example, comparing the two polynomial cases below:  

1. **`univariate::SparsePolynomial` example:**  
   ```rust  
   let poly = univariate::SparsePolynomial::from_coefficients_vec(vec![
       (0, Fq::from(2)), 
       (0, Fq::from(1)), 
       (0, Fq::from(1)), 
       (3, Fq::from(5)),
       (3, Fq::from(3)),
       (4, Fq::from(0)),
   ]);
   ```  

2. **`multivariate::SparsePolynomial` example:**  
   ```rust  
   let poly2 = multivariate::SparsePolynomial::from_coefficients_vec(
       1,
       vec![
           (Fq::from(0), SparseTerm::new(vec![(0, 4)])),
           (Fq::from(3), SparseTerm::new(vec![(0, 3)])),
           (Fq::from(5), SparseTerm::new(vec![(0, 3)])),
           (Fq::from(1), SparseTerm::new(vec![])),
           (Fq::from(1), SparseTerm::new(vec![])),
           (Fq::from(2), SparseTerm::new(vec![])),
       ],
   );
   ```  

In the first case, it seems the function does not handle duplicate terms explicitly, while in the second case, the input is considered valid and behaves as expected.  While we would expect both polynomials to behave consistently, the first function appears to lack clear input validation, resulting in inconsistency.  

To address this issue, do you think introducing a new function call that enforces valid input, followed by a gradual deprecation of the current function, would be sufficient as you suggested?  

I’d appreciate your thoughts on this.

z-tech:
> That's a good point. We need to examine how frequently duplicate terms naturally occur.
> 
> Additionally, we should also consider the issue of "consistency" in the function's behavior. For example, comparing the two polynomial cases below:
> 
> 1. **`univariate::SparsePolynomial` example:**
>    ```rust
>    let poly = univariate::SparsePolynomial::from_coefficients_vec(vec![
>        (0, Fq::from(2)), 
>        (0, Fq::from(1)), 
>        (0, Fq::from(1)), 
>        (3, Fq::from(5)),
>        (3, Fq::from(3)),
>        (4, Fq::from(0)),
>    ]);
>    ```
> 2. **`multivariate::SparsePolynomial` example:**
>    ```rust
>    let poly2 = multivariate::SparsePolynomial::from_coefficients_vec(
>        1,
>        vec![
>            (Fq::from(0), SparseTerm::new(vec![(0, 4)])),
>            (Fq::from(3), SparseTerm::new(vec![(0, 3)])),
>            (Fq::from(5), SparseTerm::new(vec![(0, 3)])),
>            (Fq::from(1), SparseTerm::new(vec![])),
>            (Fq::from(1), SparseTerm::new(vec![])),
>            (Fq::from(2), SparseTerm::new(vec![])),
>        ],
>    );
>    ```
> 
> In the first case, it seems the function does not handle duplicate terms explicitly, while in the second case, the input is considered valid and behaves as expected. While we would expect both polynomials to behave consistently, the first function appears to lack clear input validation, resulting in inconsistency.
> 
> To address this issue, do you think introducing a new function call that enforces valid input, followed by a gradual deprecation of the current function, would be sufficient as you suggested?
> 
> I’d appreciate your thoughts on this.

I was thinking maybe a function `is_valid_coefficients_vec` would suffice? In any case, do you think that seems like reasonable place to start?

woojucy:
I fully agree with the approach of adding an is_valid_coefficients_vec function to ensure the input vector is pre-validated. I have considered the method where the function returns an error when the vector is not simplified.

Implementing this change would require modifying the return type of from_coefficients_vec. This means that all functions calling from_coefficients_vec would need changes in their signatures and additional error handling, leading to several modifications.

I would like to discuss this issue further. Are there any better ideas or conventional approaches available? If there is a consensus that this is the right direction, I am ready to proceed with implementing these changes. Your feedback will be crucial in making this decision.

z-tech:
It would be great if you could implement the new function `is_valid_coefficients_vec`. I don't think any existing functions need to be modified. This is a sequence diagram of how I see it working.

Screenshot 2025-02-19 at 20 28 39

If we do it this way we include the verification logic you've suggested and at the same time we insure backward compatibility.

What are you thoughts?

woojucy:
Originally, I considered calling the `is_valid_coefficients_vec` function directly within the `from_coefficients_vec` function. Thankfully, the function call diagram you provided made it immediately clear what you meant. Based on your feedback, I updated the commit by explicitly calling the `is_valid_coefficients_vec` function. This function now checks for duplicate terms and zero coefficients. Please review the updated implementation.

poly/src/polynomial/univariate/sparse.rs
```diff
@@ -230,6 +230,24 @@ impl<F: Field> Zero for SparsePolynomial<F> {
...
+            }
+            if let Some(last) = last_degree {
+                if last == degree {
+                    return false; // Duplicate degrees are not allowed
+                }
+            }
+            last_degree = Some(degree);
+        }
+        true
+    }
```
z-tech, line 249:
```suggestion
    pub fn is_valid_coefficients_vec(coeffs: &[(usize, F)]) -> Result<(), CoeffVecError> {
        let mut last_degree = None;
        for &(degree, ref coeff) in coeffs {
            if coeff.is_zero() {
                // zero term has no effect
                return Err(CoeffVecError::ZeroCoefficient);
            }
            if let Some(last) = last_degree {
                if last == degree {
                    // like terms should be combined
                    return Err(CoeffVecError::DuplicateDegree);
                }
            }
            last_degree = Some(degree);
        }
        Ok(())
    }
```

poly/src/polynomial/univariate/sparse.rs
```diff
@@ -230,6 +230,24 @@ impl<F: Field> Zero for SparsePolynomial<F> {
 }
```
z-tech, line 231:
```suggestion

#[derive(Debug)]
pub enum CoeffVecError {
    ZeroCoefficient,
    DuplicateDegree,
}

```

poly/src/polynomial/univariate/sparse.rs
z-tech: Modified this to return Ok / Error. Caller can then do what they like:

```
match is_valid_coefficients_vec(&coeffs) {
    Ok(()) => {},
    Err(err) => {},
}
```

BTW, I think you need to sort if you use this approach to find duplicates?

CHANGELOG.md
```diff
@@ -10,6 +10,8 @@
 
 ### Bugfixes
 
+- [\#915](https://github.com/arkworks-rs/algebra/pull/915) (`ark-poly`) Added `is_valid_coefficients_vec` to `SparsePolynomial::from_coefficients_vec` to remove duplicates and zero coefficients before polynomial creation.
```
z-tech, line 13:
```suggestion
- [\#915](https://github.com/arkworks-rs/algebra/pull/915) (`ark-poly`) Added `is_valid_coefficients_vec` to `SparsePolynomial` to remove duplicates and zero coefficients before polynomial creation.
```

z-tech:
Thanks for this! I made some suggestions. It would be great if you could write some tests?

poly/src/polynomial/univariate/sparse.rs
woojucy: Good point! I will consider this and incorporate the necessary changes accordingly.

woojucy:
Absolutely! I've added three test cases. Could you please review them and let me know your thoughts? 

However, I'm also considering whether to change how the `rand_sparse_poly` function calls `from_coefficients_vec` in the existing test code. One option could be to integrate the validity check within `rand_sparse_poly` without changing its output, and implement a retry mechanism. Alternatively, we could adjust the output based on the `is_valid_coefficients_vec` check and modify how other test codes handle these cases as well. What do you think?

poly/src/polynomial/univariate/sparse.rs
```diff
@@ -343,6 +371,30 @@ mod tests {
...
+    #[test]
+    fn test_invalid_zero_coefficient() {
+        let mut coeffs = vec![(1, Fr::from(0)), (2, Fr::from(3))];
+        let validation = SparsePolynomial::is_valid_coefficients_vec(&mut coeffs);
+        assert!(matches!(validation, Err(CoeffVecError::ZeroCoefficient)));
+    }
+
+    #[test]
+    fn test_invalid_duplicate_degrees() {
+        let mut coeffs = vec![(1, Fr::from(2)), (1, Fr::from(3))];
```
z-tech, line 393:
```suggestion
        let mut coeffs = vec![(3, Fr::from(2)), (1, Fr::from(2)), (2, Fr::from(2)), (1, Fr::from(3))];
```

This way you check if the sorting works as expected.

z-tech:
Thanks for this! Looks good to me.

z-tech:
@woojucy we didn't mean to close. It got closed automatically when the base branch was deleted. Do you mind to reopen and base off master please? Thanks!

woojucy:
Sorry for the late response—I just got to this. I reopened the PR and rebase it onto current master branch. Let me know if there are any additional updates needed. Thanks for your patience!

PR #917, opened 2025-01-08
`DensePolynomial`: add `to_coefficients_vec`

zhiqiangxu:
This PR adds a `to_coefficients_vec` to `DensePolynomial`, so that external users can convert vec to `DensePolynomial` with `from_coefficients_vec`, and after computations finally converts `DensePolynomial` back to vec with `to_coefficients_vec` without copy.


z-tech (commented):
Are these different?

Pratyush:
You can always do `poly.coeffs` to extract the coefficients with a move. Can you point to a codebase where you use this?

zhiqiangxu:
> You can always do `poly.coeffs` to extract the coefficients with a move. Can you point to a codebase where you use this?

Ah correct, I didn't notice it's public. Closing.

PR #946, opened 2025-02-27
Fix: must_use on traits has no effect and causes CI to fail

z-tech:
## Fix: must_use on traits has no effect and causes CI to fail

As of Rust 1.87.0-nightly, #[must_use] on provided trait methods triggers a warning, which becomes an error under cargo check with -D warnings, breaking CI.

Since it had no effect to begin with I think it's safe to remove and we need to do this to resolve CI werrors. 

---

- [N/A ] Linked to GitHub issue with discussion and accepted design OR have an explanation in the PR that describes this work.
- [N/A ] Wrote unit tests
- [N/A ] Updated relevant documentation in the code
- [N/A] Added a relevant changelog entry to the `Pending` section in `CHANGELOG.md`
- [N/A] Re-reviewed `Files changed` in the GitHub PR explorer

### DIFF
--- a/ff/src/fields/models/quadratic_extension.rs
+++ b/ff/src/fields/models/quadratic_extension.rs
@@ -570,7 +570,6 @@ impl<P: QuadExtConfig> From<bool> for QuadExtField<P> {
 impl<P: QuadExtConfig> Neg for QuadExtField<P> {
     type Output = Self;
     #[inline]
-    #[must_use]
     fn neg(mut self) -> Self {
         self.c0.neg_in_place();
         self.c1.neg_in_place();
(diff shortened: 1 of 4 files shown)


ff/src/fields/models/quadratic_extension.rs
```diff
@@ -570,7 +570,6 @@ impl<P: QuadExtConfig> From<bool> for QuadExtField<P> {
 impl<P: QuadExtConfig> Neg for QuadExtField<P> {
     type Output = Self;
     #[inline]
-    #[must_use]
```
Pratyush, line 573: Why do we need to remove it from here as well?

z-tech: Running this from the CI on 1.87.0-nightly, `RUSTFLAGS="-D warnings" cargo check --all-features --examples --workspace --benches` gives:

error: `#[must_use]` has no effect when applied to a provided trait method
   --> ff/src/fields/models/quadratic_extension.rs:573:5
    |
573 |     #[must_use]
    |     ^^^^^^^^^^^

PR #947, opened 2025-02-27
Implement efficient addition formula for affine points

VolodymyrBg:
This PR implements the efficient addition formula for affine points in short Weierstrass curves.

Previously, when adding two affine points, they were first converted to projective coordinates and then the general addition formula was used. This implementation directly computes the sum of two affine points using the specialized formula for z=1, which is more efficient.

The implementation includes:
1. Direct addition formula for affine points
2. Special case handling for infinity and point doubling
3. A new double() method for affine points

This resolves the TODO comment in the affine.rs file.


Pratyush:
This uses an inversion, and will be much slower.

VolodymyrBg:
@Pratyush 
suggested one more variant

VolodymyrBg:
@Pratyush @z-tech @mmagician

Pratyush:
This is functionally the same as the old code.

PR #962, opened 2025-03-27
link Update README.md

romashka-btc:
without .git is correct


Pratyush:
See the contributors guidelines. We don't accept low-effort PRs for things like typos, especially when they don't even fix a typo.

PR #967, opened 2025-04-02
feat: serde-compatible adapters

emberian:
## Description

Adds wrapper types that simplify using CanonicalSerialize/CanonicalDeserialize at the boundaries of `serde` types. serde-encoded-bytes is used to encode either efficient binary data for supported data formats, or else base64 fields when human readable.

The wrapper types derive all of the common traits, as well as implementing CanonicalSerialize/CanonicalDeserialize themselves, with an implementation that ignores the runtime parameters and forces validaation / compression status.

The intention is that crates that use arkworks-using crates can slap one of these wrappers around their fields at an integration interface, or use `serde_with` for semver-compatible porting. I've done a trial integration in one of my projects, and it simplifies things considerably. 

Follow-up on https://github.com/arkworks-rs/algebra/pull/506#issuecomment-1911642960

---

- [ ] Wrote unit tests: no real functionality to test

### DIFF
--- a/serialize/Cargo.toml
+++ b/serialize/Cargo.toml
@@ -33,6 +33,7 @@ serde_with = { workspace = true, optional = true, default-features = false }
 sha2.workspace = true
 sha3.workspace = true
 blake2.workspace = true
+serde_json.workspace = true
 ark-test-curves = { workspace = true, default-features = false, features = [
     "bls12_381_curve",
 ] }
--- a/serialize/Cargo.toml
+++ b/serialize/Cargo.toml
@@ -27,6 +27,7 @@ num-bigint.workspace = true
 rayon = { workspace = true, optional = true }
 serde = { workspace = true, optional = true }
 serde-encoded-bytes = { workspace = true, optional = true, features = ["base64"] }
+serde_with = { workspace = true, optional = true, default-features = false }
 
 [dev-dependencies]
 sha2.workspace = true
@@ -40,6 +41,7 @@ ark-test-curves = { workspace = true, default-features = false, features = [
 [features]
 default = []
 parallel = ["rayon"]
-std = ["ark-std/std"]
+std = ["ark-std/std", "serde_with/std"]
 derive = ["ark-serialize-derive"]
-serde = ["dep:serde", "dep:serde-encoded-bytes"]
+serde = ["dep:serde", "dep:serde-encoded-bytes", "dep:serde_with"]
+serde_with = ["serde", "dep:serde_with"]
--- a/serialize/Cargo.toml
+++ b/serialize/Cargo.toml
@@ -26,6 +26,7 @@ digest.workspace = true
 num-bigint.workspace = true
 rayon = { workspace = true, optional = true }
 serde = { workspace = true, optional = true }
+serde-encoded-bytes = { workspace = true, optional = true, features = ["base64"] }
 
 [dev-dependencies]
 sha2.workspace = true
@@ -41,4 +42,4 @@ default = []
 parallel = ["rayon"]
 std = ["ark-std/std"]
 derive = ["ark-serialize-derive"]
-serde = ["dep:serde"]
+serde = ["dep:serde", "dep:serde-encoded-bytes"]
--- a/serialize/Cargo.toml
+++ b/serialize/Cargo.toml
@@ -41,7 +41,7 @@ ark-test-curves = { workspace = true, default-features = false, features = [
 [features]
 default = []
 parallel = ["rayon"]
-std = ["ark-std/std", "serde_with/std"]
+std = ["ark-std/std", "serde_with?/std"]
 derive = ["ark-serialize-derive"]
 serde = ["dep:serde", "dep:serde-encoded-bytes", "dep:serde_with"]
 serde_with = ["serde", "dep:serde_with"]
--- a/serialize/Cargo.toml
+++ b/serialize/Cargo.toml
@@ -25,6 +25,7 @@ arrayvec.workspace = true
 digest.workspace = true
 num-bigint.workspace = true
 rayon = { workspace = true, optional = true }
+serde = { workspace = true, optional = true }
 
 [dev-dependencies]
 sha2.workspace = true
@@ -40,3 +41,4 @@ default = []
 parallel = ["rayon"]
 std = ["ark-std/std"]
 derive = ["ark-serialize-derive"]
+serde = ["dep:serde"]
--- a/serialize/src/lib.rs
+++ b/serialize/src/lib.rs
@@ -11,11 +11,14 @@
 mod error;
 mod flags;
 mod impls;
+#[path = "serde.rs"]
+mod serde_;
 
 pub use ark_std::io::{Read, Write};
 
 pub use error::*;
 pub use flags::*;
+pub use serde_::*;
 
 #[cfg(test)]
 mod test;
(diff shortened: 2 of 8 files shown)


emberian:
hello @mmaker, i made some progress towards the comment you left, care to take a look?

emberian:
With the additional `From` impl, a usage looks like:

```rs
/// Final Aggregated Signature.
#[derive(Clone, Debug, PartialEq, Eq, CanonicalSerialize, CanonicalDeserialize)]
#[derive(Serialize,Deserialize)]
#[serde(from = "CompressedChecked<Self>")]
#[serde(into = "CompressedChecked<Self>")]
pub struct Signature {
    /// Aggregated BLS signature σ' = H(m)^aSK.
    pub agg_sig_prime: G2,
    /// The threshold for the signature.
    pub threshold: F,
    /// The SNARK proof π.
    pub proof: Proof,
}

impl From<CompressedChecked<Signature>> for Signature {
    fn from(sig: CompressedChecked<Signature>) -> Self {
        sig.0
    }
}
```

emberian:
Here is a more detailed example showing the latest integration with `serde_with`. When the top-level type isn't suitable for `#[serde(with = ...)]`, these conversions can be used instead of pushing the wrapper types into the field values. This prevents the conversion structures from polluting the real datastructure (eg having to add dummy constructors and strip them out of collections), minimizing the impact of adding serde support.

```rs

/// Global data for a hints setup.
///
/// Contains the KZG parameters and the cache of precomputed Lagrange polynomials and commitments.
#[serde_with::serde_as]
#[derive(Clone, Debug, Serialize)]
pub struct GlobalData {
    #[serde_as(as = "CompressedChecked<UniversalParams<Curve>>")]
    pub params: UniversalParams<Curve>,
    pub degree_max: usize,
    /// Used for domain separation in lockstitch
    pub domain: String,
    #[serde_as(as = "Vec<CompressedChecked<DensePolynomial<F>>>")]
    pub(crate) lagrange_polynomials: Vec<DensePolynomial<F>>,
    #[serde(with = "ark_serialize::serde::vec_compressed_checked")]
    pub(crate) lagrange_coms_g1: Vec<G1>,
    #[serde(with = "ark_serialize::serde::vec_compressed_checked")]
    pub(crate) lagrange_coms_g2: Vec<G2>,
    #[serde(skip)]
    pub(crate) lockstitch: lockstitch::Protocol,
}

impl<'de> Deserialize<'de> for GlobalData {
    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
    where
        D: serde::Deserializer<'de>,
    {
        #[serde_with::serde_as]
        #[derive(Deserialize)]
        // keeping the same name makes the generated code more consistent
        struct GlobalData {
            #[serde_as(as = "CompressedChecked<UniversalParams<Curve>>")]
            pub params: UniversalParams<Curve>,
            pub degree_max: usize,
            /// Used for domain separation in lockstitch
            pub domain: String,
            #[serde(with = "ark_serialize::serde::vec_compressed_checked")]
            pub(crate) lagrange_polynomials: Vec<DensePolynomial<F>>,
            #[serde(with = "ark_serialize::serde::vec_compressed_checked")]
            pub(crate) lagrange_coms_g1: Vec<G1>,
            #[serde(with = "ark_serialize::serde::vec_compressed_checked")]
            pub(crate) lagrange_coms_g2: Vec<G2>,
        }
        let inner = GlobalData::deserialize(deserializer)?;
        let lockstitch = lockstitch::Protocol::new(&inner.domain);
        Ok(crate::GlobalData {
            params: inner.params,
            degree_max: inner.degree_max,
            domain: inner.domain,
            lagrange_polynomials: inner.lagrange_polynomials,
            lagrange_coms_g1: inner.lagrange_coms_g1,
            lagrange_coms_g2: inner.lagrange_coms_g2,
            lockstitch,
        })
    }
}
```

serialize/Cargo.toml
```diff
@@ -38,5 +41,7 @@ ark-test-curves = { workspace = true, default-features = false, features = [
 [features]
 default = []
 parallel = ["rayon"]
-std = ["ark-std/std"]
+std = ["ark-std/std", "serde_with/std"]
```
Pratyush, line 44: I'd prefer it not be a default feature, as then every dependent crate would immediately depend on `serde`. 
```suggestion
std = ["ark-std/std"]
```

serialize/src/lib.rs
```diff
@@ -12,10 +12,13 @@ mod error;
 mod flags;
 mod impls;
 
+mod serde;
```
Pratyush, line 15: Or should this be `serde_with`?

```suggestion
#[cfg(feature = "serde")]
mod serde;
```

serialize/src/serde.rs
```diff
@@ -0,0 +1,254 @@
...
+            fn from(value: T) -> $type {
+                $cons(value)
+            }
+        }
+
+        impl<T: Valid> Valid for $type {
+            fn check(&self) -> Result<(), SerializationError> {
+                self.0.check()
+            }
+        }
```
Pratyush, line 50: Probably this should move into `impl_canonical` as it is logically related to `CanonicalDeserialize`

serialize/Cargo.toml
Pratyush: Oh nvm, I see what's going on here. You might need to use the optional feature syntax here.
```suggestion
std = ["ark-std/std", "serde_with?/std"]
```

Pratyush: See here: https://doc.rust-lang.org/cargo/reference/features.html#dependency-features

Pratyush:
Hi @emberian,

Thank you for the PR! This is a great implementation. I just had a couple of requests. Could you add your code examples of how to use `#[serde(from/into)]` and `#[serde_with]` as documentation in the code? Additionally, if you could add small unit tests that do a roundtrip serde serialization/deserialization (e.g., with serde-json or any other serde package), that would be great

serialize/src/lib.rs
emberian: i wonder if it's a good idea to keep the marker types unconditionally available, because they have minor utility for type-driven serialozation mode customization? seems rarely useful or wanted though.

as a zero-cost annotation they could foreseeably be used in public APIs.

but i'm not sure it helps in practice

Pratyush: Hm that's a good idea. Sure, we can keep them available.

Pratyush:
Hey @emberian ! The PR looks good modulo the doc examples.

emberian:
hi @Pratyush i think this is ready for merging :)

I ended up having to `ignore` the new doc example code because it is not possible to enable a crate `feature` just for the docs.

Pratyush (approved):

serialize/Cargo.toml
```diff
@@ -25,11 +25,15 @@ arrayvec.workspace = true
...
 rayon = { workspace = true, optional = true }
+serde = { workspace = true, optional = true }
+serde-encoded-bytes = { workspace = true, optional = true, features = ["base64"] }
+serde_with = { workspace = true, optional = true, default-features = false }
 
 [dev-dependencies]
 sha2.workspace = true
 sha3.workspace = true
 blake2.workspace = true
+serde_json.workspace = true
```
burdges, line 36: This should not be on by default.

Ideally serde should not even be an optinal dependency, which might inject code everywhere, but only a dependency of a seperate crate.

burdges: I suppose okay, maybe it could do so only in CI here?

The concerns over serde is first that it shipped a binary that runs during builds once before, and second that at least one of the maintainers supposedly works for Palantir.

PR #968, opened 2025-04-02
Support 5 mod 8 case in SqrtPrecomputation

serhii023:
## Description

This pull request optimizes the square root method for cases where the given element equals 5 mod 8. This specific case now runs in constant time, significantly improving performance compared to the Tonelli-Shanks algorithm.

### DIFF
--- a/ff/src/biginteger/mod.rs
+++ b/ff/src/biginteger/mod.rs
@@ -157,6 +157,13 @@ impl<const N: usize> BigInt<N> {
         (((self.0[0] << 62) >> 62) % 4) as u8
     }
 
+    #[doc(hidden)]
+    pub const fn mod_8(&self) -> u8 {
+        // To compute n % 8, we need to simply look at the
+        // 3 least significant bits of n, and check their value mod 8.
+        (((self.0[0] << 61) >> 61) % 8) as u8
+    }
+
     /// Compute a right shift of `self`
     /// This is equivalent to a (saturating) division by 2.
     #[doc(hidden)]
--- a/ff/src/fields/models/fp/montgomery_backend.rs
+++ b/ff/src/fields/models/fp/montgomery_backend.rs
@@ -93,6 +93,37 @@ pub trait MontConfig<const N: usize>: 'static + Sync + Send + Sized {
         }
     };
 
+    /// (MODULUS + 3) / 8 when MODULUS % 8 == 5. Used for square root precomputations.
+    #[doc(hidden)]
+    const MODULUS_PLUS_THREE_DIV_EIGHT: Option<BigInt<N>> = {
+        match Self::MODULUS.mod_8() == 5 {
+            true => {
+                let (modulus_plus_three, carry) = Self::MODULUS.const_add_with_carry(&BigInt!("3"));
+                let mut result = modulus_plus_three.divide_by_2_round_down();
+                // Since modulus_plus_one is even, dividing by 2 results in a MSB of 0.
+                // Thus we can set MSB to `carry` to get the correct result of (MODULUS + 1) // 2:
+                result.0[N - 1] |= (carry as u64) << 63;
+                result = result.divide_by_2_round_down();
+
+                Some(result.divide_by_2_round_down())
+            },
+            false => None,
+        }
+    };
+
+    /// (MODULUS - 1) / 4 when MODULUS % 8 == 5. Used for square root precomputations.
+    #[doc(hidden)]
+    const MODULUS_MINUS_ONE_DIV_FOUR: Option<BigInt<N>> = {
+        match Self::MODULUS.mod_8() == 5 {
+            true => {
+                let (modulus_plus_three, borrow) = Self::MODULUS.const_sub_with_borrow(&BigInt::one());
+                let mut result = modulus_plus_three.divide_by_2_round_down();
+                Some(result.divide_by_2_round_down())
+            },
+            false => None,
+        }
+    };
+
     /// Sets `a = a + b`.
     #[inline(always)]
     fn add_assign(a: &mut Fp<MontBackend<Self, N>, N>, b: &Fp<MontBackend<Self, N>, N>) {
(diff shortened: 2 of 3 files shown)


Pratyush:
Thank you for the PR! Could you please add a reference to the algorithm for this case?

serhii023:
The algorithm to find x such that $x^2 = a \pmod p$ for case $p = 8k + 5$, as when $p = 4k + 3$, is as simple and doesn't require hard computations. I find out about this method in university, so I can’t give clear resource, but I can give a direction to prove it works.

Because we have prime number $p$, we can check Legendre symbol of $(\frac{a}{p})$ as $a^{(p - 1) / 2} = a^{4k + 2} \pmod p$. As $a$ is a quadratic residue, we can go further and search for sqrt of $1$, which is $1$ or $(-1)$ $\pmod p$. If $a^{2k + 1} = 1 \pmod p$, then $a^{2k + 2} = a \pmod p$, so we can compute $x$ as $x = \pm a^{k + 1} \pmod p$. In other case, if $a^{2k + 1} = -1 \pmod p$, we can multiply this non-residue with other non-residue (for example $2$, which assuming $p = 5 \pmod 8$, is always non-residue), and receive $a^{2k + 1} 2^{(p - 1) / 2} = (-1) (-1) = 1 \pmod p$, or in other words $a^{2k + 2} 2^{(p - 1) / 2} = a^{2k + 2} 2^{4k + 2} = (\pm a^{k + 1} 2^{2k + 1})^2 = a \pmod p$, so we can compute $x$ as $x = \pm a^{k + 1} 2^{2k + 1} \pmod p$.

serhii023:
So, what do you think? Is there something wrong, or do you just not need the algorithm extension?

ff/src/biginteger/mod.rs
```diff
@@ -157,6 +157,13 @@ impl<const N: usize> BigInt<N> {
         (((self.0[0] << 62) >> 62) % 4) as u8
     }
 
+    #[doc(hidden)]
+    pub const fn mod_8(&self) -> u8 {
```
z-tech, line 161: So, this and the above `mod_4` generalize to any modulo power two and could be rewritten in a general way correct?

https://www.geeksforgeeks.org/compute-modulus-division-by-a-power-of-2-number

ff/src/fields/models/fp/montgomery_backend.rs
```diff
@@ -93,6 +93,37 @@ pub trait MontConfig<const N: usize>: 'static + Sync + Send + Sized {
         }
     };
 
+    /// (MODULUS + 3) / 8 when MODULUS % 8 == 5. Used for square root precomputations.
+    #[doc(hidden)]
+    const MODULUS_PLUS_THREE_DIV_EIGHT: Option<BigInt<N>> = {
```
z-tech, line 98: Very interesting! I too think it would be helpful to reference a source—perhaps from Hacker's Delight, a university lecture, or an ePrint paper—for further context.

ff/src/biginteger/mod.rs
serhii023: Exactly, the value a % (2^d) is equal to a & (2^d - 1) and this evaluation is more efficient, however I wrote implementation similar to the existing code for the sake of simplicity.

ff/src/fields/models/fp/montgomery_backend.rs
serhii023: It is just a partial case of Tonneli-Shanks algorithm https://en.wikipedia.org/wiki/Tonelli%E2%80%93Shanks_algorithm
For the case p=3 mod 4 the solution is straightforward and simple, for the case p=5 mod 8 we need one probing Legendre symbol evaluation and after that multiply by non-residue in appropriate power if needed. The most complicated case is p=1 mod 8 where the number of Legendre symbol evaluations can be different for different values.

z-tech: Hi @serhii023, thanks for the link—this adds some context. And BTW in general, optimizations are great and appreciated.

That being said, a reasonable concern here is that proof by example is not generally sufficient for a cryptographic library (which has a higher standard than general purpose software).

What are your thoughts? Would you be interested in writing a proof for the cases you've referenced?

TBH, maybe a cool side-project in Lean etc.

juja256: Here is a proof:
Let $x^2 \equiv a \pmod p$ is our quadratic equation where we need to find $x$ if one exists. Firstly, note that solutions modullo p exists iff $a$ is quadratic residue modullo p. Recall that $a \in \mathbb{F}_p$ is quadratic residue iff $a^{(p-1)/2} \equiv 1 \pmod p$ by [Euler criterion](https://en.wikipedia.org/wiki/Euler%27s_criterion) or equivalently Legendre symbol $(\frac{a}{p}) = 1$ so that solutions to the equation $x^2 \equiv a \pmod p$ exist.

Consider than partial case where $p$ takes form $p=4k+3$ (case that arkworks already implements):
$x^2 \equiv a \pmod p$, so that $a$ is quadratic residue iff $a^{2k+1} \equiv 1 \pmod p$, multiplying by $a$ both sides of conjugation we obtain $a^{2k+2} \equiv a \pmod p$ so that $a^{2k+2} \equiv x^2 \pmod p$ so we can easily take square root of both sides as every exponent is even: $x = \pm a^{k+1} \pmod p \quad \square$

Consider our case $p=8k+5$ (@serhii023 proposal):
$x^2 \equiv a \pmod p$, so that $a$ is quadratic residue iff $a^{4k+2} \equiv 1 \pmod p$, here we could not simply multiply each side by $a$ as in previous case since the exponent will be odd. But instead since $a^{4k+2} \equiv 1^2 \equiv 1 \pmod p$ taking square root of $1$ gives us either $a^{2k+1} \equiv 1 \pmod p$ or $a^{2k+1} \equiv -1 \pmod p$. Consider each case separately:
- If $a^{2k+1} \equiv 1 \pmod p$ than using the same technique as in $p=4k+3$ case multiplying both sides by $a$ we get $a^{2k+2} \equiv a \equiv x^2 \pmod p$ so that $x = \pm a^{k+1} \pmod p  \quad \square$
- If $a^{2k+1} \equiv -1 \pmod p$ than we could not obtain square root using the same technique, but recalling Tonneli-Shanks trick of multiplying each side by some non-residue we could obtain the square root. Firstly, $2 \in \mathbb{F}_p$ is always quadratic non-residue modullo $p$ as by Legendre symbol [properties](https://en.wikipedia.org/wiki/Legendre_symbol) $(\frac{2}{p})=(-1)^{\frac{p^2 - 1}{8}} = (-1)^{8k^2 +10k + 3} = -1$. So that by Euler criterion: $2^{(p-1)/2} \equiv 2^{4k+2} \equiv -1 \pmod p$. So multiplying both sides of $a^{2k+1} \equiv -1 \pmod p$ by $2^{4k+2}$ we get the following conjugation: $a^{2k+1} 2^{4k+2} \equiv 1 \pmod p$. Multiplying both sides by $a$ we get $a^{2k+2} 2^{4k+2} \equiv a \equiv x^2 \pmod p$ so that $x = \pm a^{k+1} 2^{2k+1} \pmod p \quad \square$

Above proof is not a proof by example, but a general proof for all primes of form $p=8k+5$ and exactly this construction is implemented in @serhii023 PR.

Pratyush:
How does this compare with Atkin's algorithm, as outlined on page 10 here: https://eprint.iacr.org/2012/685.pdf?

juja256:
> How does this compare with Atkin's algorithm, as outlined on page 10 here: https://eprint.iacr.org/2012/685.pdf?

It seems Atkin's algorithm is a constant-time optimization of the straightforward solution proposed by @serhii023

juja256:
@Pratyush @z-tech Hello ! Any ideas this PR could be merged ? There is a strange bug in general library Tonelli-Shanks implementation for the case 8k+5: sometimes it hangs infinitely so merging this PR could fix that issue. You could run `test_field!(fq; Fq; mont_prime_field)` tests for the field:
```rust
use ark_ff::{fields::{Fp256, MontBackend, MontConfig}, BigInteger, PrimeField};

#[derive(MontConfig)]
#[modulus = "7237005577332262213973186563042994240857116359379907606001950938285454250989"]
#[generator = "5"]
pub struct FqConfig;
pub type Fq = Fp256<MontBackend<FqConfig, 4>>;
```

I emphasize that constant-time implementation is not necessary for classic sqrt usages as most of the times it is used for elliptic curve random point generation and secret data does not take part in sqrt computation completely. 
If you want a formal Lean proof of implemented 8k+5 algorithm suggested by this PR you could find it [here](https://github.com/leanprover-community/mathlib4/pull/25739/files)

serhii023:
@Pratyush @z-tech Do you have any updates? Is there something you want us to do for PR?

PR #969, opened 2025-04-03
Remove redundant constraints

elefthei:
## Description

Associated types `ark_ec::pairing::Pairing::G1Prepared ` and `ark_ec::pairing::Pairing::G2Prepared` have the redundant constraints
```
for<'a> From<&'a Self::G1>
for<'a> From<&'a Self::G1Affine>
```
First, these are redundant because `G1` and `G1Affine` (similarly `G2`) implement the Copy trait. Second, we may not care for the benefits of `G1Prepared` and `G2Prepared` at all and we simply want a Pairing instance for `G1` and `G2`. 
A sensible thing to do then is to define

```
type G1 = G1;
type G1Prepared = G1;
```

The redundant constraints forbid this in this case, as `G1Prepared` is more constrained than `G1`. This commit fixes this.

- [ ] Wrote unit tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Pending
 
 - Consolidated logic into `bitreverse_permutation_in_place` and made it public.
+- Remove redundant type constraints from `Pairing::G1Prepared`.
 
 ### Breaking changes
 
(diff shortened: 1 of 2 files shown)


Pratyush:
Hm when do we set `G1Prepared = G1`?

Pratyush (approved):

Issue #855, opened 2024-09-06
SWUConfig for bn254


NikZak:
I wonder if SWUConfig can be implemented for bn254. Only need one param: Zeta (probably can take from Gnark). It is hard to do outside of the repo due to foreign trait on foreign types limitations

nickz-t3:
Oh, I guess I see the issue. It is SW method simplified by Ulas which requires a*b != 0 where are a and b coming from short Weierstrass form of the curve equation. So not applicable to BN254. Then my demand changes. Can we implement the straight SW method is in the standard https://www.ietf.org/archive/id/draft-irtf-cfrg-hash-to-curve-16.html#straightline-svdw

nickz-t3:
I am happy to give it a try if you guys accept contributions. Should not be extremely difficult

NikZak:
@drskalman @ValarDragon @daira @Pratyush @mmagician @weikengchen tagging you as authors/contributors of the pulls [343](https://github.com/arkworks-rs/algebra/pull/343)
[147](https://github.com/arkworks-rs/algebra/pull/147)
I saw that the work started to implement the SW but than was superseded by simplified SWU. 
SWU is great but does not cover BN254. Meanwhile BN254 is part of arkworks. So it makes arkworks incomplete compared to say https://github.com/ConsenSys/gnark-crypto/blob/master/ecc/bn254/hash_to_g1.go. And as I mentioned earlier implementing this outside of the repo requires not so beatiful wrappers around bn254 to avoid foreign trait implementation limitations.
So is there a reason why this feature should not be added to the repo?

Issue #815, opened 2024-03-29
Support mapping to all Weierstrass curves (Shallue-van de Woestijne method)


slumber:
Only a handful of arkworks SW curves implement mapping, following https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-hash-to-curve-09 it's done for bls curves.

Suggestions:

1. Introduce non-simplified SWU from [here](https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-hash-to-curve-09#name-shallue-van-de-woestijne-me). There're existing implementations in Rust, and sage script for finding constants. Add a note for users
> For curves supported by the Simplified SWU method ([Section 6.6.2](https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-hash-to-curve-09#simple-swu)), that mapping is the recommended one. Otherwise, the Simplified SWU method for AB == 0 ([Section 6.6.3](https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-hash-to-curve-09#simple-swu-AB0)) is recommended if the goal is best performance, while the Shallue-van de Woestijne method ([Section 6.6.1](https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-hash-to-curve-09#svdw)) is recommended if the goal is simplicity of implementation. (The reason for this distinction is that the Simplified SWU method for AB == 0 requires implementing an isogeny map in addition to the mapping function, while the Shallue-van de Woestijne method does not.)The Shallue-van de Woestijne method ([Section 6.6.1](https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-hash-to-curve-09#svdw)) works with any curve, and may be used in cases where a generic mapping is required. Note, however, that this mapping is almost always more computationally expensive than the curve-specific recommendations above.
2. Implement new trait for popular curves like BN254, ideally for all of them.
3. Implement simplified SWU for the pasta cycle, as the isogeny for it exists https://github.com/zcash/pasta_curves/blob/main/src/hashtocurve.rs

Problems:
1. Name clash of SWU with currently available (simplified) SWU. Other projects call non-simplified version [SVDW](https://github.com/privacy-scaling-explorations/halo2curves/blob/1662f881cb1a19719f3a25edd373c15e0aff3274/src/grumpkin/curve.rs#L39). Alternatively, SWU can be renamed to SSWU. This is less favorable because existing code will seamlessly start using computationally expensive version.
[...]

PR #972, opened 2025-04-08
feat: bn254 maptocurve

emberian:
## Description

allows bn254 to support hashing to curve.

the RFC does not have test vectors for bn254,
so the tests are somewhat impoverished, but
the functionality does work according to tests.

closes: #855

some progress towards #815, however the present implementation does not support "_all_" curves. I am not sure the trait setup is amenable to that. this PR makes it easy to extend support to any _particular_ curve, however. I do not think this closes that issue.


emberian:
@Pratyush can you help me move this forward? I see @burdges and @mmagician in previous PR for elligator.

PR #981, opened 2025-05-04
fix: remove unnecessary ?Sized constraints in Valid implementations

GarmashAlex:
## Description

This commit fixes compiler warnings related to the use of the ?Sized constraint in the Valid trait and CanonicalDeserialize implementations.

The problem was that trait Valid has an explicit Sized constraint in its definition:
`pub trait Valid: Sized + Sync { .... }`

However, some implementations have used the ?Sized constraint, which makes no sense, 
since Valid already requires Sized. These constraints have been removed in the following places:

1. impl<T: ?Sized + Valid + Sync + Send> Valid for ark_std::sync::Arc<T>
2. impl<T: ?Sized + CanonicalDeserialize + ToOwned + Sync + Send> CanonicalDeserialize for ark_std::sync::Arc<T>
3. impl where T: ?Sized + ToOwned + Sync + Valid + Send for Cow<'_, T>
4. impl where T: ?Sized + ToOwned + Valid + Sync + Send for CanonicalDeserialize

This change does not affect functionality, since types that implement the Valid trait already 
must be Sized due to a limitation in the trait definition.

---

write a little note why.

- [ ] Wrote unit tests

### DIFF
--- a/serialize/src/impls.rs
+++ b/serialize/src/impls.rs
@@ -371,7 +371,7 @@ impl<T: ?Sized + CanonicalSerialize + ToOwned> CanonicalSerialize for ark_std::s
 }
 
 #[cfg(target_has_atomic = "ptr")]
-impl<T: ?Sized + Valid + Sync + Send> Valid for ark_std::sync::Arc<T> {
+impl<T: Valid + Sync + Send> Valid for ark_std::sync::Arc<T> {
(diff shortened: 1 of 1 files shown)


weikengchen:
I do not think that we will make this change---we really want to remove the Sized requirement of Valid.

Issue #490, opened 2022-09-29
Implement double-odd curves


Pratyush:
See [doubleodd.group](https://doubleodd.group/)

PR #986, opened 2025-05-19
ec: implement double-odd curves

rubdos:
Implements Thomas Pornin's “A Prime-Order Group with Complete Formulas from Even-Order Elliptic Curves” [1], often referred to as "double odd" curves.

Includes the double-odd curve “JQ255s” as presented in the paper.

[1] T. Pornin, “A Prime-Order Group with Complete Formulas from Even-Order Elliptic Curves,” IACR CiC, vol. 1, no. 1, p. 33, Apr. 2024, doi: 10.62056/akmp-4c2h.

This is work by @Tarinn

## Description

closes: #490

---

write a little note why.

- [ ] Wrote unit tests


rubdos:
As a side note: I recall a discussion on Discord where there was a parameter dump for a DO curve that embeds Curve25519, but I don't have the link at hand. Here are the SW parameters:

```
a = 57259562424833564022627223828891588493930540394420821089419290001039534118212
b = 64131260960803756922173133202372554928955727810351288169387335111902918457126
|F_q| = p = 115792089237316195423570985008687907853682756971699735333147980285963064639179
```

ec/src/models/double_odd/affine.rs
```diff
@@ -0,0 +1,321 @@
...
+        })
+    }
+
+    fn mul_bigint(&self, by: impl AsRef<[u64]>) -> Self::Group {
+        P::mul_affine(self, by.as_ref())
+    }
+
+    /// Multiplies this element by the cofactor and output the
+    /// resulting projective element.
+    #[must_use]
```
Pratyush, line 167: This doesn't need to be `must_use` since `Self::Group` is `must_use`.

ec/src/models/double_odd/group.rs
```diff
@@ -0,0 +1,451 @@
...
+
+    const ZERO: Self = Self::new_unchecked(
+        P::BaseField::ONE,
+        P::BaseField::ONE,
+        P::BaseField::ZERO,
+        P::BaseField::ZERO,
+    );
+
+    fn double_in_place(&mut self) -> &mut Self {
+        self.z = -P::get_c().double() * self.t.square(); // Self.z == -2cT^2
```
Pratyush, line 172: Can you add a link/reference to the algorithm being implemented here?

Pratyush (changes requested):
Thanks for the PR! Left some comments!

rubdos:
Pushed a few fixups, will discuss the rest with @Tarinn when we both have time at the office! Thanks for the review so far.

curves/jq255s/src/curves/tests.rs
```diff
@@ -0,0 +1 @@
+
```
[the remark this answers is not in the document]
rubdos: We moved the tests from `lib.rs` over to here

curves/jq255s/src/fields/tests.rs
```diff
@@ -0,0 +1 @@
+
```
Pratyush, line 1: Probably a good idea to either delete this, or add tests like the other curve crates

curves/jq255s/src/lib.rs
```diff
@@ -0,0 +1,32 @@
+#![deny(
+    warnings,
+    unused,
+    future_incompatible,
+    nonstandard_style,
+    rust_2018_idioms
+)]
```
Pratyush, line 7: I think we need to add a `no_std` check like the other crates.

rubdos:
@Pratyush, @samvandevelde noticed that the current implementation was tailored to Jq255s, where A=-1. This has meanwhile been resolved (766b91a). He'll also take care of your further comments. Thanks for the feedback again, as always!

curves/jq255s/src/lib.rs
samvandevelde: Dear Prof. Mishra,

This check has been added in the exact same fashion as for the other curves (commit [4dbeaf1])

Kind regards,
Sam

curves/jq255s/src/constraints/curves.rs
```diff
@@ -0,0 +1,5 @@
+use crate::{constraints::FqVar, *};
+use ark_r1cs_std::groups::curves::short_weierstrass::ProjectiveVar;
+
+/// A group element in the Curve25519 curve.
```
Pratyush, line 4: Please update this comment.

curves/jq255s/src/curves/tests.rs
```diff
@@ -0,0 +1,325 @@
...
+    let a_compressed = Affine::deserialize_compressed(&*compressed_bytes).unwrap();
+    let a_uncompressed = Affine::deserialize_uncompressed(&*uncompressed_bytes).unwrap();
+
+    assert_eq!(a_compressed, a);
+    assert_eq!(a_uncompressed, a);
+}
+
+#[test]
+fn valid_decode() {
```
Pratyush, line 27: Is there a source for these test vectors? If so could you please add a link here?

ec/src/models/double_odd/affine.rs
```diff
@@ -0,0 +1,346 @@
...
+            u: P::BaseField::ZERO,
+        }
+    }
+
+    pub fn n() -> Self {
+        Self {
+            e: -P::BaseField::ONE,
+            u: P::BaseField::ZERO,
+        }
+    }
```
Pratyush, line 77: What is the point represented by this function?

ec/src/models/double_odd/affine.rs
```diff
@@ -0,0 +1,346 @@
...
+    pub fn get_es_from_u_unchecked(u: P::BaseField) -> Option<(P::BaseField, P::BaseField)> {
+        // Compute the curve equation x(x^2 + Ax + B).
+        let e = (P::get_c() * u.square().square() - (P::COEFF_A * u.square()).double()
+            + P::BaseField::ONE)
+            .sqrt()?;
+        let neg_e = -e;
+        match e < neg_e {
+            true => Some((e, neg_e)),
+            false => Some((neg_e, e)),
+        }
```
Pratyush, line 99: How is this value for `e` derived? It doesn't seem to match the curve equation?

ec/src/models/double_odd/affine.rs
```diff
@@ -0,0 +1,346 @@
...
+    }
+
+    /// Checks if `self` is a valid point on the curve.
+    pub fn is_on_curve(&self) -> bool {
+        let e_squared = P::get_c() * self.u.square().square()
+            - (P::COEFF_A * self.u.square()).double()
+            + P::BaseField::ONE;
```
Pratyush, line 106: Same here.

ec/src/models/double_odd/group.rs
```diff
@@ -0,0 +1,461 @@
...
+
+    #[inline]
+    fn neg(mut self) -> Self {
+        self.u = -self.u;
+        self
+    }
+}
+
+impl<P: DOCurveConfig, T: Borrow<Affine<P>>> AddAssign<T> for Projective<P> {
+    fn add_assign(&mut self, other: T) {
```
Pratyush, line 251: Can you add a reference for this code? Perhaps a pointer to existing "projective + projective" formulae, along with a description of the simplification when one of the points is affine.

ec/src/models/double_odd/group.rs
```diff
@@ -0,0 +1,461 @@
...
+        let c = P::get_c();
+        let cn4 = c * n4;
+        self.z = n2 - cn4;
+
+        let n3d = n3.double();
+
+        self.e = (n2 + cn4) * (n1 - P::COEFF_A * n3d) + c * n3d * n5;
+        self.u = self.z * self.t;
+        self.z.square_in_place();
+        self.t.square_in_place();
```
Pratyush, line 336: Same here; please add a reference.

ec/src/models/double_odd/mod.rs
```diff
@@ -0,0 +1,133 @@
...
+                } else {
+                    item.u
+                }
+                .serialize_uncompressed(writer)
+            },
+            Compress::No => {
+                item.e.serialize_with_mode(&mut writer, compress)?;
+                item.u.serialize_with_mode(&mut writer, compress)
+            },
+        }
```
Pratyush, line 85: Could you provide a reference as to how this matches up with the official encoding described in Section 6.1.1 of https://doubleodd.group/doubleodd.pdf?

ec/src/models/double_odd/mod.rs
```diff
@@ -0,0 +1,133 @@
...
+    ) -> Result<Affine<Self>, SerializationError> {
+        let (e, u) = match compress {
+            Compress::Yes => {
+                let u: Self::BaseField = CanonicalDeserialize::deserialize_uncompressed(reader)?;
+                let e: Self::BaseField = (Self::get_c() * u.square().square()
+                    - (Self::COEFF_A * u.square()).double()
+                    + Self::BaseField::ONE)
+                    .sqrt()
+                    .ok_or(SerializationError::InvalidData)
+                    .unwrap();
```
Pratyush, line 101: Could this be deduplicated by calling the `get_e_from_u_unchecked` fn on `Affine`?

ec/src/models/double_odd/mod.rs
samvandevelde: Dear Prof. Mishra,

The referenced official encoding was for do255e and do255s only (and rephrased in the follow-up paper for jq255e/s, **https://doubleodd.group/doubleodd-jq.pdf** page 23). However, was meant to standardize encoding for these specific curves, so it seemed too specific to add in the comments of the model for all double odd curves.

That being said, I compared the specifications with the arkworks implementation of these operations for Fp,
and Fp matches up with the described encoding (see below).

Should these results perhaps be writen in the comments of algebra/curves/jq255s instead?

Kind regards,
Sam

---------------------------------------------------------------------------------------------------------------------------------------------

1) A field element x ∈ Fq is considered as an integer in the 0 to q − 1 range, which is then encoded using the unsigned little-endian convention (least significant byte comes first).
> OK: Serialization of Fp uses copy_from_u64_slice(), which uses from_le_bytes.
2) Encoding always uses 32 bytes, even if the integer could have fit into fewer bytes
> OK: Without any flags, 32 bytes are used for the q and r described in the paper.
3) Decoders MUST reject the source bytes, and return Invalid, if any of the following conditions hold:
• The source length is not exactly 32 bytes.
• The integer resulting from unsigned little-endian interpretation of the source bytes is not in the 0 to q − 1 range.
In particular, there is no ignored bit: even though the value of q implies that the most significant bit of the last byte is always zero, that bit MUST NOT be ignored by decoders. Moreover, checking that the most significant bit of the last byte is zero is not sufficient validation; for instance, an encoding of the integer q itself, over 32 bytes, MUST be rejected as Invalid, since it is not in the 0 to q − 1 range. There is no implicit reduction modulo q.
> OK: Deserialization to Fp uses from_bigint(), which returns an Option depending on is_geq_modulus(), as well as from_le_bytes(), over the entirety of the 32 bytes (fixed to 32 by read_exact_up_to(...,output_byte_size))

ec/src/models/double_odd/affine.rs
samvandevelde: Dear Prof. Mishra,

More explanation about N has been provided in commit [572b797](https://github.com/arkworks-rs/algebra/pull/986/commits/572b797caca66c08b4f8c464fd20e957199d651c)

Kind regards,
Sam

ec/src/models/double_odd/affine.rs
samvandevelde: Dear Prof. Mishra,
The correct (Jacobi Quartic) equation has been provided in commit [572b797](https://github.com/arkworks-rs/algebra/pull/986/commits/572b797caca66c08b4f8c464fd20e957199d651c)

Kind regards,
Sam

ec/src/models/double_odd/group.rs
samvandevelde: Dear Prof. Mishra,
The missing reference to the addition formulae has been provided in commit [572b797](https://github.com/arkworks-rs/algebra/pull/986/commits/572b797caca66c08b4f8c464fd20e957199d651c)

Kind regards,
Sam

ec/src/models/double_odd/affine.rs
```diff
@@ -69,6 +76,10 @@ impl<P: DOCurveConfig> Affine<P> {
         }
     }
 
+    /// Returns N as defined in the original double-odd paper (<https://doubleodd.group/doubleodd.pdf>):
+    /// N is the only point of order 2, so that the representants of each group element are {P; P+N},
+    /// with P being a point on the r-torsion of the curve, and P+N not on said r-torsion..
+    /// In (x,y)-coordinates, N is (0,0).
     pub fn n() -> Self {
```
Pratyush, line 83: hm the invariant that `Affine` is supposed to preserve (and largely does for the other curves) is that it represents points in the prime-order group, and not the point of order 2. I would recommend making this `pub(crate)` if this is used only internally, and adding some discussion of why for this choice of curve it's okay to construct an `Affine` of this order (most likely because the addition and doubling formulae are complete for every curve point, not just the prime-order ones)

ec/src/models/double_odd/mod.rs
Pratyush: Great, thank you! Do the test-vector tests cover this kind of equivalence check?

samvandevelde: They do! I added a reference to the standard above the test valid_decode() which covers the en/decryption

ec/src/models/double_odd/affine.rs
samvandevelde: Due to N and O being both representants of the identity, and the code only requiring one, I have rectified this issue by simply removing the unnecessary function. I did however add in the documentation references to the existence of the other representant for the identity, as to not confuse the reader for certain functions where its existence is important.

samvandevelde:
> Thank you for the follow-up commits! I only have one remaining comment; otherwise this looks good. Do you folks have any follow-up questions from your end?

Dear Professor,

I do in fact have a follow-up question with how MSM's have developed thus far in the Arkworks framework, and how MSM's can be improved for double-odd. However, this would open issues on all of the Arkworks repos, so it is no small issue I raise, and the performance benefits are mild.

The problem is that MSM's have been inherently linked to the Affine representation. Although the trait `ScalarMul` has a separate `MulBase` type by design, the `CurveGroup' is defined as:
pub trait CurveGroup:
    PrimeGroup
    + Add<Self::Affine, Output = Self>
    + [...]
    + VariableBaseMSM
    + ScalarMul<MulBase = Self::Affine>
    + [...]

This makes sense for Short Weierstrass and Twisted Edwards, because there are specialized addition formulas for affine addition so that normalizing in batch and then computing the MSM is more efficient.

However, Double-Odd lacks such benefits: there are not really formulas benefitting from affine addition, which is not what DO-curves were designed for anyway. However, this means that the normalization to affine (removing the additional information in the projective coords) to then perform MSM (and having to recalculate the T-coordinate specifically) is more costly.

I already tinkered locally to decouple affine from MulBase in the CurveGroup trait. This requires changes in the following:
-ark-algebra
-ark-crypto-primitives
-presumably all the proof system crates (I only checked with Spartan)
Depending on the double-odd curve I used, I got 5%-10% performance gains when benching the indexation and verification processes in Spartan using this method, but a smaller difference for proving. However, I only checked Spartan and it is safe to assume that, the more a proof system requires on MSMs, the heavier the impact will be.

Do I open up an issue for this manner?

Kind regards,
Sam

PR #1031, opened 2025-09-14
Refactor MSM to reduce code duplication

waamm:
## Description

- **Consolidated duplicate functions:** The previously separate functions `msm_u8`, `msm_u16`, `msm_32`, and `msm_u64` have been replaced with a single generic function `msm_small_int`.
- Edited the comment above `msm_binary` regarding its range.
- Reduced the window size in `msm_serial` (it doesn't seem to have performance implications).

---

write a little note why.

- [ ] Wrote unit tests (not needed for small refactor)
- [ ] Updated relevant documentation in the code (not needed for small refactor)
- [ ] Added a relevant changelog entry to the `Pending` section in `CHANGELOG.md` (not needed for small refactor)

### DIFF
--- a/ec/src/scalar_mul/variable_base/mod.rs
+++ b/ec/src/scalar_mul/variable_base/mod.rs
@@ -93,25 +93,25 @@ pub trait VariableBaseMSM: ScalarMul + for<'a> AddAssign<&'a Self::Bucket> {
     /// Performs multi-scalar multiplication when the scalars are known to be `u8`-sized.
     /// The default implementation is faster than [`Self::msm_bigint`].
     fn msm_u8(bases: &[Self::MulBase], scalars: &[u8]) -> Self {
-        msm_u8(bases, scalars)
+        msm_small_int_generic(bases, scalars)
     }
 
     /// Performs multi-scalar multiplication when the scalars are known to be `u16`-sized.
     /// The default implementation is faster than [`Self::msm_bigint`].
     fn msm_u16(bases: &[Self::MulBase], scalars: &[u16]) -> Self {
-        msm_u16(bases, scalars)
+        msm_small_int_generic(bases, scalars)
     }
 
     /// Performs multi-scalar multiplication when the scalars are known to be `u32`-sized.
     /// The default implementation is faster than [`Self::msm_bigint`].
     fn msm_u32(bases: &[Self::MulBase], scalars: &[u32]) -> Self {
-        msm_u32(bases, scalars)
+        msm_small_int_generic(bases, scalars)
     }
 
     /// Performs multi-scalar multiplication when the scalars are known to be `u64`-sized.
     /// The default implementation is faster than [`Self::msm_bigint`].
     fn msm_u64(bases: &[Self::MulBase], scalars: &[u64]) -> Self {
-        msm_u64(bases, scalars)
+        msm_small_int_generic(bases, scalars)
     }
 
     /// Streaming multi-scalar multiplication algorithm with hard-coded chunk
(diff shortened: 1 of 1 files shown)


ec/src/scalar_mul/variable_base/mod.rs
```diff
@@ -389,47 +389,20 @@ fn msm_binary<V: VariableBaseMSM>(mut bases: &[V::MulBase], mut scalars: &[bool]
         .sum()
 }
 
-fn msm_u8<V: VariableBaseMSM>(mut bases: &[V::MulBase], mut scalars: &[u8]) -> V {
+/// Generic MSM for small integer scalars.
+fn msm_small_int_generic<V, S>(mut bases: &[V::MulBase], mut scalars: &[S]) -> V
```
burdges, line 393: Why not call this one `msm_u64`?  No point adding the word generic unless you're doing something strange.

burdges: I've no real opinion aside from `generic` seemed redundent.

waamm: Ok I've renamed it (and replaced the invocations inside `msm_signed`). In hindsight, I'm not entirely sure this PR is timely as the code feels quite WIP: when I try a recent arkworks commit in my codebase instead of version 0.5.0, some functions using MSMs run a bit faster while others are much slower...

Pratyush: Thanks! I'd definitely be interested in figuring out why you're seeing a slowdown...

Is there a benchmark I can run?

waamm: Did you try running the benchmark in the first link? (So navigate to `aptos-dkg` and run `cargo bench --bench range_proof`.) 

Meanwhile I'll write some code to split the verification function into smaller pieces...

Pratyush: oh oops I missed the first link, only saw the link to the direct code. Thanks!

waamm: > I believe the bottleneck ought to be the 3 MSMs

Oops this is false, these MSMs are relatively small and the bottleneck here is the two multi-pairings. Going to switch to a recent commit now and check the slowdown...

waamm: With `env RAYON_NUM_THREADS=2` the difference is still there, but smaller (this is comparing 0.5.0 to commit [7601bf9](https://github.com/arkworks-rs/algebra/commit/7601bf924560c8e0963c6cde9a065cf8da7595ca):

bench

> We switched to computing the multi-pairings in parallel when there are more than 4 pairings, but maybe we need to increase the parallelization threshold.

The second multi-pairing is only for 2 pairings and it still became slower by about 20%-25%.

Pratyush: Hm do you have a sense of how large these MSMs are? It's a bit difficult to tell from the code.

Pratyush: Hm actually maybe the difference lies elsewhere; 0.5.0 also parallelizes the miller loop...

waamm: By default `benches/range_proof.rs` sets the length `"ELL"` of these MSMs to 48, but this can be changed at invocation.

For the [first](https://github.com/aptos-labs/aptos-core/blob/6c9c453b68393bcaba3f246aaed445ad1e648c60/crates/aptos-dkg/src/range_proofs/dekart_univariate.rs#L573) MSM, the scalars are powers of two, so $$1, 2, ..., 2^{\ell - 1}$$. For the [other](https://github.com/aptos-labs/aptos-core/blob/6c9c453b68393bcaba3f246aaed445ad1e648c60/crates/aptos-dkg/src/range_proofs/dekart_univariate.rs#L606) [two](https://github.com/aptos-labs/aptos-core/blob/6c9c453b68393bcaba3f246aaed445ad1e648c60/crates/aptos-dkg/src/range_proofs/dekart_univariate.rs#L609) MSMs they are obtained from random 128-bit values.

burdges: As I understand the layers in rustc, this change should not hurt performance, but maybe they have some SIMD logic occuring at a different layer somehow?  It's maybe worth reporting this as a rustc bug?

waamm: Could you clarify which change you’re referring to? (The changes in this PR shouldn’t affect performance.)

PR #1032, opened 2025-09-17
Skip unrolling labeled for-loops to preserve labeled control-flow semantics

Forostovec:
This change avoids unrolling any for-loop that has a label. In the previous implementation, the original loop label was moved onto an enclosing Expr::Block. That preserved break 'label but broke or changed the semantics of continue 'label, which is only valid for loops, not blocks. By skipping unrolling for labeled loops, we guarantee that both break 'label and continue 'label retain their original behavior. This is a safe, minimal fix that prevents subtle control-flow bugs without affecting unlabeled loops.

### DIFF
--- a/ff-macros/src/unroll.rs
+++ b/ff-macros/src/unroll.rs
@@ -101,6 +101,12 @@ fn unroll(expr: &Expr, unroll_by: usize) -> Expr {
             })
         };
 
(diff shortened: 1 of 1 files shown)


Pratyush:
Does this actually affect any code that you are writing? Because it doesn't affect anything in arkworks.

PR #1033, opened 2025-09-21
fix: remove redundant take() calls on iterators

DeVikingMark:
Remove unnecessary take(size) and take(coset_size) calls where vectors are already created with exact required size. 

Improves performance by eliminating redundant bounds checking.


Pratyush:
Can you provide some benchmarks which show that the code improves performance?

DeVikingMark:
master 
image
my branch
image

@Pratyush 
oh. seems to be my implementation doesn't make any sense

DeVikingMark:
but i guess my bench file could be wrong therefore if it seems appropriate for you to close my pr, than in would be the best decision, i'll probable find another option to contribute

PR #1051, opened 2025-10-15
Temp Pr1044 SmallFp Trait

benbencik:
This PR is only temporary, and it is related to https://github.com/arkworks-rs/algebra/pull/1044. Here we selected implementation in crate `ff` that defines the small field `SmallFp`. The trait lives in `ff/src/fields/models/small_fp/small_fp_backend.rs`.

The trait was copied from `FpConfig` located in `ff/src/fields/models/fp/mod.rs` and iteratively changed to support field elements of native integer types instead of `BigInts`. We decided not to use `MontConfig` because we want to have support for the standard backend as well as the Montgomery backend.


z-tech:

PR #1053, opened 2025-10-15
 Temp Pr1044 SmallFp Tests

benbencik:
This PR is only temporary, and it is related to https://github.com/arkworks-rs/algebra/pull/1044. Here, we add test curves as well as extend the test suite to run CI on the implementation of `SmallFp`. Moreover, we add benchmark `test-curves/benches/smallfp.rs` which uses existing bench templates for fields.


z-tech:

PR #1054, opened 2025-10-15
Temp Pr1044 SmallFp Macro

benbencik:
This PR is only temporary, and it is related to https://github.com/arkworks-rs/algebra/pull/1044. Here, we selected implementation in crate `ff-macros` that defines a proc macro to implement `SmallFp`. The macro lives in `ff-macros/src/small_fp/mod.rs`. We define implementation for the arithmetic operations with the Montgomery backend as well as the standard backend.


z-tech:

PR #1061, opened 2025-11-12
perf(poly): reduce allocations in DenseMultilinearExtension concat, fix_variables, and evaluate

Fibonacci747:
[concat] Use from_evaluations_vec to move the built Vec<F> instead of cloning via from_evaluations_slice, eliminating one full-buffer copy. 
[fix_variables] After folding, truncate to 1 << (nv - dim) and return with from_evaluations_vec to avoid cloning the sliced prefix.
[evaluate] Compute the scalar directly by folding over a temporary buffer (cloned once from evaluations) and return a[0], avoiding constructing a new DenseMultilinearExtension


weikengchen:
👋 Hi @Fibonacci747 — this is **Claude** (Anthropic's AI assistant), working with @weikengchen on the review. I pushed a follow-up commit (`34936408`) on top of your branch and want to explain the reasoning and share benchmark numbers, since it changes the approach for two of the three functions.

First off — thanks for this PR. The core idea (cut redundant allocations on the `DenseMultilinearExtension` hot paths) is the right instinct, and the `concat` change in particular is a clear win that I kept as-is.

## What the follow-up commit does

- **`concat`** — kept your `from_evaluations_vec` move (just added an explanatory comment). ✅
- **`fix_variables`** — switched from *"clone the full table → fold → truncate"* to a **half-buffer fold**: round 1 reads `self` directly into a `2^(nv-1)` buffer, so the full `2^nv` table is never cloned at all.
- **`evaluate`** — reverted to the original one-line `self.fix_variables(point)[0]` delegation instead of the inlined fold.

## Why — the benchmarks

I benchmarked the original (`master`), your version, and the half-buffer variant **side-by-side in one harness** (criterion, BLS12-381 `Fr`). Run-to-run noise is ~±5%, so the figures below are within-run medians, and I only drew conclusions from deltas that reproduced across 3 runs.

**`concat`** (two `nv`-var MLEs → `2^(nv+1)` table) — your change, kept:

| nv | `master` (slice-clone) | this PR (move) | speedup |
|----|------|------|---------|
| 12 | 9.5 µs | 4.6 µs | **2.06×** |
| 16 | 152 µs | 68 µs | **2.2×** |
| 20 | 3.25 ms | 1.26 ms | **2.6×** |

It's memcpy-bound, so dropping the clone ~halves the work — and also halves *transient* peak memory (no second full buffer). 👍

**`fix_variables`** (bind 1 of `nv` vars — the common sumcheck case):

| nv | `master` | this PR (truncate+move) | half-buffer |
|----|------|------|------|
| 12 | 56 µs | 55 µs *(≈ noise)* | **51.6 µs** |
| 16 | 1.07 ms | 1.11 ms *(≈ noise)* | **1.01 ms** |
| 20 | 17.9 ms | 17.8 ms *(≈ noise)* | **16.5 ms** |

Your `truncate`+move is correct and removes one allocation, but the wall-time came out *within noise*: it still clones the full `2^nv` table, and the dominant cost is the field-arithmetic fold, not the avoided copy. The half-buffer version never materializes the full table — it folds round 1 straight from `self` into a half-size buffer — which is reliably **~5–8% faster** and, more importantly, **halves peak memory (`2^(nv-1)` vs `2^nv`)**.

**`evaluate`** (full point):

| nv | `master` (delegation) | this PR (inline) | half-buffer (delegation) |
|----|------|------|------|
| 12 | 109 µs | 110 µs *(≈ noise)* | **102 µs** |
| 16 | 2.12 ms | 2.09 ms *(≈ noise)* | **2.01 ms** |
| 20 | 33.7 ms | 33.8 ms *(≈ noise)* | **32.5 ms** |

The inlined fold measured *identical* to the original `self.fix_variables(point)[0]` delegation (within noise — the only thing it avoided was a single 1-element allocation), while duplicating the fold loop. Since the half-buffer `fix_variables` already captures the gain, `evaluate` can stay a one-liner and inherit the speedup **and** the 2× peak-memory reduction with **zero duplication** — which is why I reverted it.

## Correctness & checks

- Verified **bit-identical output** to the original `master` implementation across `nv = 0..=12` and *every* partial-point length `0..=nv` (random `Fr` inputs).
- Full `ark-poly` suite passes: **113 unit tests + 11 doctests** (the doctests exercise the `concat`/`fix_variables`/`evaluate` examples), and `cargo fmt --check` is clean.
- Added the missing CHANGELOG entry under `## Pending`.

Net effect vs. `master`: `concat` ~2–2.6× faster, `fix_variables`/`evaluate` ~5–8% faster with half the peak allocation, and no duplicated logic.

Totally open to adjusting the split if you'd prefer a different shape — and thanks again for kicking this off. 🙏

---
*— Claude (Anthropic), pushed via @weikengchen's account. The commit and analysis above are mine; happy to iterate.*

weikengchen (approved):

PR #1072, opened 2025-12-17
perf: optimize Projective + Bucket addition with explicit formula

PivasDesant:
Replaced the TODO that was converting Bucket to Projective before adding.

Now we compute the addition directly using Bucket's zz/zzz coordinates, which should be faster since we skip the conversion step.

Also fixed a bug in the U2 calculation - was missing the z2z2 multiplication.


Pratyush:
Thanks for the PR! The CI seems to be failing. It would also be nice to have a reference to how the formula was derived. Thanks!

PR #1074, opened 2025-12-28
perf: optimize batch_check for BTreeMap and HashMap

avorylli:
Replace inefficient unzip() approach with direct flat_map iteration in batch_check methods. 

This eliminates intermediate Vec allocations for iterator pairs and aligns the implementation with the pattern used for sequence types.


Pratyush:
This does not reduce the allocated amount.

PR #1079, opened 2026-03-07
Work around for wasm_bindgen::JsError lacking no_std support

burdges:
I think wasm_bindgen should fix this error, and if they did then this PR would break, so I'm not sure this PR should be merged, but this PR provides a work around for now, which can be patched in, so maybe worth having in a branch here.

https://github.com/wasm-bindgen/wasm-bindgen/issues/5025

---

write a little note why.

- [ ] Wrote unit tests


weikengchen:
Look at the chat history, I think it depends on wasm-bindgen MSRV --- and given that the next release of arkworks-rs probably would be later than that (and arkworks might also bump MSRV), let us wait for that :)

PR #1086, opened 2026-03-29
In define_field macro set plausibly useful FFT params for user

### LINKED ISSUE #1087: define_field! macro sets subgroup base and power to none for Fp and SmallFp fields
# Description

The idea here is that define_field! is this little bit higher level function that sets config for the user such as selecting the correct backend + representation and I think selecting reasonable FFT params is also nice.

We search for a small subgroup base in [3,5,7] and set to None otherwise.

The user is still totally free to select their own params using the traditional macros for both Fp and SmallFp btw.

z-tech:
When using the define_field! macro, we were setting FFT params to None irrespective of when nice params do exist. This PR searches for nice seeming params at compile time and sets them None otherwise.

I think this is cool because define_macro! is supposed to be a little bit hand holding for the user.

The traditional macros for both Fp and SmallFp are untouched the user can customize there however they want.

closes: #1087

---

- [ x] Targeted PR against correct branch (master)
- [x ] Linked to GitHub issue with discussion and accepted design OR have an explanation in the PR that describes this work.
- [ ] Wrote unit tests
- [ x] Re-reviewed `Files changed` in the GitHub PR explorer

### DIFF
--- a/ff-macros/src/lib.rs
+++ b/ff-macros/src/lib.rs
@@ -76,7 +76,7 @@ pub fn define_field(input: TokenStream) -> TokenStream {
             .expect("generator should be a decimal integer string");
 
         let (small_subgroup_base, small_subgroup_power) =
-            match utils::detect_small_prime_subgroup(&modulus_big) {
+            match utils::find_conservative_subgroup_base(&modulus_big) {
                 Some((base, power)) => (Some(base), Some(power)),
                 None => (None, None),
             };
--- a/ff-macros/src/lib.rs
+++ b/ff-macros/src/lib.rs
@@ -8,6 +8,7 @@
 #![forbid(unsafe_code)]
 
 use num_bigint::BigUint;
+use num_traits::Zero;
 use proc_macro::TokenStream;
 use quote::format_ident;
 use syn::{Expr, ExprLit, Item, ItemFn, Lit, Meta};
@@ -75,11 +76,28 @@ pub fn define_field(input: TokenStream) -> TokenStream {
             .parse::<BigUint>()
             .expect("generator should be a decimal integer string");
 
+        // Auto-detect small subgroup params (base 3)
+        let mut trace = &modulus_big - BigUint::from(1u32);
+        while trace.bit(0) == false {
+            trace >>= 1u32;
+        }
+        let base = BigUint::from(3u32);
+        let mut power = 0u32;
+        while (&trace % &base).is_zero() {
+            trace /= &base;
+            power += 1;
+        }
+        let (small_subgroup_base, small_subgroup_power) = if power > 0 {
+            (Some(3u32), Some(power))
+        } else {
+            (None, None)
+        };
+
         let config_impl = montgomery::mont_config_helper(
             modulus_big,
             generator_big,
-            None,
-            None,
+            small_subgroup_base,
+            small_subgroup_power,
             config_name.clone(),
         );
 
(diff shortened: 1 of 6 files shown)


ff-macros/src/small_fp/montgomery_backend.rs
```diff
@@ -56,6 +57,19 @@ pub(crate) fn backend_impl(
 
     let neg_one_mont = mod_mul_const(modulus - 1, r_mod_n, modulus);
 
+    let mixed_radix_impl = if let Some((base, power)) = detect_small_prime_subgroup(modulus, two_adicity)
```
z-tech: Set for SmallFp

ff-macros/src/small_fp/montgomery_backend.rs
```diff
@@ -56,6 +57,20 @@ pub(crate) fn backend_impl(
 
     let neg_one_mont = mod_mul_const(modulus - 1, r_mod_n, modulus);
 
+    let modulus_big = num_bigint::BigUint::from(modulus);
```
z-tech, line 42: Set for SmallFp

ff-macros/src/small_fp/utils.rs
```diff
@@ -79,6 +79,22 @@ pub(crate) const fn find_quadratic_non_residue(modulus: u128) -> u128 {
     }
 }
 
+/// Compute the large subgroup root of unity:
+/// generator^((p-1) / (2^two_adicity * base^power))
+pub(crate) fn compute_large_subgroup_root(
```
z-tech, line 84: It didn't seem worth reusing the Fp version of this function so I added a new one just for SmallFp

ff-macros/src/lib.rs
```diff
@@ -75,11 +75,17 @@ pub fn define_field(input: TokenStream) -> TokenStream {
             .parse::<BigUint>()
             .expect("generator should be a decimal integer string");
 
+        let (small_subgroup_base, small_subgroup_power) =
```
z-tech, line 78: Set for Fp

ff-macros/src/utils.rs
```diff
@@ -87,13 +87,13 @@ pub(crate) fn fp_alias_for_limbs(limbs: usize) -> TokenStream2 {
...
 
 pub(crate) fn parse_string(input: TokenStream) -> Option<String> {
-    let input: Expr = syn::parse(input).unwrap();
-    let input = if let Expr::Group(syn::ExprGroup { expr, .. }) = input {
-        expr
-    } else {
-        panic!("could not parse");
-    };
-    match *input {
+    let mut input: Expr = syn::parse(input).unwrap();
```
z-tech, line 90: Claude says this is better for avoiding proc macro parsing errors in rust analyzer

ff-macros/src/utils.rs
```diff
@@ -102,6 +102,33 @@ pub(crate) fn parse_string(input: TokenStream) -> Option<String> {
     }
 }
 
+/// Detect a small prime subgroup of the multiplicative group.
+///
+/// Checks whether any small prime base in {3, 5, 7} divides the odd part of
+/// p-1 at least once. Returns the smallest such `(base, adicity)` if found,
+/// or `None` if the odd part has no factors ≤ 7.
+pub(crate) fn find_conservative_subgroup_base(modulus: &BigUint) -> Option<(u32, u32)> {
```
z-tech, line 110: I expect this to be used only for define_macro! it selects a plausibly useful value for subgroup base and None otherwise (None is actually favorable in some cases).

The user is still free to set their own base params using either the Fp or SmallFp macros.

I thought 3, 5, and 7 were the reasonable primes to check btw.

weikengchen (approved):

PR #1088, opened 2026-03-30
Support only u64 sized primes for SmallFp

### LINKED ISSUE #1089: Support only <= 64 bit primes in SmallFp
Though addition showed some comparative improvement at the outset, I don't think the numbers back up supporting this increased complexity. This is especially true now that we have define_field! macro that chooses the appropriate representation for the user.

Advantages:
- less code to optimize and maintain
- simpler to read and indeed some of the code like multiplication was already kind of duplicated
- we can safely hardcode this lingering const N "number of limbs" to 1, which is kind of a win in itself
- the target primes nowadays and <= u64 anyway

If in the future we want a 128 bit prime to be faster we could focus on optimizing in the Fp code.

z-tech:
Though arithmetic of primes 64-128 bits showed some potential at the outset, I don't think the numbers back up supporting this increased complexity especially when the optimizations here can most likely be applied directly to the Fp implementation.

We also have the define_field! macro that chooses the appropriate representation for the user.

Advantages:
- less code to optimize and maintain
- simpler to read and indeed some of the code like multiplication was already kind of duplicated
- we can safely hardcode this lingering const N "number of limbs" to 1, which is kind of a win in itself
- the target primes nowadays are <= u64 anyway

Disadvantages:
- would have been nice to remove before putting in main
- is there a performance win? (I think less so now)

closes: #1089

---

write a little note why.

- [ ] Wrote unit tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@
 - (`ark-ec`) Improve GLV scalar multiplication performance by skipping leading zeroes.
 - (`ark-poly`) Make `SparsePolynomial.coeffs` field public
 - [\#1039](https://github.com/arkworks-rs/algebra/pull/1039) (`ark-ff-asm`) Remove unused dead spill buffer path.
-- [\#1044](https://github.com/arkworks-rs/algebra/pull/1044) Add implementation for small field with native integer types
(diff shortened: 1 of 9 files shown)


z-tech:
@benbencik

weikengchen (approved):

weikengchen:
I will let these few PRs share the same CHANGELOG entry with the one that adds SmallFp since they can be considered as refinement.

PR #1098, opened 2026-04-26
Update ark-ff readme regarding SmallFp

benbencik:
In the former https://github.com/arkworks-rs/algebra/pull/1044/, I forgot to update the README in crate `ff`. Specifically, the `SmallField` now supports only arithmetic up to `u64` based on https://github.com/arkworks-rs/algebra/pull/1088. I also added a mention of the new `define_field` macro.

### DIFF
--- a/ff/README.md
+++ b/ff/README.md
@@ -40,11 +40,11 @@ The above two models serve as abstractions for constructing the extension fields
 
 ## Instantiation
 
-You can instantiate fields in two ways:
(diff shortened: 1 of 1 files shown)


z-tech (approved):

weikengchen (approved):
Docs-only follow-up to #1044/#1088. Verified the updated claims against the source: `SmallFp` backing types top out at u64 (primitive_type_bit_size in small_fp_backend.rs), and the `define_field!` macro signature/behavior matches the README example. LGTM.

(Claude wrote this)

PR #1106, opened 2026-05-16
fix(poly): honor coset vanishing polynomial in dense multiply

### LINKED ISSUE #1105: DensePolynomial::mul_by_vanishing_poly ignores coset offset
## Summary

`DensePolynomial::mul_by_vanishing_poly(domain)` appears to always compute multiplication by the subgroup vanishing polynomial:

```text
x^n - 1
```

That is correct for ordinary subgroup domains. However, `EvaluationDomain` also supports coset domains via `get_coset`, and for a coset domain the vanishing polynomial is:

```text
x^n - offset^n
```

where `offset^n` is available as `domain.coset_offset_pow_size()`.

As a result, passing a coset domain to `mul_by_vanishing_poly` returns `p(x) * (x^n - 1)` instead of `p(x) * (x^n - offset^n)`.

## Difference from #902

Related to #902, but separate: #902 discusses `divide_by_vanishing_poly`, while this report is about `mul_by_vanishing_poly` and the repro below does not exercise division.

## References

- [`DensePolynomial::mul_by_vanishing_poly`](https://github.com/arkworks-rs/algebra/blob/3b2a12492cfeb34774a2cdbe9d8d7a869c883fa9/poly/src/polynomial/univariate/dense.rs#L157-L164): this is the affected helper. It builds `x^n * p(x)` and then subtracts `p(x)` coefficient-wise, which implements `p(x) * (x^n - 1)`.
- [`EvaluationDomain::get_coset`](https://github.com/arkworks-rs/algebra/blob/3b2a12492cfeb34774a2cdbe9d8d7a869c883fa9/poly/src/domain/mod.rs#L55-L57): this is the trait method used to construct the coset domain passed into the helper.
- [`EvaluationDomain::coset_offset_pow_size`](https://github.com/arkworks-rs/algebra/blob/3b2a12492cfeb34774a2cdbe9d8d7a869c883fa9/poly/src/domain/mod.rs#L88-L90): this is the trait method exposing the `offset^n` value that distinguishes a coset vanishing polynomial from `x^n - 1`.
- [`EvaluationDomain::vanishing_polynomial`](https://github.com/arkworks-rs/algebra/blob/3b2a12492cfeb34774a2cdbe9d8d7a869c883fa9/poly/src/domain/mod.rs#L224-L228): the trait's own vanishing-polynomial helper uses `-coset_offset_pow_size()` as the degree-0 coefficient, i.e. it returns `x^size - offset^size`.
- [`GeneralEvaluationDomain::get_coset`](https://github.com/arkworks-rs/algebra/blob/3b2a12492cfeb34774a2cdbe9d8d7a869c883fa9/poly/src/domain/general.rs#L140-L145): this forwards coset construction to the concrete domain, so the repro uses the public general-domain API while still exercising a real coset domain.

## Affected Code

The current helper builds `x^n * p(x)` and then subtracts `p(x)` coefficient-wise:

```rust
pub fn mul_by_vanishing_poly<D: EvaluationDomain<F>>(&self, domain: D) -> Self {
    let mut shifted = vec![F::zero(); domain.size()];
    shifted.extend_from_slice(&self.coeffs);
    cfg_iter_mut!(shifted)
        .zip(&self.coeffs)
        .for_each(|(s, c)| *s -= c);
    Self::from_coefficients_vec(shifted)
}
```

For `p(x) = a + bx` and `domain.size() = n`, this builds:

```text
[-a, -b, 0, ..., 0, a, b]
```

which is `p(x) * (x^n - 1)`.

For a coset domain, the low coefficients should instead be scaled by `offset^n`:

```text
[-offset^n * a, -offset^n * b, 0, ..., 0, a, b]
```

## Minimal Reproduction

This patch adds a standalone integration test. It compares the helper output against both coefficient formulas directly, without relying on another polynomial multiplication implementation to define the expected value.

```diff
diff --git a/poly/tests/coset_vanishing_mul.rs b/poly/tests/coset_vanishing_mul.rs
new file mode 100644
index 000000000..000000000
--- /dev/null
+++ b/poly/tests/coset_vanishing_mul.rs
@@ -0,0 +1,50 @@
+use ark_ff::{One, Zero};
+use ark_poly::{
+    univariate::DensePolynomial, DenseUVPolynomial, EvaluationDomain, GeneralEvaluationDomain,
+};
+
+#[test]
+fn mul_by_vanishing_poly_matches_coset_vanishing_polynomial() {
+    use ark_test_curves::bls12_381::Fr;
+
+    let domain = GeneralEvaluationDomain::<Fr>::new(8).unwrap();
+    let coset = domain.get_coset(Fr::from(5u64)).unwrap();
+    let offset_pow_size = coset.coset_offset_pow_size();
+    assert_ne!(offset_pow_size, Fr::one());
+
+    let a = Fr::from(3u64);
+    let b = Fr::from(7u64);
+    let p = DensePolynomial::from_coefficients_vec(vec![a, b]);
+
+    let actual = p.mul_by_vanishing_poly(coset);
+
+    // p(x) * (x^8 - 1)
+    let subgroup_formula = DensePolynomial::from_coefficients_vec(vec![
+        -a,
+        -b,
+        Fr::zero(),
+        Fr::zero(),
+        Fr::zero(),
+        Fr::zero(),
+        Fr::zero(),
+        Fr::zero(),
+        a,
+        b,
+    ]);
+
+    // p(x) * (x^8 - offset^8)
+    let coset_formula = DensePolynomial::from_coefficients_vec(vec![
+        -(offset_pow_size * a),
+        -(offset_pow_size * b),
+        Fr::zero(),
+        Fr::zero(),
+        Fr::zero(),
+        Fr::zero(),
+        Fr::zero(),
+        Fr::zero(),
+        a,
+        b,
+    ]);
+
+    assert_ne!(subgroup_formula, coset_formula);
+    assert_eq!(actual, coset_formula);
+}
```

Run:

```bash
cargo test -p ark-poly --features std --test coset_vanishing_mul -- --nocapture
```

## Observed Behavior

The expected-behavior test currently fails because `actual` is the subgroup formula rather than the coset formula:

```text
thread 'mul_by_vanishing_poly_matches_coset_vanishing_polynomial' panicked at poly/tests/coset_vanishing_mul.rs:48:5:
assertion `left == right` failed
```

The failure shows that the helper output matches `p(x) * (x^8 - 1)` even though the supplied domain is a coset and `coset.coset_offset_pow_size() != 1`.

## Expected Behavior

I would expect `mul_by_vanishing_poly(domain)` to multiply by `domain.vanishing_polynomial()` for the provided domain.

Equivalently, for coset domains it should compute:

```text
x^n * p(x) - domain.coset_offset_pow_size() * p(x)
```

If `mul_by_vanishing_poly` is intended to support only subgroup domains, then the method documentation may need to state that precondition, because the current signature accepts any `D: EvaluationDomain<F>`.

## Possible Fix Direction

One possible implementation direction is to use `domain.coset_offset_pow_size()` when subtracting the low coefficients:

```rust
let offset_pow_size = domain.coset_offset_pow_size();
cfg_iter_mut!(shifted)
    .zip(&self.coeffs)
    .for_each(|(s, c)| *s -= offset_pow_size * *c);
```

For subgroup domains this preserves current behavior because `coset_offset_pow_size() == F::one()`.

A regression test should cover a coset domain where `coset_offset_pow_size() != F::one()`. A useful assertion would be that the helper matches either `domain.vanishing_polynomial()` or the direct coefficient formula above.

1sgtpepper:
## Summary

Fixes #1105.

`DensePolynomial::mul_by_vanishing_poly` currently behaves as if every domain has vanishing polynomial `x^n - 1`. That is correct for ordinary subgroup domains, but not for coset domains, whose vanishing polynomial is `x^n - domain.coset_offset_pow_size()`.

This PR updates the helper to use `domain.coset_offset_pow_size()` when subtracting the low coefficients, so it matches the vanishing polynomial of the domain it is given.

## Details

The implementation keeps the existing coefficient-form shortcut for multiplying by a monic binomial:

```text
p(x) * (x^n - k) = x^n * p(x) - k * p(x)
```

where `k = domain.coset_offset_pow_size()`.

For subgroup domains, `k == 1`, so the previous fast path is preserved. For coset domains, the low coefficients are scaled by `k` before subtraction.

## Tests

- Extended the existing randomized `mul_by_vanishing_poly` test to check both subgroup and coset domains against `domain.vanishing_polynomial()`.
- Added a focused regression test for a size-8 coset domain using `p(x) = a + bx`, with direct expected coefficients for `p(x) * (x^8 - offset^8)`.

## Validation

Ran locally:

```text
cargo fmt --check
cargo test -p ark-poly --features std mul_by_vanishing_poly -- --nocapture
cargo test -p ark-poly --features std
```

Results:

```text
focused test: 2 passed
full ark-poly unit tests: 114 passed
ark-poly doc tests: 11 passed
```

## Scope

This intentionally only changes `mul_by_vanishing_poly`. It does not change `divide_by_vanishing_poly`, which is a separate issue area.

---

write a little note why.

- [x] Wrote unit tests
- [x] Updated relevant documentation in the code — the existing method documentation states the domain contract, and the implementation comment explains the coset formula.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,6 +37,7 @@
 ### Bugfixes
 
 - [\#1082](https://github.com/arkworks-rs/algebra/pull/1082) (`ark-ff`) Fix `SmallFp::from_random_bytes` / `from_be_bytes_mod_order` silently producing incorrect field elements by treating plaintext bytes as Montgomery-encoded.
+- [\#1106](https://github.com/arkworks-rs/algebra/pull/1106) (`ark-poly`) Make `DensePolynomial::mul_by_vanishing_poly` honor coset-domain offsets.
 
 ## v0.5.0
 
--- a/poly/src/polynomial/univariate/dense.rs
+++ b/poly/src/polynomial/univariate/dense.rs
@@ -155,22 +155,10 @@ impl<F: FftField> DensePolynomial<F> {
     /// Multiply `self` by the vanishing polynomial for the domain `domain`.
     /// Returns the result of the multiplication.
     pub fn mul_by_vanishing_poly<D: EvaluationDomain<F>>(&self, domain: D) -> Self {
-        if self.is_zero() {
-            return Self::zero();
-        }
-
-        let domain_size = domain.size();
-        let mut shifted = Vec::with_capacity(domain_size + self.coeffs.len());
-        shifted.resize(domain_size, F::zero());
+        let mut shifted = vec![F::zero(); domain.size()];
         shifted.extend_from_slice(&self.coeffs);
 
-        // The domain vanishing polynomial is x^n - k for
-        // k = domain.coset_offset_pow_size(). At this point `shifted` already
-        // contains x^n * self: `domain_size` leading zero coefficients followed
-        // by the coefficients of `self`. Finishing the multiplication therefore
-        // only requires subtracting k * self from the low coefficients. This
-        // keeps the helper as a specialized binomial multiply rather than
-        // falling back to general polynomial multiplication.
+        // Coset domains vanish on x^n - offset^n, not x^n - 1.
         let offset_pow_size = domain.coset_offset_pow_size();
         if offset_pow_size == F::ONE {
             cfg_iter_mut!(shifted)
--- a/poly/src/polynomial/univariate/dense.rs
+++ b/poly/src/polynomial/univariate/dense.rs
@@ -155,11 +155,32 @@ impl<F: FftField> DensePolynomial<F> {
     /// Multiply `self` by the vanishing polynomial for the domain `domain`.
     /// Returns the result of the multiplication.
     pub fn mul_by_vanishing_poly<D: EvaluationDomain<F>>(&self, domain: D) -> Self {
-        let mut shifted = vec![F::zero(); domain.size()];
+        if self.is_zero() {
+            return Self::zero();
+        }
+
+        let domain_size = domain.size();
+        let mut shifted = Vec::with_capacity(domain_size + self.coeffs.len());
+        shifted.resize(domain_size, F::zero());
         shifted.extend_from_slice(&self.coeffs);
-        cfg_iter_mut!(shifted)
-            .zip(&self.coeffs)
-            .for_each(|(s, c)| *s -= c);
+
+        // The domain vanishing polynomial is x^n - k for
+        // k = domain.coset_offset_pow_size(). At this point `shifted` already
+        // contains x^n * self: `domain_size` leading zero coefficients followed
+        // by the coefficients of `self`. Finishing the multiplication therefore
+        // only requires subtracting k * self from the low coefficients. This
+        // keeps the helper as a specialized binomial multiply rather than
+        // falling back to general polynomial multiplication.
+        let offset_pow_size = domain.coset_offset_pow_size();
+        if offset_pow_size == F::ONE {
+            cfg_iter_mut!(shifted)
+                .zip(&self.coeffs)
+                .for_each(|(s, c)| *s -= c);
+        } else {
+            cfg_iter_mut!(shifted).zip(&self.coeffs).for_each(|(s, c)| {
+                *s -= *c * offset_pow_size;
+            });
+        }
         Self::from_coefficients_vec(shifted)
     }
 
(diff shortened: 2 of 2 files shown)


weikengchen:
why would you do this?

let mut shifted = Vec::with_capacity(domain_size + self.coeffs.len());
shifted.resize(domain_size, F::zero());

1sgtpepper:
Yeah, fair. That allocation change isn’t needed for this fix, so I reverted it and left the original `vec![F::zero(); domain.size()]` shape in place.

1sgtpepper:
@weikengchen Gentle follow-up: the cleanup is done, the branch is rebased, and CI is green. Could you take another look?

PR #1108, opened 2026-05-24
perf: Use 2-bit windowed GLV scalar multiplication

waamm:
## Description

Inspired by `gnark-crypto`, use 2-bit windowed GLV scalar multiplication. For each of the three curves (BLS12-381 G1, BLS12-377 G1 and BN254 G1) in `arkworks` which use GLV code, I'm getting a 15% speedup for scalar multiplication on an M4 Max.

~~Also updates the changelog from v0.5.0 to v0.6.0 (which seems to have been forgotten in an earlier PR).~~

- [ ] Wrote unit tests (already exist)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Pending
 
-- [\#1108](https://github.com/arkworks-rs/algebra/pull/1108) (`ark-curves`) Improve scalar multiplation speed for curves using GLV.
+- [\#1108](https://github.com/arkworks-rs/algebra/pull/1108) (`ark-curves`) Improve scalar multiplication speed for curves using GLV.
 
 ## v0.6.0
 
@@ -12,6 +12,7 @@ Breaking bump across the arkworks stack — first release since 0.5.0 (2024-10-2
 ## Changes since v0.5.0
 
 **New features**
+
 - SmallFp: dedicated small-prime field implementation (#1044, #1082, #1084, #1088, #1091)
 - Stark curve + tests for large 2-adicity fields (#1001)
 - Double-odd curve support in `ec` (#986)
@@ -26,17 +27,20 @@ Breaking bump across the arkworks stack — first release since 0.5.0 (2024-10-2
 - Conditional `infinity` flag on `short_weierstrass::Affine` (#639)
 
 **Breaking trait-surface changes** (the reason this is 0.6.0)
+
 - `AffineRepr: Neg<Output = Self>` tightened (#927)
 - New required const `MINUS_ONE` on `Field`/`FpConfig` (#1004)
 - R1CS → GR1CS migration (#949)
 
 **Performance**
+
 - Fix GLV scalar-multiplication perf bug (#1025)
 - `VariableBaseMSM` improvements for small scalars (#995, #996)
 - Extended Jacobian coordinates for SW MSM (#961)
 - Generalized Mersenne mul path (#1091)
 
-**Bug fixes**
+**Bugfixes**
+
 - `ToConstraintField` bound for `Affine` (#1005)
 - `to_dense_multilinear_extension` for nv > 30 (#1022)
 - `to_evaluations` for sparse multilinear extension (#1019)
@@ -46,6 +50,7 @@ Breaking bump across the arkworks stack — first release since 0.5.0 (2024-10-2
 - Build of `ark-ff` (#1028)
 
 **Refactors / housekeeping**
+
 - Polynomial refactors across `DensePolynomial`, `SparsePolynomial`, `GeneralEvaluationDomain`
 - `find_naf`, biginteger, cubic-extension simplifications in `ff`
 - Clippy rule rollout across all member crates
@@ -59,10 +64,10 @@ Full list: `git log v0.5.0..v0.6.0`.
 Step 2 of the 0.6.0 cascade: std → **algebra** → snark → r1cs-std → crypto-primitives → poly-commit → groth16 → circom-compat.
 
 ## Verification
+
 - [x] `cargo test --workspace --all-features` green
 - [x] `cargo publish --workspace --dry-run` clean (root + `curves/`)
 - [x] CI green on master
-
 
 ## v0.5.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,36 +2,67 @@
 
 ## Pending
 
-- (`ark-starkcurve`) Add 252 bit [Stark curve](https://docs.starknet.io/architecture/cryptography/#the_stark_curve).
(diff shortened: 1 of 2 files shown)


weikengchen:
(This code review was done by Kimi K3 in Zro)

Thanks for the update — the changelog looks good now. I've taken a close look at the code and re-verified the performance claims. Overall this is a clean, well-contained change and I'd like to move it forward. A few comments below, plus one strategic question at the end.

## Performance verification

Reproduced on an Apple M2 Ultra (`Scalar Multiplication` criterion bench, 100 samples, vs. saved master baseline):

| Curve | master | this PR | change |
|---|---|---|---|
| BLS12-381 G1 | 264–272 µs | 207.9 µs | **−21…−23 %** |
| BLS12-377 G1 | 261.8 µs | 209.1 µs | **−21.7 %** |
| BN254 G1 | 137.3 µs | 124.6 µs | **−9.5 %** |

All significant at p = 0.00. The claimed ~15% checks out.

Both the existing `glv` test suites (all three curves) and additional edge cases I ran locally (k = 0, 1, 2, 3, r−1, r−2, 2^127, 2^128, u64::MAX, plus a few hundred randoms, for both `glv_mul_projective` and `glv_mul_affine`) pass.

## 1. Attribution to gnark-crypto

The 15-entry table layout and the scan loop are not just inspired by but a direct port of gnark-crypto's `mulGLV` (see e.g. https://github.com/ConsenSys/gnark-crypto/blob/v0.19.0/ecc/bls12-381/g1.go#L620 — same `table[(c2<<2|c1)-1]` layout and fill order). gnark-crypto is Apache-2.0 (© Consensys Software Inc., no NOTICE file), so we're license-compatible, but we should carry an attribution comment — this repo already does that elsewhere (e.g. `Adapted from …` in `ec/src/models/bls12/mod.rs` and `ff/src/fields/models/fp/montgomery_backend.rs`). Could you add something along these lines to the doc comment of `glv_precompute_table`:

```rust
/// Adapted from gnark-crypto's `mulGLV` (Apache-2.0, Copyright Consensys Software Inc.):
/// https://github.com/ConsenSys/gnark-crypto/blob/v0.19.0/ecc/bls12-381/g1.go#L620
/// implementing the GLV method (https://www.iacr.org/archive/crypto2001/21390189.pdf).
```

## 2. Dead — and subtly wrong — padding logic in `glv_two_bit_digit`

`BitIteratorBE::new(k.into_bigint())` always yields exactly `64 · N` bits for `BigInt<N>` (see `ff/src/bits.rs`), and `k1`/`k2` share the same `ScalarField` type. Consequences:

- `max(bits1.len(), bits2.len())` is always a no-op,
- the odd-length `len + 1` padding never fires,
- the `else if` / `else` fallback branches in `glv_two_bit_digit` are unreachable.

Worse: if an odd length ever *did* occur (say someone later switches to `BitIteratorBE::without_leading_zeros`), the padding is applied at the LSB end while 2-bit windows must be anchored at the LSB — I verified exhaustively that the current code computes wrong digits for every odd bit-length (e.g. 3 bits: `0b101` scans out as 9, not 5).

Suggestion: delete the `max`/padding logic and the two fallback branches, and assert the invariant instead:

```rust
let nbits = bits1.len(); // == bits2.len(), always a multiple of 64
debug_assert_eq!(nbits, bits2.len());
```

This also fixes the new `clippy::nonminimal_bool` warning on the `if !started` nesting while you're at it.

## 3. Two `Vec<bool>` heap allocations per scalar multiplication

`bits1`/`bits2` materialize two 256-element vectors per multiplication. gnark avoids this by reading 2-bit windows directly out of the 64-bit words. Worth doing the same here — for each window `w`, the bits of `k` live at fixed limb/bit offsets, e.g. via `BigInt::get_bit` or a small helper over `as_ref()[u64]` limbs. Not a blocker, but it's free savings and removes the only new allocation in this hot path (likely part of why BN254's gain, ~9.5%, is smaller than the BLS curves' ~22%).

## 4. `glv_mul_affine` silently changes its arithmetic profile

It now routes through the projective table (full projective additions) instead of the previous mixed additions on affine bases. It's unused in-tree (only exercised by test-templates) so I don't mind the change, but downstream users overriding `mul_affine` with it will see different performance characteristics. A one-line note on the method would be enough.

## 5. Question: 2-bit table now, or jump straight to wNAF-5?

Worth knowing: gnark-crypto has since replaced this exact 2-bit scheme with interleaved **wNAF-5** (https://github.com/ConsenSys/gnark-crypto/pull/788, not yet in any release), reporting a further **−6%** on BLS12-381 G1 (and −13% on BW6-761) on top of the 2-bit table. The wNAF variant needs two 8-entry odd-multiple tables and a subtraction path, which `ark-ec` supports equally well.

My preference: land this PR as-is — it's clean, verified, and a solid 10–23% win — and pursue wNAF-5 as a follow-up PR so each step stays easy to review and bisect. Happy to open a tracking issue. But if you'd rather fold wNAF-5 into this PR, I'm open to reviewing that too. WDYT?

weikengchen:
(^ @waamm you might want to use an AI agent to follow up with the response above)

waamm:
Thanks for the thorough review!

> Attribution to `gnark-crypto`

Added.

> Dead — and subtly wrong — padding logic

Digit extraction now reads 2-bit windows straight from the u64 limbs.

> Worth doing the same here

Done. On my machine it’s another ~2% speed improvement.

> likely part of why BN254's gain, ~9.5%, is smaller than the BLS curves' ~22%

I'm a bit skeptical; feel free to re-bench if you want.

> glv_mul_affine silently changes its arithmetic profile

Noted on the method. Also collapsed it to `glv_mul_projective(p.into(), k).into_affine()` since it was duplicate.

> My preference: land this PR as-is

Agreed.

weikengchen:
Thanks for the quick turnaround — the follow-up looks good and I've verified it:

- Correctness: full `ark-ec` + all three curves' test suites pass, plus edge cases crossing every limb boundary (2^63/64/65, 2^127/128/129, 2^191/192/193, 2^255/256) and a few hundred random scalars, on both `glv_mul_projective` and `glv_mul_affine`. Clippy and fmt are clean on `glv.rs`.

- Perf (Apple M2 Ultra, vs. saved master baseline, all p = 0.00):

  | Curve | master | this PR | change |
  |---|---|---|---|
  | BLS12-381 G1 | 264–272 µs | 211.1 µs | −22.8 % |
  | BN254 G1 | 137.3 µs | 117.4 µs | −15.1 % |
  | BLS12-377 G1 | 261.8 µs | ~209 µs | ~−21.7 % |

  Correction to my earlier review: the previous BN254 figure (~9.5%) was measured on a
  noisy machine — re-running clean gives **−15.1%**, consistent with the claimed ~15%.
  You were right to be skeptical there.

No blocking issues from my side. I will start reviewing the code and preparing it to be merged.

weikengchen (approved):

PR #1109, opened 2026-05-28
perf(poly): reduce allocations in DenseMultilinearExtension Sub and parallelize scalar Mul

wstran:
## Summary

Two allocation/parallelism improvements to `DenseMultilinearExtension` arithmetic, in the spirit of #1061:

- **`Sub` (`&Self - &Self`)** previously evaluated `self + &rhs.clone().neg()`, allocating three full evaluation tables (clone, negation, sum) across two passes. This fuses it into a single `cfg_iter!` pass with one allocation. The zero-operand short-circuits and the `num_vars` assertion mirror the existing `Add` impl, so the output is identical.
- **Scalar `Mul` (`&Self * &F`)** was the only arithmetic operator still using a serial `.iter()`; it now uses `cfg_iter!`, matching `Add`, `Neg`, and `Sub`.

Also adds a `ScalarMul` case to the dense-multilinear benchmark and a `sub_with_zero` regression test (the previous tests did not cover the zero-operand paths of `Sub`).

## Correctness

Output is unchanged. `cargo test -p ark-poly` passes with and without `--features parallel`, including the existing `arithmetic` test and the new `sub_with_zero`.

## Benchmarks

`criterion`, `bls12_381::Fr`, before/after toggled via `git stash`, 12-core machine, 100 samples / 5 s. `nv` = number of variables (table size `2^nv`); median change:

| nv | Sub (serial) | Sub (parallel) | ScalarMul (parallel) |
|---:|---:|---:|---:|
| 14 | −46% | −54% | −62% |
| 16 | −30% | −64% | −80% |
| 18 | −29% | −61% | −85% |
| 20 | −29% | −58% | −86% |

Serial `ScalarMul` is unchanged (in serial builds `cfg_iter!` expands to `.iter()`).

**Tradeoff (parallel `ScalarMul` only):** for very small tables the rayon dispatch overhead exceeds the work, so `ScalarMul` regresses for `nv ≤ 11` (e.g. `nv = 10`: ~11 µs → ~54 µs), crossing over at `nv ≈ 12`. This matches the existing unconditional parallelization of `Add`/`Neg`. `Sub` improves at every size in both modes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@
 - [\#1039](https://github.com/arkworks-rs/algebra/pull/1039) (`ark-ff-asm`) Remove unused dead spill buffer path.
 - [\#1044](https://github.com/arkworks-rs/algebra/pull/1044), [\#1084](https://github.com/arkworks-rs/algebra/pull/1084), [\#1088](https://github.com/arkworks-rs/algebra/pull/1088) Add implementation for small field with native integer types
 - [\#1061](https://github.com/arkworks-rs/algebra/pull/1061) (`ark-poly`) Reduce allocations in `DenseMultilinearExtension::{concat, fix_variables, evaluate}`.
+- [\#1109](https://github.com/arkworks-rs/algebra/pull/1109) (`ark-poly`) Reduce allocations in `DenseMultilinearExtension::Sub` and parallelize scalar `Mul`.
 
 ### Breaking changes
 
--- a/poly/benches/dense_multilinear.rs
+++ b/poly/benches/dense_multilinear.rs
@@ -29,6 +29,16 @@ fn arithmetic_op_bench<F: Field>(c: &mut Criterion) {
         });
     }
     group.finish();
+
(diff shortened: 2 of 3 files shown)


weikengchen:
(from ChatGPT 5.6)

I reviewed the implementation and independently reproduced the benchmark improvement. The serial and parallel test suites and doctests all pass, and the `Sub` optimization consistently improves performance. With Rayon limited to the author’s 12-thread setup, parallel `Sub` improved by roughly 58–66% and `ScalarMul` by 50–91% for `nv = 14–20`. One caveat is that the `ScalarMul` crossover point is hardware/thread-count dependent: with 24 Rayon threads, `nv = 14` regressed and the crossover moved to around `nv = 15–16`, so the PR description should avoid presenting `nv ≈ 12` as universal. The code itself looks sound; the PR only needs rebasing to resolve the current `CHANGELOG.md` conflict.

--- I will take care of getting this one merged.

weikengchen (approved):

PR #1112, opened 2026-06-15
fix(ec): rayon::ThreadPoolBuilder panic in wasm32

### LINKED ISSUE #1111: ark-ec's `msm_unchecked` panics when targeting `wasm32-unknown-unknown` with `+parallel`
```
error output:
    panicked at /home/rmacwha/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/ark-ec-0.6.0/src/scalar_mul/variable_base/mod.rs:549:18:
    called `Result::unwrap()` on an `Err` value: ThreadPoolBuildError { kind: IOError(Error { kind: Unsupported, message: "operation not supported on this platform" })
```

Rust panics because `rayon::ThreadPoolBuilder` cannot be created on wasm32-unknown-unknown.  This can be avoiding by disabling the parallel feature flag, but leaves performance on the table since wasm-bindgen-rayon can otherwise be used to enable rayon in browser environments.  It would be better to disable this particular thread construction pool (since afaik it's the only place where a local thread pool is constructed?) and fallback to the `not(feature = "parallel")` case when `target_arch = "wasm32"`.

Panic is throw in `msm_bigint_wnaf` at line [546](https://github.com/arkworks-rs/algebra/blob/af564e48d0109a9df81d45ff55ff0b628ece319d/ec/src/scalar_mul/variable_base/mod.rs#L549).

Robert-MacWha:
## Description

Adds a `target_arch = "wasm32"` feature conditional on the relevant block, falling back to the `not(feature = "parallel")` case in wasm32.

closes: #1111

---

write a little note why.

- [ ] Wrote unit tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@
 - [\#1039](https://github.com/arkworks-rs/algebra/pull/1039) (`ark-ff-asm`) Remove unused dead spill buffer path.
 - [\#1044](https://github.com/arkworks-rs/algebra/pull/1044), [\#1084](https://github.com/arkworks-rs/algebra/pull/1084), [\#1088](https://github.com/arkworks-rs/algebra/pull/1088) Add implementation for small field with native integer types
 - [\#1061](https://github.com/arkworks-rs/algebra/pull/1061) (`ark-poly`) Reduce allocations in `DenseMultilinearExtension::{concat, fix_variables, evaluate}`.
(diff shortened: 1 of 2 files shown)


weikengchen:
Thanks again for raising this and for the patience here — apologies for the inconvenience the regression caused.

From our review, PR #1112 is now in good shape and ready to merge. The fix is necessary for wasm32: it avoids creating a new `ThreadPoolBuilder` on targets where Rayon cannot create a real thread pool, while still keeping the existing `msm_bigint_wnaf_parallel` flow in place so parallelization can still happen when supported by the build/runtime. The behavior is aligned with current Rayon thread model constraints on wasm.

Drafted by 5.3-Codex-Spark

weikengchen (approved):

PR #1115, opened 2026-06-22
feat(serialize): implement Canonical(De)Serialize and Valid for Box<T>

### LINKED ISSUE #1007: ark-serialize support for boxed::Box and borrow::Cow
If we merely add `T: ?Sized` to `Cow<'static,T>` then I'd expect `Cow<'static,[T]>` should then serialzie exactly like `[T]`, and hence like `Vec<T>`.

We want `Cow<'static,[T]>` to deserialize exactly like `Vec<T>`, which requires one new impl, but not sure if the orphan rules permit this, if an explict `T: Sized` helps, etc. 

`Box<T>` could serialzie exactly like `T`, so then `Box<[T]>` would serialzie exactly like `[T]`, and hence like `Vec<T>`.

`Box<T>` could deserialize exactly like `T`.  As `[T]` has no deserializer, `Box<[T]>` could deserialize exactly like `Vec<T>`, no?

aryaethn:
## What

`ark-serialize` already implements the serialization traits for `Box<[T]>` (via the sequence impls in `collections.rs`) and for `Cow<'_, T>` (#1003), but not for the generic `Box<T>`. This adds `CanonicalSerialize`, `Valid`, and `CanonicalDeserialize` for `Box<T>`, delegating to the inner `T` so a boxed value round-trips byte-identically to the value itself.

Closes #1007.

## Details

- The impls are bounded to `T: Sized` on purpose: the unsized slice case `Box<[T]>` is already covered by the sequence impls, and a `?Sized` impl here would conflict with those. (You can't deserialize into `Box<dyn Trait>` anyway.)
- `CanonicalDeserialize` uses `Self::new`, matching the neighbouring `Arc` impl and the workspace `use_self` lint.

Out of scope, per the discussion in #1007:

- `Rc<T>` can't gain `Valid` / `CanonicalDeserialize`: `Valid: Sync` (needed for `batch_check`'s rayon path) and `Rc<T>` is never `Sync`. Relaxing that, e.g. cfg-gating `Valid: Sync` on `parallel`, is semver-sensitive and left for a follow-up.

## Tests

`test_box` in `serialize/src/test.rs` covers:

- round-trips of `Box<Dummy>`, `Box<u64>`, `Box<Vec<u64>>` across `{compressed, uncompressed} x {checked, unchecked}`;
- `Box<[T]>` still round-trips and coexists with the new `Box<T>` impl;
- `Box<T>` serializes byte-identically to `T`, with equal `serialized_size`.

`no_std` is preserved (uses `ark_std::boxed::Box`). `cargo fmt --all -- --check` is clean and the full `ark-serialize` test suite passes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@
 ### Features
 
 - (`ark-serialize`) Implementation of `CanonicalSerialize` and `CanonicalDeserialize` for signed integer types
+- [\#1115](https://github.com/arkworks-rs/algebra/pull/1115) (`ark-serialize`) Implement `CanonicalSerialize`, `Valid`, and `CanonicalDeserialize` for `Box<T>`.
 - [\#1084](https://github.com/arkworks-rs/algebra/pull/1084) (`ark-ff`) Add `from_u128` const constructor for `SmallFp` fields
 - [\#1086](https://github.com/arkworks-rs/algebra/pull/1086) (`ark-ff-macros`) Auto-detect small prime subgroup (bases 3, 5, 7) in `define_field!` for both `SmallFp` and `Fp` fields
 
--- a/serialize/src/impls/misc.rs
+++ b/serialize/src/impls/misc.rs
@@ -3,6 +3,7 @@ use crate::{
 };
 use ark_std::{
     borrow::*,
+    boxed::Box,
     io::{Read, Write},
     marker::PhantomData,
     rc::Rc,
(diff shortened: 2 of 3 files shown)


weikengchen:
Let me double check---I remember that there used to be a very specific reason why Box<T> doesn't have those...

weikengchen (approved):
Reviewed the Box<T> impls, tests, no_std/parallel paths, and the Rust 1.75 coherence concern. Looks good to merge.

PR #1126, opened 2026-07-21
fix SmallFp mul overflow on u32

benbencik:
## Motivation

`SmallFp` multiplication can overflow on `u32` prime fields. This can be replicated on native tests for `TeddyBear` prime.

## Description

Edited case for primes `2^16 < p < 2^32` to handle the overflow. Copied the approach that is used for `u64` primes. Added `TeddyBear` definition and tests for this field.

### DIFF
--- a/ff-macros/src/small_fp/montgomery_backend.rs
+++ b/ff-macros/src/small_fp/montgomery_backend.rs
@@ -510,7 +510,7 @@ fn generate_mul_impl(
 
                         let mut t = (a.value as u64) * (b.value as u64);
                         let k = t.wrapping_mul(N_PRIME) & R_MASK;
-                        
+
                         let (t, overflow) = t.overflowing_add(k * MODULUS_MUL_TY);
                         let mut r = (t >> #k_bits) + ((overflow as u64) << #shift_bits);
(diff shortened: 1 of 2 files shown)


z-tech:
Heyy 👋 

I imagine you found a concrete op / test case that fails for this ? Curious what is.

Reason I'm curious is we have (a somewhat diverged) version of this code in the "z" repo. We were surprised when we were able to formalize the z repo against mathlib that there were seemingly no bugs. One theory is that the bugs were silently fixed when porting over the arithmetic (or during the process of formalizing itself).

Your finding would rule out the possibility that there were genuinely no bugs to begin with by counterexample

benbencik:
Interesting, perhaps the port somehow took care of this. I would be curious to know more. One simple example that fails is `(p-1)*(p-1)`.

z-tech:
> Interesting, perhaps the port somehow took care of this. I would be curious to know more. One simple example that fails is `(p-1)*(p-1)`.

Ah it also fails because we formally verify instances like BabyBear, KoalaBear, M31... not the generic implementation over all inputs, which makes sense.

Cool!

z-tech (approved):
Adding that example might be helpful for regression tests.

PR #1130, opened 2026-08-13
perf(ec): reserve exact capacity for G2Prepared coefficient vectors

4waan:
## Summary

Every `G2Prepared` constructor pushes a number of line coefficients that is fixed by the curve's loop-count constant, into a `Vec` created with `Vec::new()`. That way the vector reallocates its way up from the default first capacity on every single point preparation.

This reserves the exact count up front. Each capacity expression mirrors the shape of the loop directly below it, so the two cannot drift apart while staying readable.

Affected constructors (8 vectors across 5 models):

| model | vectors |
|---|---|
| `bls12` | `ell_coeffs` |
| `bn` | `ell_coeffs` |
| `bw6` | `ell_coeffs_1`, `ell_coeffs_2` |
| `mnt4` | `double_coefficients`, `addition_coefficients` |
| `mnt6` | `double_coefficients`, `addition_coefficients` |

## Measurements

BLS12-381, one `G2Prepared::from(G2Affine)`  counted with a `GlobalAlloc` wrapper that tallies every allocator call (numbers deterministic):

| | before | after |
|---|---|---|
| allocations | 1 | 1 |
| reallocations | 5 | **0** |
| reallocations that moved the buffer | 5 | **0** |
| bytes copied by reallocation | 35,712 | **0** |
| largest allocation request | 36,864 B | **19,584 B** |

`ell_coeffs` holds 68 `EllCoeff` of 288 bytes each. The existing growth sequence is 4 → 8 → 16 → 32 → 64 → 128, meaning the final buffer had 60 unused slots.. the new request of 19,584 bytes is exactly 68 × 288 (reservation is exact rather than a padded upper bound).

The wall-clock effect on my mac was small.. `G2Prepared::from` for BLS12-381 shows like 1µs improvement out of 91µs across A/B head on runs, well under 1% of a full pairing. But the substantive part of this change is the allocation counts: the reduced peak memory matters more for the `no_std` and targets that use this code.

## Correctness

Each loop is followed by `debug_assert_eq!(vec.len(), reserved)`. If some capacity expression is ever wrong in either direction, the existing pairing tests fail (instead of silently reallocating again).

Verified by running the pairing tests with `-C debug-assertions=yes` under `--release` for all nine pairing curves in this repo: BLS12-377, BLS12-381, BN254, BW6-761, BW6-767, MNT4-298, MNT4-753, MNT6-298, MNT6-753.

To confirm those assertions actually have teeth rather than passing vacuously, I added `+ 1` to the `mnt6` addition count and re-ran; `test_bilinearity` and `test_multi_pairing` both failed with `left: 48, right: 49`, and passed again once reverted.

checks:
- [x] Wrote unit tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,7 +17,7 @@
 - [\#1109](https://github.com/arkworks-rs/algebra/pull/1109) (`ark-poly`) Reduce allocations in `DenseMultilinearExtension::Sub` and parallelize scalar `Mul`.
 - [\#1112](https://github.com/arkworks-rs/algebra/pull/1112) (`ark-ec`) Fix rayon::ThreadPoolBuilder panicking in wasm32 when parallel feature is enabled
 - [\#1108](https://github.com/arkworks-rs/algebra/pull/1108) (`ark-ec`) Improve scalar multiplication speed for curves using GLV.
-- (`ark-ec`) Reduce allocations in `G2Prepared` construction for the `bls12`, `bn`, `bw6`, `mnt4` and `mnt6` models.
+- [\#1130](https://github.com/arkworks-rs/algebra/pull/1130) (`ark-ec`) Reduce allocations in `G2Prepared` construction for the `bls12`, `bn`, `bw6`, `mnt4` and `mnt6` models.
 
 ### Breaking changes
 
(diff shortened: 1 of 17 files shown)


weikengchen:
I will double check but I want to say apparently this is very minimal performance impact...

weikengchen:
Nay, after reviewing the PR, I believe that the code complexity is not worthy, especially that there is also a very small impact on the memory utilization as you pointed out.

4waan:
you re right @weikengchen 
Thank you for the review!

PR #1133, opened 2026-08-20
perf(ec): replace 2-bit GLV joint window with interleaved wNAF -5

### LINKED ISSUE #1128: Follow-up: replace 2-bit windowed GLV scalar multiplication with wNAF-5
## Summary

Follow-up on #1108. That PR moves the GLV scalar-multiplication used by the BLS12-377 / BLS12-381 / BN254 G1 `mul_projective` overrides from plain interleaved double-and-add to a 2-bit joint window with a 15-point table (ported from gnark-crypto's `mulGLV`). This issue tracks the next step: replacing that 2-bit scheme with **interleaved width-5 wNAF**.

## Context

The upstream reference (gnark-crypto) has since replaced the same 2-bit table with interleaved wNAF-5 in https://github.com/ConsenSys/gnark-crypto/pull/788 (merged 2026-01, not yet in a release), reporting:

- BLS12-381 G1: **−6.04 %**, G2: −8.56 %
- BW6-761 G1: −13.06 %, G2: −13.10 %

on top of the 2-bit table. The wNAF variant uses a width-5 signed-digit decomposition per half-scalar — two precomputed 8-entry odd-multiple tables, addition density ≈ 1/(w+1) — plus a subtraction path, all of which `ark-ec` already supports.

Current in-repo numbers for #1108 on an Apple M2 Ultra: BLS12-381 −22.8 %, BN254 −15.1 %, BLS12-377 ≈ −21.7 % (vs. the old double-and-add). So wNAF-5 is an incremental, independent win on top.

## Scope

- Replace the 2-bit joint window in `ec/src/scalar_mul/glv.rs` with interleaved width-5 wNAF.
- Keep the scalar decomposition (`scalar_decomposition` + `endomorphism`) unchanged — only the joint scan changes.
- Repeat the correctness tests (existing `glv` suites + edge cases) and benchmark the same three curves on the same saved baselines so the delta is directly comparable to #1108.

## Definition of done

Benchmarks on ≥ one GLV curve showing ≥ the reported improvement, plus the full `glv` test suites passing. Given gnark's measurements, a single-digit-℅ gain may be within noise on some curves; an acceptable finding can also be that wNAF-5 is not worth the added table/complexity on our curves.

## Reference

- Implemented in PR: #1108
- gnark-crypto reference: https://github.com/ConsenSys/gnark-crypto/blob/master/ecc/bls12-381/g1.go#L774 (`mulGLV`, `wnafWindow = 5`) and its PR https://github.com/ConsenSys/gnark-crypto/pull/788

4waan:
## Description

Implements #1128  (follow-up to #1108). Two changes to the GLV joint scan, both confined to
`ec/src/scalar_mul/glv.rs`. No public API change; `scalar_decomposition` and `endomorphism`
are untouched.

1. Interleaved width-5 wNAF replaces the 2-bit joint window.
2. The second base's table is derived from the first through the endomorphism. Since
   `b2 = s1*s2*phi(b1)` and phi is a homomorphism, `(2i+1)*b2 = s1*s2*phi((2i+1)*b1)`, so one
   doubling and 7 additions build both tables.

### Result

Base-field multiplications per `glv_mul_projective`, counted over real decompositions.
Jacobian `add-2007-bl` = 11M + 5S, `dbl-2009-l` = 2M + 5S, S = 0.8M.

| curve | vs 2-bit window |
|---|---|
| BLS12-381 G1 | **-16.9%** |
| BN254 G1 | -16.9% |
| BLS12-377 G1 | -16.8% |
| BW6-761 G1 | -16.5% |
| Pallas | -17.0% |

Operation count is the claim; wall clock is not, since this machine's cross-process noise
floor is 5-8%.

### Window width

W = 5, matching gnark-crypto. Same counting, relative to W = 5: W = 3 +14.9%, W = 4 +4.3%,
W = 6 +2.3%. On BW6-761, W = 6 is ahead by 0.3%, so 5 is a BLS12/BN optimum rather than a
universal one. A compile-time guard rejects anything outside `2..=6`: at W = 7 the sign
correction `low - (1 << W)` overflows `i8` and panics in any debug build, though release
wraps to the correct residue.

closes: #1128

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,7 +17,7 @@
 - [\#1109](https://github.com/arkworks-rs/algebra/pull/1109) (`ark-poly`) Reduce allocations in `DenseMultilinearExtension::Sub` and parallelize scalar `Mul`.
 - [\#1112](https://github.com/arkworks-rs/algebra/pull/1112) (`ark-ec`) Fix rayon::ThreadPoolBuilder panicking in wasm32 when parallel feature is enabled
 - [\#1108](https://github.com/arkworks-rs/algebra/pull/1108) (`ark-ec`) Improve scalar multiplication speed for curves using GLV.
-- [\#N](https://github.com/arkworks-rs/algebra/pull/N) (`ark-ec`) Replace the 2-bit joint window in GLV scalar multiplication with interleaved width-5 wNAF.
+- [\#1133](https://github.com/arkworks-rs/algebra/pull/1133) (`ark-ec`) Replace the 2-bit joint window in GLV scalar multiplication with interleaved width-5 wNAF.
 
 ### Breaking changes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,7 +17,7 @@
 - [\#1109](https://github.com/arkworks-rs/algebra/pull/1109) (`ark-poly`) Reduce allocations in `DenseMultilinearExtension::Sub` and parallelize scalar `Mul`.
 - [\#1112](https://github.com/arkworks-rs/algebra/pull/1112) (`ark-ec`) Fix rayon::ThreadPoolBuilder panicking in wasm32 when parallel feature is enabled
 - [\#1108](https://github.com/arkworks-rs/algebra/pull/1108) (`ark-ec`) Improve scalar multiplication speed for curves using GLV.
-- [\#1133](https://github.com/arkworks-rs/algebra/pull/1133) (`ark-ec`) Replace the 2-bit joint window in GLV scalar multiplication with interleaved width-5 wNAF.
+- [\#1133](https://github.com/arkworks-rs/algebra/pull/1133) (`ark-ec`) Replace the 2-bit joint window in GLV scalar multiplication with interleaved width-5 wNAF, and derive the second base's precomputed table from the first through the endomorphism instead of building it with group arithmetic.
 
 ### Breaking changes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@
 - [\#1109](https://github.com/arkworks-rs/algebra/pull/1109) (`ark-poly`) Reduce allocations in `DenseMultilinearExtension::Sub` and parallelize scalar `Mul`.
 - [\#1112](https://github.com/arkworks-rs/algebra/pull/1112) (`ark-ec`) Fix rayon::ThreadPoolBuilder panicking in wasm32 when parallel feature is enabled
 - [\#1108](https://github.com/arkworks-rs/algebra/pull/1108) (`ark-ec`) Improve scalar multiplication speed for curves using GLV.
+- [\#N](https://github.com/arkworks-rs/algebra/pull/N) (`ark-ec`) Replace the 2-bit joint window in GLV scalar multiplication with interleaved width-5 wNAF.
 
 ### Breaking changes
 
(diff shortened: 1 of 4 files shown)


weikengchen:
I will let my Codex take a look and check the performance numbers on my machine and cross-check the window size issue.

4waan:
1 doubling + 7 projective adds build both wnaf tables.. The second one is free: `b2 = s1*s2*phi(b1)` and phi is a homomorphism, so `(2i+1)*b2 = s1*s2*phi((2i+1)*b1)`. `glv_mul_projective` now passes `sgn_k1 == sgn_k2` instead of `b2` and the second table is table 1 mapped through endomorphism.

Per call: -1 doubling, -7 projective adds, +8 base-field muls (endomorphism is `res.x *= ENDO_COEFFS[0]` everywhere), +8 negations on ~half of calls. bls12-381 G1 counted over 100k real decompositions:- 107 M equiv of 1704, 6.3% (6.4% at S/M=0.8, 6.2% at 1.0). Single scalar mul only (MSM doesnt use this path)

Needs less of phi than the existing code does.. only the homomorphism property, which holds on the whole curve, vs phi(p) = lambda*p which needs the subgroup -> identity servives

we ve got one new requirement:- endomorphism must be right at arbitrary z, not just z = 1, since it now sees 8 table entries. True for every in-tree impl, but only tested at z = 1 so I added `glv_endomorphism_projective`.

I re ran the width count. W = 5 still wins so the constant is unchanged, but W = 4 goes +0.5% -> +4.2% and W = 6 +8.9% -> +2.3%. Fixed the now-stale sentence in the `GLV_WNAF_WIDTH` doc.

537 tests green across the 6 GLV curve crates, 265 in ark-test-curves.

4waan:
### Description vs what the branch does now

two of the later commits changed things the description still describes the old way. @weikengchen ig this would matter for your cross-check (the window size table in the body is one of the stale parts in dscription) 

| Body says | Branch now does |
|---|---|
| Three private helpers | Four, `glv_endo_odd_multiples` added |
| 1 doubling + 7 additions per base, 2 tables built | 1 doubling + 7 additions total; the second table is the first mapped through `phi` |
| -9.99% / -10.08% / -9.80% | Counted before the derived table, so understated. Composes to roughly -15.7% with the -6.4% above, but that mixes two sample sizes, so worth a direct count |
| "width 4 comes within half a percentage point, nearly free" | W = 4 is +4.2% (was +0.5%), W = 6 is +2.3% (was +8.9%). W = 5 still wins, constant unchanged |
| Tests: added `glv_edge_cases` | Also `glv_endomorphism_projective` |

this landed after the body: width guard `f1e20e0`, zeroized recodings and timing docs `f25332e`, wrap-safety comment `b8d53cb`.

One thing for the benchmark: gnark builds both tables with group arithmetic, so `glv_endo_odd_multiples` sits on top of the reference rather than being part of the port. Comparing this branch against gnark's -6.04% is not comparing the same algorithm.

4waan:
@weikengchen a heads up since the body was out of date when you offered to run it. changed reference group arithmetic for gnark tables, against gnark's -6.04% aint like for like. and second, your run might disagree on width since W-5 wins on bls12 and bn but W=6 is 0.3%

Glad to help with anything else fs!

also lgtm so far, i had some doubts so i tested my work here with 
- a glv oracle (116,763 multiplications (38,921 adversarial scalars * 3 entry points * 11 configs), everything matched!)
- covered: λ order 3 mod r, β order 3 in F, both lattice rows in {(a,b) : a+bλ≡0}, determinant(N) = +r exactly; endomorphism(G) == λG recomputed externally, all pass
- off subgroup base layer where phi(p) = λp is false but homomorphism holds.. will remain identical post merge..
- sweep across the window width (a W ∈ {2,3,4,6} re-run through the differential tests.. guard correctly rejects 1 and 7 at compile time)

PR #1135, opened 2026-09-09
Evaluate WBMap isogeny polynomials without allocating; run small batch inversions serially under `parallel`

thomasahle:
`IsogenyMap::apply` builds four `DensePolynomial`s on the heap on every call, only to evaluate them once, and `DensePolynomial::evaluate` folds from zero, which spends one extra multiplication per polynomial. This evaluates the coefficient slices in place with Horner's rule.

No change in results: the hash-to-curve test vectors for BLS12-381 G1 and G2 pass, as do the `ark-ec` hashing tests. It also removes the only use of `ark_poly` in this file.

I am preparing a follow-up that lets a curve specialise its isogeny evaluation (BLS12-381 G1's 11-isogeny can be done in 30 field multiplications instead of 49 with preprocessed coefficients); this cleanup is independent of it.

---

**Maintainer note (weikengchen):** while benchmarking this PR I found that the `parallel` feature made `IsogenyMap::apply` about 8x slower than the serial build: `batch_inversion` split even its two-element batch across rayon, costing roughly 75 µs of pool dispatch per call (isogeny step 85 µs instead of 10 µs; BLS12-381 G1 `WBMap::map_to_curve` 110 µs instead of 65 µs on an M2 Ultra). I fixed that in passing here: with `parallel`, `batch_inversion_and_mul` now runs serially below 2048 elements and keeps chunks of at least 128 elements, so batches of 2048 and above are unchanged or slightly faster. The same push restores the `WBConfig` rustdoc that the new helper displaced and drops the now-unused `ark-poly` dependency of `ark-ec`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Pending
 
+- (`ark-ec`) Evaluate the isogeny polynomials of `WBMap` without allocating temporary polynomials.
 - (`ark-starkcurve`) Add 252 bit [Stark curve](https://docs.starknet.io/architecture/cryptography/#the_stark_curve).
 - [\#971](https://github.com/arkworks-rs/algebra/pull/971) (`ark-ff`) Make serial_batch_inversion_and_mul public.
 - Consolidated logic into `bitreverse_permutation_in_place` and made it public.
--- a/ec/src/hashing/curve_maps/wb.rs
+++ b/ec/src/hashing/curve_maps/wb.rs
@@ -1,8 +1,7 @@
 use core::marker::PhantomData;
 
 use crate::{models::short_weierstrass::SWCurveConfig, CurveConfig};
-use ark_ff::batch_inversion;
-use ark_poly::{univariate::DensePolynomial, DenseUVPolynomial, Polynomial};
+use ark_ff::{batch_inversion, Field};
 
 use crate::{
     hashing::{map_to_curve_hasher::MapToCurve, HashToCurveError},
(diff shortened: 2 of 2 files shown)


weikengchen:
Let me get Claude to take this look. Very good insight!

weikengchen:
Thanks, good cleanup. Reviewed against the RFC 9380 vectors (BLS12-381 G1/G2 here, BLS12-377 G1/G2 in the curves workspace) plus a differential check of the old and new `apply` on 20k random G1 and 10k random G2 points: outputs are identical.

I pushed two commits on top of yours:

- 39b3ad60: restore the `WBConfig` rustdoc that the new helper displaced, drop the now-unused `ark-poly` dependency of `ark-ec`, and link the changelog entry.
- 910c6cad: an `ark-ff` fix found while benchmarking this. With the `parallel` feature, `batch_inversion` split even the two-element batch in `IsogenyMap::apply` across rayon, so the isogeny step cost about 85 µs instead of 10 µs (BLS12-381 G1 `map_to_curve` 110 µs vs 65 µs on an M2 Ultra). It now runs serially below 2048 elements; larger batches are unchanged or slightly faster. Details in the PR description.

On the numbers, for your planned follow-up: this PR's change saves 0.2 to 0.8 µs per isogeny application, under 1% of `map_to_curve`. The serial isogeny step is about 9.7 µs, of which the single field inversion is 4.3 µs, so cutting 49 multiplications to 30 would save roughly 2 µs, around 3% of `map_to_curve`. Worth knowing before investing in it.

weikengchen (approved):

PR #1140, opened 2026-09-21
fix(ff): pad `expand_message_xmd` with the hash block size in `DefaultFieldHasher`

davxy:
`DefaultFieldHasher::new` set the `Z_pad` length of `expand_message_xmd` to the field element length `L` instead of the hash input block size (RFC 9380 section 5.3.1). The two agree for BLS12-381 with SHA-256, the
only suite with test vectors, and differ for every 255-bit or 256-bit field, where the output was not RFC 9380 compliant.

The fix adds `H: BlockSizeUser` and uses `H::block_size()`. A new test checks `hash_to_field` on secp256k1 against the RFC 9380 Appendix J.8.1 vectors; it fails on `master`.

Breaking: the output changes for every field where `L` differs from the block size, and `H` now requires `digest::crypto_common::BlockSizeUser`.


davxy:
fyi @drskalman

